PR #11, opened 2025-06-21
feat(mangareader): add more sources (incomplete)

kkantan:
here for when/if https://mangareader.to and https://manganow.to fix their page images.

the mangareader.to page interceptor isn't complete since I couldn't test on any images.


kkantan:
it looks like the sites are broken permanently, and there is no reason to leave this open.

PR #73, opened 2025-07-28
feat: add mangadistrict

dhayab:
Hi, this source pulls content from [MangaDistrict](https://mangadistrict.com/) (⚠️ NSFW on the homepage), taking inspiration from the structure and implementation of the en.weebcentral source.


sources/en.mangadistrict/res/filters.json
```diff
@@ -0,0 +1,280 @@
...
+		"options": ["AND", "OR"]
+	},
+	{
+		"type": "select",
+		"id": "adult",
+		"title": "Adult content",
+		"usesTagStyle": true,
+		"options": ["Yes", "No", "Only"]
+	}
+]
```
kkantan: make sure to end all files with a newline, including jsons
```suggestion
]

```

sources/en.mangadistrict/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "en.mangadistrict",
+		"name": "MangaDistrict",
+		"version": 1,
+		"url": "https://mangadistrict.com/",
```
kkantan: the base url shouldn't have a trailing slash
```suggestion
		"url": "https://mangadistrict.com",
```

sources/en.mangadistrict/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+
+			manga.content_rating = if html.select_first(".manga-title-badges.adult").is_some() {
+				ContentRating::NSFW
+			} else {
+				ContentRating::Safe
+			};
+
+			manga.viewer = Viewer::Webtoon;
+
+			if needs_chapters {
```
kkantan, line 120: this shouldn't be nested inside the `needs_details` check, since i think it may possible for aidoku to request chapters but not manga details

sources/en.mangadistrict/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+		Ok(pages)
+	}
+}
+
+impl ListingProvider for MangaDistrict {
+	fn get_manga_list(&self, _listing: Listing, _page: i32) -> Result<MangaPageResult> {
+		Err(AidokuError::Unimplemented)
+	}
+}
+
```
kkantan: you can just remove this if you don't implement it
```suggestion
```

kkantan (changes requested):
thank you very much for the contribution!
I left a few comments about some small things, and here are a few more things:
- there are a couple of clippy warnings to fix. you can see these locally by doing `cargo clippy`, and might just be able to autofix them
- it looks like the cover images for a few series aren't loading. the common factor might be that they're jpgs. is it the same for you? you might need to add an image request provider with a referer header
- this source is actually a madara source, which is a template I've been meaning to write. this is fine as-is, but in the future it might make the source easier to maintain if it's migrated to the template when it's done, but it depends on how much the source differs from the base template

dhayab:
I figured out the image issue. Cloudflare (or a WP plugin?) returns different HTML in the simulator/device than when running tests locally. It puts the cover URL in a `data-cfsrc` attribute instead.

I've looked at the template to migrate, and I'll probably do a passable job porting it + implementing it for this source, as I don't usually work with Rust. I'll leave that to someone more familiar or interested.

In the meantime I've applied your suggestions and I'll let you decide whether to merge it or wait for a more comprehensive migration.

sources/en.mangadistrict/src/lib.rs
dhayab: I extracted it from the `needs_details` branch, and I moved the HTML fetching outside its scope. Let me know if that works for you, if you'd prefer fetching twice (within each branch), or something else.

kkantan (approved):
looks good to me, thank you! 👍

PR #75, opened 2025-07-31
feat(libgroup): implement core MVP for the source template

remytomme:
**Features:**
- Full search functionality (name, filters, tags, genres)
- Manga details & chapters list
- Homepage listing
- User-configurable settings
- Alternate covers support
- Authenticated access
- Rich home screen
- Backward compatibility with older source

Complete implementation of LibGroup source template including core MVP and comprehensive search and filtering capabilities.


templates/libgroup/src/auth/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		Ok(response)
+	}
+}
+
+/// Retrieves the stored authentication token from defaults.
+fn get_token() -> Result<TokenResponse> {
+	defaults_get_json::<TokenResponse>(TOKEN_KEY)
+		.map_err(|_| AidokuError::Message("No token".into()))
```
kkantan, line 58: for future reference, there's an `error!` macro that shortens this. I believe you can also do `AidokuError::message` to avoid having to do `.into()` as well, as an alternative. there's no need to change it from what you have though.
```suggestion
		.map_err(|_| error!("No token"))
```

kkantan (approved):
looks good to me, thank you for the contribution!

PR #84, opened 2025-08-18
feat: add fr.bigsolo

apix0n:
this pr adds a source for french scanlation team website [BigSolo](https://bigsolo.org).

## why is x code written like that ?

this is the second time i’ve used rust lmao, please don’t judge too harshly

i made the choice to replicate how the website works instead of parsing its html directly, as it’s a client only website with no pre-rendering
and i mostly wasn’t sure if was possible to parse js-loaded sites on aidoku at all.

while this solves some issues like:
- the card on home being divs with event handlers, not \<a\> elements (so, even if i found the way to handle the page load, i have no idea how i would’ve gotten the series links.)
  > here, they are loaded from the json directly, so no need for js parsing
- the manga reader taking too long to load (because it’s preloading every page)
  > here, i’m querying the reader api directly which is actually really efficient and responds in like max 200ms 

it also causes problems:
- because the chapter keys are set to the imagechest upload ids, it means that we are _forced_ to use imagechest and that if the dev was to change the website cdn, we’d need to update the source.
- for above, also no idea how it’d work if they were to mix and match cdns, maybe i’ll implement a `{service}:{id}` chapter key ?
- the slug of a series is handled client-side and i didn’t manage to make the DeepLink work so i just omit it

the website is being reworked so most of these issues may be fixable Soon™; when it’s the case i’ll update this source.


apix0n:
> overall this looks good, but there's one main thing that should to be changed: not implementing `get_manga_update` will likely cause issues with parts of the aidoku app, such as the history tab. I'm not sure entirely what else, but it goes against the general convention all other sources follow. ideally, the `get_search_manga_list` result should contain minimal data that's then fully populated in `get_manga_update`.

updated that. i'm guessing this `get_manga_update` also handles DeepLinks? may be the case why i didn't manage to make the manga pages load.

> the site also loads very slowly for me. is this because I'm not located in france?

weird; it shoudln't be loading slowly so it may be a location problem. at least for me it's loading in acceptable times.

kkantan (approved):

kkantan:
> i'm guessing this `get_manga_update` also handles DeepLinks?

yes, I believe that is the case.

PR #93, opened 2025-09-07
Implement Suwayomi source

gh-adrl:


gh-adrl:
Also I don't know rust. Was porting over my implementation: https://github.com/gh-adrl/aidoku-community-sources/tree/suwayomi to the new api and it was mostly copying the other sources.

sources/multi.suwayomi/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+				.map(|m| m.into_manga(&base_url))
+				.collect(),
+			has_next_page: false,
+		})
+	}
+
+	fn get_manga_update(
+		&self,
+		mut manga: Manga,
+		_needs_details: bool,
```
gh-adrl: Maybe as a follow up, can look into not fetching the description for `Manga` and fetch it here

sources/multi.suwayomi/res/settings.json
kkantan, line 1: make sure this file uses tabs for indentation as well

sources/multi.suwayomi/res/source.json
```diff
@@ -0,0 +1,71 @@
+{
+	"info": {
+		"id": "mulit.suwayomi",
+		"name": "Suwayomi",
+		"version": 1,
+		"url": "https://github.com/Suwayomi",
+		"contentRating": 0,
+		"languages": [
```
kkantan, line 8: I wonder if it would be better to tag this as `"multi"` for the only language, rather than providing a bunch of languages that may or may not actually contained in the source's content, as it changes for the user. I believe this is what the built-in komga source does.

sources/multi.suwayomi/res/source.json
```diff
@@ -0,0 +1,71 @@
...
+            "uk",
+            "vi"
+		]
+	},
+    "listings": [
+      {
+        "id": "Library",
+        "name": "Library"
+      }
+    ]
```
kkantan: you're using spaces here but tabs earlier in the file, make sure it's all tabs

sources/multi.suwayomi/src/lib.rs
kkantan: if you don't handle detail fetching in this function then I believe it will causes issues with parts of the app, as we saw with another source being pr'd here. things like the history tab and deep link handlers require detail fetching given only the manga id

sources/multi.suwayomi/res/source.json
```diff
@@ -0,0 +1,71 @@
...
+            "th",
+            "tr",
+            "uk",
+            "vi"
+		]
+	},
+    "listings": [
+      {
+        "id": "Library",
+        "name": "Library"
```
kkantan: I think the id is used as the name if it's not specified, or vice versa, so you should only have to specify one of the fields here since they're the same

sources/multi.suwayomi/res/source.json
gh-adrl: Which one should I use?

Looking at the schema https://github.com/Aidoku/aidoku-rs/blob/db934c1d181d04048258dbcb90f872dfcc7e1919/crates/cli/src/supporting/schema/source.schema.json#L43, it would require a valid language code

gh-adrl: Wait that's for cli. I don't see other instances of a multi though.

kkantan: the built-in komga and local files sources use it, so I think it should be fine: https://github.com/Aidoku/Aidoku/blob/38b48a415e6fa61e903c08cc70ab42d9e499b162/Shared/Models/SourceInfo.swift#L25

gh-adrl: Sounds good, changed it.

sources/multi.suwayomi/res/source.json
kkantan, line 1: this file should also use tabs for indentation

sources/multi.suwayomi/res/source.json
```diff
@@ -0,0 +1,17 @@
+{
+	"info": {
+		"id": "mulit.suwayomi",
```
kkantan: oops, I didn't see this before
```suggestion
		"id": "multi.suwayomi",
```

kkantan (changes requested):
after trying it locally, I think there are a few more things that should be changed:

if there is no configured base url, the source returns a network error. I think you should check this first, and if it's empty just return a message (using `error!` or `bail!`) prompting users to configure it in the source settings.

the base url text field setting should use the url keyboard type. you can copy the options from here: https://github.com/Aidoku-Community/sources/blob/b6356df8af842f8e8ae5e51cfc7709610529d423/sources/ru.slashlib/res/settings.json#L11-L13 (and set the `refreshes` option to `["listings"]`)

there's a `BaseUrlProvider` option for sources to implement (https://github.com/Aidoku/aidoku-rs/blob/db934c1d181d04048258dbcb90f872dfcc7e1919/crates/lib/src/structs/source.rs#L94). I think this would allow the configured base url to be opened from the web view button instead of the github repo, which I think would be better.

gh-adrl:
Also it looks like I fail the lint source schema lint if I remove the url. Fails this schema: https://github.com/Aidoku/aidoku-rs/blob/main/crates/cli/src/supporting/schema/source.schema.json#L67

PR #108, opened 2025-09-26
feat: add en.flamecomics

### LINKED ISSUE #24: flamecomics.xyz
https://flamecomics.xyz/

Sheepodeepo:
Added/Rewrote flamecomics with A-Z and Z-A sorting + filtering mangas by genre tags.

close #24

### DIFF
--- a/sources/en.flamecomics/src/filter.rs
+++ b/sources/en.flamecomics/src/filter.rs
@@ -17,16 +17,14 @@ pub fn get_genre_filter(filters: &Vec<FilterValue>) -> String {
 }
 
 pub fn sort(filters: &Vec<FilterValue>, mut entries: Vec<Manga>) -> Vec<Manga> {
-	let mut sort_by_ascending: bool = true;
 	for filter in filters {
 		if let FilterValue::Sort { ref ascending, .. } = *filter {
-			sort_by_ascending = *ascending;
+			if *ascending {
+				entries.sort_by(|a: &Manga, b: &Manga| b.title.cmp(&a.title));
+			} else {
+				entries.sort_by(|a: &Manga, b: &Manga| a.title.cmp(&b.title));
+			}
 		}
-	}
-	if sort_by_ascending {
-		entries.sort_by(|a, b| a.title.cmp(&b.title));
-	} else {
-		entries.sort_by(|a, b| b.title.cmp(&a.title));
 	}
 	entries
 }


sources/en.flamecomics/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "enflamecomics"
```
kkantan: I think it's better to name the rust package more clearly when possible, and since there are no other "flame" sources this should be fine
```suggestion
name = "flamecomics"
```

sources/en.flamecomics/res/filters.json
```diff
@@ -0,0 +1,128 @@
...
+            "Swordmaster",
+            "System",
+            "Thriller",
+            "Time Travel",
+			"Tragedy",
+			"VR",
+            "Youth"
+        ]
+	}
+]
```
kkantan: all files should have a newline at the end for consistency
```suggestion
]

```

sources/en.flamecomics/res/source.json
```diff
@@ -0,0 +1,10 @@
+{
+	"info": {
+		"id": "en.flamecomics",
+		"name": "FlameComics",
```
kkantan: the previous source had a space here, and it looks like the website title also has a space
```suggestion
		"name": "Flame Comics",
```

sources/en.flamecomics/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+	imports::{html::*, net::*},
+	prelude::*,
+};
+use chrono::{DateTime, FixedOffset, NaiveDateTime, TimeZone, Utc};
+
+struct FlameComics;
+
+const BASE_URL: &str = "https://flamecomics.xyz";
+mod filter;
+impl Source for FlameComics {
```
kkantan, line 18:
```suggestion
use chrono::{DateTime, FixedOffset, NaiveDateTime, TimeZone, Utc};

const BASE_URL: &str = "https://flamecomics.xyz";

mod filter;

struct FlameComics;

impl Source for FlameComics {
```

sources/en.flamecomics/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		let manga_url = format!("{BASE_URL}{}", manga.key);
+		let html = Request::get(&manga_url)?.html()?;
+
+		if needs_details {
+			let manga_authors = ".mantine-Grid-root .mantine-Grid-col:nth-of-type(1) .mantine-Stack-root .mantine-Stack-root";
+			let manga_details = ".mantine-Grid-root .mantine-Grid-col:nth-of-type(2) .mantine-Stack-root .mantine-Paper-root:nth-of-type(1)";
+			let desc_selector = format!("{manga_details}{}", " p");
```
kkantan: if you want to use format! here I think it makes more sense to have the space inside the format string
```suggestion
			let desc_selector = format!("{manga_details} {}", "p");
```
and this should apply to the other ones as well

however, I think it's more efficient if you instead just select the details root element, then select the other elements with the simplified selector:
```rs
			let details = html.select_first(manga_details).ok_or(error!("missing details"))?;
			let description = details.select_first("p"); // ...
```

sources/en.flamecomics/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+			// ex: https://flamecomics.xyz/series/2/79c2cf38ecc5fd25
+			let html = Request::get(&url)?.html()?;
+			let manga_key = html
+				.select_first(
+					".mantine-Container-root .mantine-Grid-root .mantine-Grid-col .mantine-focus-auto",
+				)
+				.and_then(|e| e.attr("href"))
+				.and_then(|url| url.strip_prefix(BASE_URL).map(|s| s.to_owned()))
+				.ok_or(AidokuError::message("Missing manga key"))?;
+			let chapter_key = key.strip_prefix(&manga_key).unwrap_or_default().into();
```
kkantan: it looks like you're just stripping `/series/2` from the url path. is there a reason you need to select this from the html instead of just assuming the final section of the url is a chapter? the series key won't contain more `/`s, right?

Sheepodeepo: yeah, you're right. in hindsight this is a easy fix of stripping the chapter section first then getting the manga_key but for some reason I thought it was super complicated so I parsed from html instead.

sources/en.flamecomics/res/filters.json
kkantan, line 1: it looks like this file has an odd mixture of both spaces and tabs for indentation

sources/en.flamecomics/res/filters.json
```diff
@@ -0,0 +1,128 @@
...
+            "Survival",
+            "Swordmaster",
+            "System",
+            "Thriller",
+            "Time Travel",
+			"Tragedy",
+			"VR",
+            "Youth"
+		],
+		"ids": [
```
kkantan: am I missing something, or are the ids the same as the options? if they are, the `ids` field can be omitted since it uses the option names as the default values.

sources/en.flamecomics/res/filters.json
```diff
@@ -0,0 +1,128 @@
...
+		"canAscend": true,
+		"options": [
+			"A-Z"
+		],
+		"ids": [
+			"A-Z"
+		],
+		"default": {
+			"index": 0,
+			"ascending": true
```
kkantan: I think it makes more sense to default to non-ascending (but still sort A-Z), so that the arrow pointing down means A-Z and pointing up means Z-A.
```suggestion
			"ascending": false
```

sources/en.flamecomics/res/filters.json
```diff
@@ -0,0 +1,128 @@
...
+		}
+	},
+	{
+		"type": "select",
+		"id": "genre",
+		"title": "Tags",
```
kkantan, line 18: I think if you add the "isGenre" field, you will be able to tap tags inside entries to view tag listings.

kkantan (changes requested):
just a few small things with the filters, otherwise looks good

PR #113, opened 2025-09-29
fix(multi.mangadex): fix wrong sort from alt cover art, add show locked chapters & current seasonal list

c0ntens:
Change:
- Adds show locked chapters which chapters are missing/unavailable.
- Adds current seasonal list with updated checks.
- Fixes wrong sort desending from alternative cover art.

### DIFF
--- a/sources/multi.mangadex/res/settings.json
+++ b/sources/multi.mangadex/res/settings.json
@@ -32,8 +32,8 @@
 			{
 				"type": "switch",
 				"key": "lockedChapters",
-				"title": "Shows Unavailable Chapters",
-				"subtitle": "Displays chapters that has been removed.",
+				"title": "Show Unavailable Chapters",
+				"subtitle": "Display chapters that has been removed",
 				"default": true,
 				"refreshes": ["content"]
 			},
--- a/sources/multi.mangadex/res/settings.json
+++ b/sources/multi.mangadex/res/settings.json
@@ -33,7 +33,7 @@
 				"type": "switch",
 				"key": "lockedChapters",
 				"title": "Shows Unavailable Chapters",
-				"subtitle": "Usually due to DMCA takedown requests, scanlator requests, or an official copyright claim",
+				"subtitle": "Displays chapters that has been removed.",
 				"default": true,
 				"refreshes": ["content"]
 			},
--- a/sources/multi.mangadex/res/settings.json
+++ b/sources/multi.mangadex/res/settings.json
@@ -30,6 +30,14 @@
 				"refreshes": ["content"]
 			},
 			{
+				"type": "switch",
+				"key": "lockedChapters",
+				"title": "Shows Unavailable Chapters",
+				"subtitle": "Usually due to DMCA takedown requests, scanlator requests, or an official copyright claim",
+				"default": true,
+				"refreshes": ["content"]
+			},
+			{
 				"type": "multi-select",
 				"key": "contentRating",
 				"title": "Default Content Rating",
--- a/sources/multi.mangadex/res/source.json
+++ b/sources/multi.mangadex/res/source.json
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "multi.mangadex",
 		"name": "MangaDex",
-		"version": 10,
+		"version": 11,
 		"url": "https://mangadex.org",
 		"contentRating": 1,
 		"languages": [
--- a/sources/multi.mangadex/src/home.rs
+++ b/sources/multi.mangadex/src/home.rs
@@ -16,6 +16,7 @@ use aidoku::{
 };
 use chrono::{TimeZone, Utc};
 use hashbrown::HashSet;
+use regex::Regex;
 
 impl Home for MangaDex {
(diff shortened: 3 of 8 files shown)


sources/multi.mangadex/res/settings.json
```diff
@@ -29,6 +29,14 @@
 				"default": false,
 				"refreshes": ["content"]
 			},
+			{
+				"type": "switch",
+				"key": "lockedChapters",
+				"title": "Shows Unavailable Chapters",
+				"subtitle": "Displays chapters that has been removed.",
```
kkantan:
```suggestion
				"title": "Show Unavailable Chapters",
				"subtitle": "Display chapters that has been removed",
```

sources/multi.mangadex/res/source.json
```diff
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "multi.mangadex",
 		"name": "MangaDex",
-		"version": 9,
+		"version": 10,
```
kkantan: mangadex is already version 10, I think you need to pull main

sources/multi.mangadex/src/home.rs
```diff
@@ -57,6 +58,49 @@ impl Home for MangaDex {
 			})
 			.collect::<Vec<_>>();
 
+		// fetch seasonal list
+		struct SeasonalList<'a> {
+			name: String,
+			entries: Vec<&'a str>,
+		}
+		let owner_user_id = "d2ae45e0-b5e2-4e7f-a688-17925c2d7d6b";
+		let mut seasonal_res = Request::get(format!("{API_URL}/user/{owner_user_id}/list"))?.send()?;
```
kkantan: this request should be included in the `send_all` above so that it can be parallelized

sources/multi.mangadex/src/home.rs
```diff
@@ -296,6 +347,55 @@ impl Home for MangaDex {
...
+				Request::get(format!(
+					"{API_URL}/manga\
+						?limit=100\
+						&includes[]=cover_art\
+						{content_ratings}\
+						&ids[]={}",
+					list.entries.join("&ids[]=")
+				))
+				.unwrap()
+			}));
```
kkantan: these requests can be included in the `send_all` after sending the basic home layout to parallelize

sources/multi.mangadex/src/lib.rs
```diff
@@ -296,7 +299,7 @@ impl Source for MangaDex {
 				let base_url = format!(
 					"{}/{}/{}",
 					response.base_url,
-					if data_saver { "dataSaver" } else { "data" },
+					if data_saver { "data-saver" } else { "data" },
```
kkantan, line 301: this was already fixed on main

sources/multi.mangadex/src/lib.rs
```diff
@@ -366,6 +370,71 @@ impl MangaDex {
 		Ok(ids)
 	}
 
+	// get a seasonal list
+	fn get_current_seasonal_list(&self) -> Result<MangaPageResult> {
```
kkantan: can we reduce the code duplication here? it's pretty much the same as in the home function. you can split out the filtering/parsing logic, but you'll have to duplicate the request sending

kkantan (changes requested):

sources/multi.mangadex/src/lib.rs
c0ntens: removed for code duplication.

sources/multi.mangadex/src/home.rs
c0ntens: removed for code duplication.

sources/multi.mangadex/res/settings.json
kkantan: you still have the period on the subtitle. I don't think this needs to be included since it isn't in the other subtitles.

sources/multi.mangadex/src/home.rs
c0ntens: i think i dont need to use `send_all` it?

sources/multi.mangadex/res/settings.json
c0ntens: oh i see.

sources/multi.mangadex/src/home.rs
kkantan: ideally, if it were included in the `send_all` above, all the requests would execute at once and it would move on after they all complete. right now it executes all of the previous requests, blocks until they all complete, then executes this request, then blocks until it completes. since the skeleton layout isn't sent until all of these requests complete and this request doesn't depend on results from the previous ones, it would make more sense to execute them all together.

however, it does add some complexity so if you don't really care to implement it I can't really blame you, as long as the home page doesn't take noticeably longer to load.

c0ntens: how i change `send_all` that? but i added checks this response is okay.

kkantan: alright I ran it and it doesn't seem to be noticeably slower, so it's probably fine to just leave it

kkantan (approved):

PR #117, opened 2025-10-02
feat: add multi.nhentai

dyphire:


sources/multi.nhentai/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+crate-type = ["cdylib"]
+
+[profile.dev]
+panic = "abort"
+
+[profile.release]
+panic = "abort"
+opt-level = "s"
+strip = true
+lto = true
```
kkantan: we should have a newline at the end of all files for consistency
```suggestion
lto = true

```

sources/multi.nhentai/res/filters.json
```diff
@@ -0,0 +1,8481 @@
...
+			"chinese",
+			"korean",
+			"french",
+			"spanish",
+			"german",
+			"russian",
+			"portuguese",
+			"italian"
+		]
+	},
```
kkantan: instead of having a filter for languages, I think the convention is to provide language codes for the source, so the user's system languages are automatically selected. it can allow multi-select for languages like the mangadex source.

sources/multi.nhentai/res/filters.json
```diff
@@ -0,0 +1,8481 @@
...
+	{
+		"type": "multi-select",
+		"id": "tags",
+		"title": "Tags",
+		"isGenre": true,
+		"canExclude": true,
+		"options": [
+			".crew",
```
kkantan: it looks like having this many tags causes the app to lag for a few seconds when loading filters. I believe the previous version of the nhentai source filtered out the tags it included in the filter by popularity, so only tags with a certain amount of linked titles were put in.

sources/multi.nhentai/src/lib.rs
kkantan, line 1: it looks like `cargo fmt` could be run on this file to correct some formatting

sources/multi.nhentai/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+			if needs_details {
+				manga.copy_from(gallery.clone().into_manga());
+			}
+
+			if needs_chapters {
+				// nhentai galleries are single chapter
+				let chapter = Chapter {
+					key: manga.key.clone(),
+					title: Some(format!("{} pages", gallery.num_pages)),
```
kkantan: I'm not sure about putting the page count as the chapter title. If anything, it should be included in the manga description instead.

sources/multi.nhentai/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+					chapter_number: Some(1.0),
+					url: Some(format!("{}/g/{}", BASE_URL, manga.key)),
+					..Default::default()
```
kkantan: it looks like the previous source also included the date for the chapter upload. is this possible to do?

sources/multi.nhentai/src/models.rs
```diff
@@ -0,0 +1,148 @@
...
+				"character" => characters.push(tag.name),
+				"language" => languages.push(tag.name),
+				_ => {}
+			}
+		}
+
+		let mut info_parts = Vec::new();
+
+		if !languages.is_empty() {
+			info_parts.push(format!("Languages: {}", languages.join(", ")));
```
kkantan: I'm not sure if this is necessary if you set the chapter language instead, given that the source should have a global language filter.

sources/multi.nhentai/src/models.rs
```diff
@@ -0,0 +1,148 @@
...
+				"group" => groups.push(tag.name),
+				"category" => categories.push(tag.name),
+				"parody" => parodies.push(tag.name),
+				"character" => characters.push(tag.name),
+				"language" => languages.push(tag.name),
+				_ => {}
+			}
+		}
+
+		let mut info_parts = Vec::new();
```
kkantan: the previous source included the id number in the description, and it looks like this is what the "description" is on the website as well, so that should probably also be put at the top.

sources/multi.nhentai/src/models.rs
```diff
@@ -0,0 +1,148 @@
...
+			info_parts.push(format!("Groups: {}", groups.join(", ")));
+		}
+		if !parodies.is_empty() {
+			info_parts.push(format!("Parodies: {}", parodies.join(", ")));
+		}
+		if !characters.is_empty() {
+			info_parts.push(format!("Characters: {}", characters.join(", ")));
+		}
+
+		let description = info_parts.join("<br>");
```
kkantan: is there a reason to use `<br>` here instead of an actual newline, `\n`?

sources/multi.nhentai/src/models.rs
```diff
@@ -0,0 +1,148 @@
...
+				Some(artists.clone())
+			},
+			artists: if artists.is_empty() {
+				if groups.is_empty() { None } else { Some(groups) }
+			} else {
+				Some(artists)
+			},
+			tags: Some(tags),
+			content_rating: ContentRating::NSFW,
+			..Default::default()
```
kkantan: the url for the manga page should also be set on the manga in addition to on the chapter.

sources/multi.nhentai/src/models.rs
```diff
@@ -0,0 +1,148 @@
...
+			authors: if artists.is_empty() {
+				if groups.is_empty() { None } else { Some(groups.clone()) }
+			} else {
+				Some(artists.clone())
+			},
+			artists: if artists.is_empty() {
+				if groups.is_empty() { None } else { Some(groups) }
+			} else {
+				Some(artists)
+			},
```
kkantan: should the authors be left blank, considering that the website only lists "artists"? I'm not sure if aidoku actually displays artists or not, though.

sources/multi.nhentai/res/filters.json
```diff
@@ -0,0 +1,8481 @@
...
+		"placeholder": "Scanlation group name"
+	},
+	{
+		"type": "sort",
+		"title": "Sort",
+		"options": [
+			"Popular Today",
+			"Popular Week",
+			"Popular All",
+			"Latest"
```
kkantan: it looks like the default sort for searching on the website is "Recent". should this be moved first?

kkantan (changes requested):

sources/multi.nhentai/Cargo.toml
dyphire: Done

sources/multi.nhentai/res/filters.json
dyphire: Done

sources/multi.nhentai/src/lib.rs
dyphire: Done

sources/multi.nhentai/src/lib.rs
dyphire: Add to description

dyphire: Done

sources/multi.nhentai/src/models.rs
dyphire: The `\n` generated description will be compressed into a line before it is expanded, and I don't like this effect very much. But try it first.

sources/multi.nhentai/src/lib.rs
dyphire: I think page count is a relatively important piece of information, and it's better to show it in the chapter title than in the description, because it often needs to be expanded to be seen.

sources/multi.nhentai/src/models.rs
dyphire: Aidoku doesn't actually show artists, and authors in nhentai aren't always is "artist" in namespace or maybe "group". I don't think there is a problem with the current processing.

Edit: respect the way files are named in nhentai: "group" before "artist". like https://nhentai.net/g/600908/

sources/multi.nhentai/res/filters.json
dyphire: The biggest feature of the new sources provided by Aidoku since version 0.7 is that the author and genre information supports click-and-jump, and if it is not fully supported, then the meaning of porting to the new implementation is lost.

If you've used the Aidoku's internal source komga, you'll notice that when there is a lot of metadata such as tags, genres, and authors, trying to open the filter will also lag for a few seconds. This is a performance issue with the current implementation of aidoku, and should be optimized by aidoku side rather than on the source.

**Edit:** After looking at the tags page on nhentai, I realized that it is possible to collect only tags with a popularity of 10 or more, which will greatly reduce the number of tags. Finished.

kkantan: you should add the `supportsTagSearch` flag to the source config, which will allow any tag to be tapped on regardless of whether they're contained in the static filters or not: https://github.com/Aidoku/aidoku-rs/blob/79e396c96d9b6beafe2b19d2aaf28be12f99ce8d/crates/cli/src/supporting/schema/source.schema.json#L107

sources/multi.nhentai/src/models.rs
kkantan: you're right, that's interesting. you should go back to `<br>`.

sources/multi.nhentai/src/models.rs
```diff
@@ -0,0 +1,170 @@
...
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct NhentaiImages {
+	pub pages: Vec<NhentaiImage>,
+	pub cover: NhentaiImage,
+	pub thumbnail: NhentaiImage,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct NhentaiGallery {
+	pub id: i32,
```
kkantan: I think this sometimes breaks, since the api returns some titles with a string id instead of a number. you can use the serde_json `Value` type instead.
```suggestion
	pub id: Value,
```
and above,
```rs
use serde_json::Value;
```

dyphire: Done

sources/multi.nhentai/res/filters.json
dyphire: Done

sources/multi.nhentai/src/models.rs
dyphire: Well, actually it's an Aidoku issue, I'll fix it over there.

dyphire: See https://github.com/Aidoku/Aidoku/pull/717

sources/multi.nhentai/res/filters.json
```diff
@@ -0,0 +1,1539 @@
+[
+	{
+		"type": "text",
+		"id": "author",
+		"title": "Artist",
+		"placeholder": "artist name"
```
kkantan:
```suggestion
		"placeholder": "Artist name"

```

dyphire: Done

kkantan (approved):

PR #125, opened 2025-10-07
feat(nhentai): use language as scanlater

dyphire:
The language option can be empty, so there is still a way to display the source language information

### DIFF
--- a/sources/multi.nhentai/res/source.json
+++ b/sources/multi.nhentai/res/source.json
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "multi.nhentai",
 		"name": "nhentai",
-		"version": 10,
+		"version": 11,
 		"url": "https://nhentai.net",
 		"contentRating": 2,
 		"languages": [
--- a/sources/multi.nhentai/src/lib.rs
+++ b/sources/multi.nhentai/src/lib.rs
@@ -133,12 +133,24 @@ impl Source for Nhentai {
 
 			if needs_chapters {
 				// nhentai galleries are single chapter
+				let mut languages = Vec::new();
+				for tag in &gallery.tags {
+					if tag.r#type == "language" && tag.name != "translated" && tag.name != "rewrite" {
+						languages.push(tag.name.clone());
+					}
+				}
+
 				let chapter = Chapter {
 					key: manga.key.clone(),
 					title: Some(format!("{} pages", gallery.num_pages)),
 					chapter_number: Some(1.0),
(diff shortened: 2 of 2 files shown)


sources/multi.nhentai/res/settings.json
```diff
@@ -13,5 +13,18 @@
 				"refreshes": ["content"]
 			}
 		]
+	},
+	{
+		"type": "group",
+		"title": "Blocklist",
+		"items": [
+			{
+				"type": "text",
```
kkantan: does it make more sense to use the `editable-list` setting type here instead? this is what the mangadex source uses for blocking scanlators.

dyphire: There's a bug here: https://github.com/Aidoku/Aidoku/issues/730

dyphire: This issue was fixed upstream, but a preemption value was retained for backward compatibility.

dyphire: I ran into another weird issue: https://github.com/Aidoku/Aidoku/issues/731, not sure if you can reproduce it.

sources/multi.nhentai/res/settings.json
```diff
@@ -13,5 +13,20 @@
...
+		"title": "Blocklist",
+		"items": [
+			{
+				"type": "editable-list",
+				"key": "blocklist",
+				"title": "Add Blockword",
+				"lineLimit": 1,
+				"inline": true,
+				"placeholder": "Tag to block",
+				"default": ["this is a example"]
```
kkantan: I think this sounds better, but otherwise looks good to me.
```suggestion
				"title": "Add Blocked Tag",
				"lineLimit": 1,
				"inline": true,
				"placeholder": "Tag to block",
				"default": ["example"]

```

dyphire: Done

Jamerrone:
> @Jamerrone I believe this PR should fix your issues, as it corrects some cases of tag filtering not working and multiple page results not loading. you could try the `.aix` file built by the workflow: https://github.com/Aidoku-Community/sources/actions/runs/18506425614/artifacts/4268696929, or wait until it's merged. feel free to create an issue if your problems aren't resolved.

Thanks for the reply, I tested it and nothing changed the source is still in a completely broken state were the homepage does not reflect the website at all and it still just shows the same series from like a month ago. Filtering and searching just results into the same series again. I will create an issue later tonight with a video.

PR #128, opened 2025-10-13
feat: add LANraragi

dyphire:


sources/multi.lanraragi/res/source.json
```diff
@@ -0,0 +1,15 @@
+{
+	"info": {
+		"id": "multi.lanraragi",
+		"name": "LANraragi",
+		"version": 1,
+		"url": "https://github.com/Difegue/LANraragi",
+		"contentRating": 2,
```
kkantan: since the source content is user-controlled, I think it makes more sense to mark it as safe rather than nsfw.
```suggestion
		"contentRating": 0,
```

sources/multi.lanraragi/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+	MangaStatus, Page, PageContent, Result, SelectFilter, Source,
+	alloc::{String, Vec},
+	helpers::uri::encode_uri,
+	imports::net::Request,
+	prelude::*,
+};
+use alloc::string::ToString;
+use alloc::vec;
+
+fn extract_date_added(tags: &str) -> Option<i64> {
```
kkantan: this function is only used in `get_manga_update`, correct? I think it could probably just be done inline rather than split into a separate function.

sources/multi.lanraragi/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+		if let Some(q) = query {
+			query_parts.push(q);
+		}
+
+		let mut sort_by = "date_added";
+		let mut order = "desc";
+
+		for filter in filters {
+			match filter {
+				FilterValue::Text { id: _, value } => query_parts.push(value),
```
kkantan: I think this syntax is more conventional
```suggestion
				FilterValue::Text { value, .. } => query_parts.push(value),
```

sources/multi.lanraragi/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+		}
+		if !query_parts.is_empty() {
+			let combined_query = query_parts.join(" ");
+			url.push_str(&format!("&filter={}", encode_uri(combined_query)));
+		}
+
+		if page > 1 {
+			let start = unsafe { CURRENT_START };
+			url.push_str(&format!("&start={}", start));
+		}
```
kkantan: the logic here should be simpler using `QueryParameters`. you shouldn't need to store mutable variables and manually encode values.
```rs
let mut qs = QueryParameters::new();
qs.push("sort_by", Some("date_added"));

for filter in filters {
	// example:
	qs.set("sort_by", Some("title"));
}

let url = format!("{base_url}/api/search?{qs}");
// ...

File: sources/multi.lanraragi/src/lib.rs [hunk 5]
```diff
@@ -0,0 +1,310 @@
...
+			let encoded_key = base64::Engine::encode(
+				&base64::engine::general_purpose::STANDARD,
+				api_key.as_bytes(),
+			);
+			request = request.header("Authorization", &format!("Bearer {}", encoded_key));
+		}
+		let data = request.data()?;
+
+		let search_result: SearchResult = serde_json::from_slice(&data)
+			.map_err(|e| aidoku::error!("JSON parse error for search API: {}", e))?;
```
kkantan: you should add the `json` feature to aidoku-rs and use the `get_json` or `json_owned` function instead. it looks like the latest aidoku-rs commits added the serde json error to the `AidokuError` result, so you shouldn't need to handle the logging yourself.

sources/multi.lanraragi/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+
+		let base_url = settings::get_base_url()?;
```
kkantan: you already got the `base_url` at the top of this function

sources/multi.lanraragi/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+}
+
+static mut CURRENT_START: i32 = 0;
+
+struct Lanraragi;
+
+impl Source for Lanraragi {
+	fn new() -> Self {
+		Self
+	}
```
kkantan: you shouldn't use static muts if possible. you can add a field to the source struct to use instead.

```suggestion
#[derive(Default)]
struct Lanraragi {
	current_start: RefCell<i32>,
}

impl Source for Lanraragi {
	fn new() -> Self {
		Self::default()
	}
```

sources/multi.lanraragi/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+			CURRENT_START += current_result_count;
+		}
+
+		Ok(MangaPageResult {
+			entries: search_result
+				.data
+				.into_iter()
+				.map(|archive| archive.into_manga(&base_url))
+				.collect(),
+			has_next_page: true,
```
kkantan: this will infinite loop if it reaches the end. is it not possible to see when you've fetched all results?

sources/multi.lanraragi/src/models.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct SearchResult {
+	pub data: Vec<Archive>,
+	pub draw: i32,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
```
kkantan: there are no `_`s, so these renames don't do anything, correct?

sources/multi.lanraragi/src/models.rs
```diff
@@ -0,0 +1,80 @@
...
+
+		let arcid = self.arcid.clone();
+		Manga {
+			key: self.arcid,
+			title: self.title,
+			cover: Some(format!("{}/api/archives/{}/thumbnail", base_url, arcid)),
+			url: Some(url),
+			tags: Some(tags),
+			status: MangaStatus::Completed,
+			content_rating: ContentRating::NSFW,
```
kkantan: likewise I don't think it makes sense to put a `content_rating` here.
```suggestion
```

sources/multi.lanraragi/res/source.json
dyphire: done

sources/multi.lanraragi/src/lib.rs
dyphire: done

sources/multi.lanraragi/src/lib.rs
dyphire: fix

sources/multi.lanraragi/src/lib.rs
dyphire: done

sources/multi.lanraragi/src/models.rs
dyphire: remove

sources/multi.lanraragi/src/models.rs
dyphire: fix

sources/multi.lanraragi/src/lib.rs
dyphire: done

sources/multi.lanraragi/src/lib.rs
dyphire: fix

sources/multi.lanraragi/res/source.json
kkantan: it looks like it was changed to 1 (17+) instead of 0

sources/multi.lanraragi/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+	BaseUrlProvider, Chapter, DynamicFilters, Filter, FilterValue, Manga, MangaPageResult,
+	MangaStatus, Page, PageContent, Result, SelectFilter, Source,
+	alloc::{String, Vec},
+	helpers::uri::QueryParameters,
+	imports::net::Request,
+	prelude::*,
+};
+use alloc::string::ToString;
+use alloc::vec;
+use core::cell::RefCell;
```
kkantan: you shouldn't need to import `alloc` as an extern crate since aidoku-rs re-exposes it. also, it's more conventional to define modules after importing.

```suggestion
use crate::models::{Archive, ArchiveMetadata, Category, SearchResult};
use aidoku::{
	BaseUrlProvider, Chapter, DynamicFilters, Filter, FilterValue, Manga, MangaPageResult,
	MangaStatus, Page, PageContent, Result, SelectFilter, Source,
	alloc::{String, Vec, string::ToString, vec},
	helpers::uri::QueryParameters,
	imports::net::Request,
	prelude::*,
};
use core::cell::RefCell;

mod models;
mod settings;
```

sources/multi.lanraragi/res/settings.json
```diff
@@ -0,0 +1,26 @@
...
+		"title": "Server settings",
+		"items": [
+			{
+				"type": "text",
+				"placeholder": "LANraragi Base Url",
+				"key": "baseUrl",
+				"keyboardType": 3,
+				"returnKeyType": 9,
+				"autocorrectionDisabled": true,
+				"refreshes": ["content"]
```
kkantan: since listings are also fetch from the source, they should probably be refreshed here as well
```suggestion
				"refreshes": ["content", "listings"]
```

sources/multi.lanraragi/res/settings.json
```diff
@@ -0,0 +1,26 @@
...
+		"type": "group",
+		"title": "Server settings",
+		"items": [
+			{
+				"type": "text",
+				"placeholder": "LANraragi Base Url",
+				"key": "baseUrl",
+				"keyboardType": 3,
+				"returnKeyType": 9,
+				"autocorrectionDisabled": true,
```
kkantan, line 13: the text fields should also disable autocapitalization since they're a url and case-sensitive key
```suggestion
				"autocorrectionDisabled": true,
				"autocapitalizationType": 0,
```

sources/multi.lanraragi/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+					None
+				}
+			});
+
+			let chapter = Chapter {
+				key: archive.arcid.clone(),
+				title: Some(format!("Read ({}p)", archive.pagecount)),
+				chapter_number: Some(1.0),
+				date_uploaded,
+				url: Some(format!("{}/api/archives/{}/files", base_url, archive.arcid)),
```
kkantan: this is a link to the json api, but it would probably be better to use the same link as the manga struct, so it's a link to the actual webpage where the chapter can be read.

sources/multi.lanraragi/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+			} else {
+				Some(url_tags.join("  \n"))
+			};
+
+			manga.cover = Some(format!(
+				"{}/api/archives/{}/thumbnail",
+				base_url, archive.arcid
+			));
+			manga.tags = Some(display_tags);
+			manga.status = MangaStatus::Completed;
```
kkantan: does it make sense to have the completed badge when the site doesn't list any publication status?

something else worth considering might be setting the update_strategy to never, since it's a single chapter source.

sources/multi.lanraragi/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+				if tag.starts_with("date_added:") {
+					tag.strip_prefix("date_added:")?.parse::<i64>().ok()
+				} else {
+					None
+				}
+			});
+
+			let chapter = Chapter {
+				key: archive.arcid.clone(),
+				title: Some(format!("Read ({}p)", archive.pagecount)),
```
kkantan: why is this title format used specifically? I don't see it anywhere on the web frontend. would it be better to just do the page count, `{} pages`?

sources/multi.lanraragi/src/settings.rs
```diff
@@ -0,0 +1,16 @@
+use aidoku::{AidokuError, alloc::string::String, imports::defaults::defaults_get, prelude::bail};
+
+const BASE_URL_KEY: &str = "baseUrl";
+const API_KEY_KEY: &str = "apiKey";
+
+pub fn get_base_url() -> Result<String, AidokuError> {
+	let base_url = defaults_get::<String>(BASE_URL_KEY);
+	match base_url {
+		Some(url) if !url.is_empty() => Ok(url),
+		_ => bail!("Base Url not configured - please set LANraragi server URL in settings"),
```
kkantan: this is a pretty long message. maybe something a little shorter would be better.
```suggestion
		_ => bail!("Missing base URL: configure in settings"),

```

sources/multi.lanraragi/src/models.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct SearchResult {
+	pub data: Vec<Archive>,
+	pub draw: i32,
+	#[serde(rename = "recordsFiltered")]
+    pub records_filtered: i32,
+    #[serde(rename = "recordsTotal")]
+    pub records_total: i32,
+}
```
kkantan, line 19: you can go back to using `#[serde(rename_all = "camelCase")]` since you have multiple fields following that convention

sources/multi.lanraragi/src/models.rs
```diff
@@ -0,0 +1,81 @@
+use aidoku::{
+	Manga, MangaStatus, Viewer,
+	alloc::{String, Vec, format},
+};
+use alloc::string::ToString;
```
kkantan:
```suggestion
use aidoku::{
	Manga, MangaStatus, Viewer,
	alloc::{String, Vec, format, string::ToString},
};
```

sources/multi.lanraragi/res/settings.json
```diff
@@ -0,0 +1,26 @@
+[
+	{
+		"type": "group",
+		"title": "Server settings",
+		"items": [
+			{
+				"type": "text",
+				"placeholder": "LANraragi Base Url",
```
kkantan: I think it goes without saying that the base url is for LANraragi, however we could also use the title key here for localization, and then have the demo url for a placeholder to align with komga.
```suggestion
				"title": "BASE_URL",
				"placeholder": "https://lrr.tvc-16.science",

```

sources/multi.lanraragi/res/settings.json
```diff
@@ -0,0 +1,26 @@
+[
+	{
+		"type": "group",
+		"title": "Server settings",
```
kkantan: this should either be "Server Settings" for title case capitalization, since it matters for iOS 26, or "SETTINGS" to get localization.

sources/multi.lanraragi/res/source.json
dyphire: LANraragi is mainly NSFW content in terms of design and actual usage, and I think the current contentRating is more appropriate.

sources/multi.lanraragi/src/lib.rs
dyphire: Done

sources/multi.lanraragi/src/lib.rs
dyphire: Done

sources/multi.lanraragi/src/models.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct SearchResult {
+	pub data: Vec<Archive>,
+	pub draw: i32,
+	#[serde(rename = "recordsFiltered")]
+    pub records_filtered: i32,
+    #[serde(rename = "recordsTotal")]
+    pub records_total: i32,
```
kkantan: since you have the `rename_all`, the individual `rename`s aren't necessary. it looks like there are also spaces here instead of tabs, make sure to run `cargo fmt` and it should fix everything.
```suggestion
	pub records_filtered: i32,
	pub records_total: i32,
```

dyphire: Fix

kkantan (approved):

PR #133, opened 2025-10-15
fix(en.mangapark) retrieved volume + chapter numbers

### LINKED ISSUE #106: MangaPark chapters misbehaving
If you migrate to MangaPark from a previously used source, your read chapter count will not migrate. The same happens if you move from MangaPark to another source.

When trying to select "Sync Local History" on the Tracker menu, it'll recognize that it needs to be updated to match, but won't match once you press Ok. 

Reading a chapter on MangaPark doesn't update your trackers.

Sheepodeepo:
Updates Volume + Chapter Numbers on most manga chapter formats.

close #106

### DIFF
--- a/sources/en.mangapark/src/helper.rs
+++ b/sources/en.mangapark/src/helper.rs
@@ -9,23 +9,23 @@ pub fn get_volume_and_chapter_number(
 	let tokens: Vec<&str> = title_input.split_whitespace().collect();
 	let mut i = 0;
 	while i < tokens.len() {
-			let token = tokens[i];
-			let token_lower = token.to_lowercase();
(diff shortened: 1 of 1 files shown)


Sheepodeepo:
I see, I didn't update the chapter title as well. Do we need that because I thought if we had the Chapter (and Volume Number if applicable) that we wouldn't need the title as it would just be displayed as something like Chapter 783.

kkantan:
chapter 783 in kingdom still doesn't have a chapter number for some reason, it's just the title "A Childish Play" now

kkantan (approved):
probably good enough

PR #135, opened 2025-10-16
feat: add zh.mangabz

dyphire:
This is a rewrite of the old source


sources/zh.mangabz/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+crate-type = ["cdylib"]
+
+[profile.dev]
+panic = "abort"
+
+[profile.release]
+panic = "abort"
+opt-level = "s"
+strip = true
+lto = true
```
kkantan: make sure there's an extra newline here for consistency with the other sources
```suggestion
lto = true

```

sources/zh.mangabz/res/source.json
```diff
@@ -0,0 +1,15 @@
+{
+	"info": {
+		"id": "zh.mangabz",
+		"name": "Māngabz",
+		"version": 5,
+		"urls": ["https://mangabz.com/"],
```
kkantan: if there's just one url I believe you just use the `url` field, and make sure there's no trailing slash.
```suggestion
		"url": "https://mangabz.com",
```

sources/zh.mangabz/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+		Ok(Request::get(url)?
+			.header("User-Agent", USER_AGENT)
+			.header("Referer", BASE_URL))
+	}
+}
+
+impl DeepLinkHandler for Mangabz {
+	fn handle_deep_link(&self, url: String) -> Result<Option<DeepLinkResult>> {
+		if url.contains("bz/") {
+			let _id = url
```
kkantan: for variables that are actually used, you shouldn't use the `_` prefix. you can call this `key` and then use the shorthand syntax for the struct initialization.

sources/zh.mangabz/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+				.replace("bz/", "")
+				.split('/')
+				.next_back()
+				.expect("manga id")
```
kkantan: avoid using `expect` unless you are absolutely certain it won't fail. you can use something like `.or_else(error!("invalid url"))?` instead.

sources/zh.mangabz/src/net/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		let mut genre = String::from("0");
+		let mut status = String::from("0");
+		let mut sort = String::from("10");
+
+		for filter in filters {
+			match filter {
+				FilterValue::Text { id, value } => {
+					let _ = id.as_str();
```
kkantan: why is this done? if the variable warns about being unused, you can just exclude it like `FilterValue::Text { value, .. }`.

sources/zh.mangabz/src/net/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+					return Ok(Self::Search {
+						query: encode_uri(value.clone()),
+						page,
+					});
+				},
```
kkantan: this `,` shouldn't be here; make sure to run `cargo fmt`.

sources/zh.mangabz/src/lib.rs
dyphire: Done

sources/zh.mangabz/src/lib.rs
dyphire: Done

sources/zh.mangabz/src/net/mod.rs
dyphire: Done

sources/zh.mangabz/src/net/mod.rs
dyphire: Done

kkantan (approved):

PR #136, opened 2025-10-16
feat: add zh.manhuagui

dyphire:
This is a rewrite of the old source


sources/zh.manhuagui/res/filters.json
kkantan, line 1: this json file should use tabs for indentation instead of spaces

sources/zh.manhuagui/res/source.json
```diff
@@ -0,0 +1,18 @@
+{
+	"info": {
+		"id": "zh.manhuagui",
+		"name": "Manhuagui",
+		"version": 6,
+		"urls": [
+			"https://www.manhuagui.com/", 
+			"https://tw.manhuagui.com/"
```
kkantan:
```suggestion
			"https://www.manhuagui.com", 
			"https://tw.manhuagui.com"
```

sources/zh.manhuagui/src/decoder.rs
```diff
@@ -0,0 +1,416 @@
+// use core::{iter::Map, ops::{Index, IndexMut}};
```
kkantan: you can remove these imports if they're not used

sources/zh.manhuagui/src/helper.rs
```diff
@@ -0,0 +1,45 @@
...
+			|| (curr == b'-'
+				|| curr == b'_'
+				|| curr == b'.'
+				|| curr == b'!'
+				|| curr == b'~'
+				|| curr == b'*'
+				|| curr == b'\''
+				|| curr == b'('
+				|| curr == b')')
+			|| (curr == b'#')
```
kkantan: it looks like you could just use the included `internal_encode_uri` aidoku-rs helper with a custom charset, since the logic is the same.

sources/zh.manhuagui/src/html/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+	Chapter, Manga, MangaPageResult, MangaStatus, Page, PageContent, Result, SelectFilter,
+	alloc::{String, Vec, borrow::Cow, vec},
+	imports::{html::Document, net::Request},
+	prelude::*,
+};
+use regex::Regex;
+
+fn extract_chapter_number(title: &str) -> Option<f32> {
+	let re = Regex::new(r"(?:第\s*)(\d+(?:\.\d+)?)|(\d+(?:\.\d+)?)\s*(?:话|話|章|回|卷|册|冊)")
+		.unwrap_or_else(|_| Regex::new("").unwrap());
```
kkantan: initializing an empty regex pattern seems weird to me. I think you could just do `.ok()?;` instead to return `None`, but I don't think the regex initialization should fail anyways if the regex pattern is valid.

sources/zh.manhuagui/src/html/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+				// Search page
+				(".cf > .book-cover > a", "#AspNetPagerResult > a")
+			} else {
+				// Home page
+				("#contList > li", "#AspNetPager1 > a")
+			};
+
+		let elements = self
+			.select(selector)
+			.ok_or_else(|| aidoku::error!("Failed to select manga list"))?;
```
kkantan: you shouldn't need to namespace any of these `error!` calls since `aidoku::prelude::*` is imported.
```suggestion
			.ok_or_else(|| error!("Failed to select manga list"))?;
```

dyphire: Done

sources/zh.manhuagui/src/html/mod.rs
dyphire: Done

sources/zh.manhuagui/src/helper.rs
dyphire: Done

sources/zh.manhuagui/res/filters.json
dyphire: Done

dyphire:
> the comments on the other pr mostly apply here as well: the `aidoku` dependency needs to be updated, and `cargo fmt` should be run.
> 
> I tried opening the first manga to test, `哥布林之血 哀叹的女剑士与血之谱系`, but it didn't load any chapters despite the site appearing to have one volume available. maybe due to an 18+ warning? the manga also doesn't have the nsfw content rating. I'm not sure if that's possible to detect or not, but it does look like the site has the information.

Fix

kkantan (approved):

PR #140, opened 2025-10-17
feat: add en.batcave

### LINKED ISSUE #74: batcave.biz
please add [batcave.biz](https://batcave.biz/)
thank you

w-sanches:
closes #74

First contact with Rust, let me know what I can improve.
Likely forgot to implement some features I don´t know about, but it should be enough to search and load comics from source.

### DIFF
--- a/sources/en.batcave/src/lib.rs
+++ b/sources/en.batcave/src/lib.rs
@@ -6,8 +6,8 @@ use aidoku::{
 		std::{parse_date, send_partial_result},
 	},
 	prelude::*,
-	AidokuError, Chapter, DeepLinkHandler, DeepLinkResult, FilterValue, Manga, MangaPageResult,
-	MangaStatus, Page, PageContent, Result, Source,
+	AidokuError, Chapter, DeepLinkHandler, DeepLinkResult, FilterValue, ImageRequestProvider,
+	Manga, MangaPageResult, MangaStatus, Page, PageContent, Result, Source,
 };
 use regex::Regex;
 use serde::Deserialize;
@@ -15,6 +15,7 @@ use serde::Deserialize;
 mod home;
 
 const BASE_URL: &str = "https://batcave.biz";
+const REFERER: &str = "https://batcave.biz/";
 
 struct BatCave;
 
@@ -247,9 +248,15 @@ impl Source for BatCave {
 						let pages = page_list
 							.images
 							.into_iter()
-							.map(|page_url| Page {
-								content: PageContent::url(page_url),
-								..Default::default()
+							.map(|mut page_url| {
+								if page_url.starts_with("/") {
+									page_url = format!("{BASE_URL}{}", page_url);
+								}
+
+								Page {
+									content: PageContent::url(page_url),
+									..Default::default()
+								}
 							})
 							.collect::<Vec<Page>>();
 
@@ -261,6 +268,20 @@ impl Source for BatCave {
 			.unwrap_or_default();
 
 		Ok(pages)
+	}
+}
+
+impl ImageRequestProvider for BatCave {
+	fn get_image_request(
+		&self,
+		url: String,
+		_context: Option<aidoku::PageContext>,
+	) -> Result<Request> {
+		if url.contains("batcave.biz") {
+			Ok(Request::get(url)?.header("Referer", REFERER))
+		} else {
+			Ok(Request::get(url)?)
+		}
 	}
 }
 
@@ -287,4 +308,4 @@ impl DeepLinkHandler for BatCave {
 	}
 }
 
-register_source!(BatCave, Home, DeepLinkHandler);
+register_source!(BatCave, Home, ImageRequestProvider, DeepLinkHandler);


sources/en.batcave/res/icon.png
kkantan, line 1: could we add a bit more padding to this logo? it looks like on the site header it has some.

sources/en.batcave/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "en.batcave",
+		"name": "BatCave",
+		"version": 1,
+		"url": "https://batcave.biz/",
```
kkantan: there shouldn't be a trailing slash here for the url
```suggestion
		"url": "https://batcave.biz",
```

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+            if needs_chapters {
+                send_partial_result(&manga);
+            }
+        }
+
+        if needs_chapters {
+            let chapter_list: ChapterList = serde_json::from_str(
+                html.select_first(".page__chapters-list>script")
+                    .and_then(|x| x.data())
+                    .expect("No script data")
```
kkantan: it's best to avoid panics, which happen with `expect(...)` and `unwrap()`, with aidoku sources since it breaks the source until the app is restarted. instead, you could use `.ok_or(error!("No script data"))?` to return the same error message instead, which is also displayed to the user.

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+
+                        let title = Some(chapter
+                            .title
+                            .strip_prefix(&manga.title)
+                            .map(str::trim)
+                            .map(String::from)
+                            .unwrap_or_else(|| chapter.title));
+
+                        let date_uploaded = Some(
+                            chrono::NaiveDate::parse_from_str(&chapter.date, "%-d.%-m.%Y")
```
kkantan: instead you could use the `parse_date` function included with aidoku-rs, but chrono is also okay. make sure to avoid the unwrap here as well though.

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+        let hot_releases_section = html
+            .select_first("main section.sect--hot")
+            .expect("No hot release section");
+
+        let title = hot_releases_section
+            .select_first("div")
+            .and_then(|x| x.text());
+
+        let hot_releases = hot_releases_section
+            .select("div>a")
```
kkantan: just for consistency
```suggestion
            .select("div > a")
```

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+                                .unwrap()
+                                .and_time(chrono::NaiveTime::MIN)
+                                .and_utc()
+                                .timestamp(),
+                        );
+
+                        Chapter {
+                            key: url.clone(),
+                            url: Some(url),
+                            title,
```
kkantan: could we obtain the chapter number from the title? it would allow people to use trackers and such with the source. it looks like most of the titles are formatted `Issue #X`, so you could just parse the number after the `#` if it exists. I believe the readcomiconline source should have the same logic which would work here.

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+            manga.status = match status_str
+                .strip_prefix("Release type: ")
+                .unwrap_or_default()
+            {
+                "Completed" => MangaStatus::Completed,
+                "Complete" => MangaStatus::Completed,
+                "Ongoing" => MangaStatus::Ongoing,
+                _ => MangaStatus::Unknown,
+            };
+
```
kkantan: it looks like some entries also have genres/tags you could parse as well: https://batcave.biz/genres/Action/

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+        Ok(HomeLayout {
+            components: vec![HomeComponent {
+                title,
+                subtitle: None,
+                value: aidoku::HomeComponentValue::Scroller {
+                    entries: hot_releases,
+                    listing: None,
+                },
+            }],
```
kkantan: this is a pretty empty home page with only one section, so I'm not sure if it makes too much sense to include it. could we maybe parse some of the other sections too, like `The newest comic releases` and `Top-rated comics`? usually our goal with the home page is to try and reflect the look of the (mobile) site.

kkantan (changes requested):
thank you for the pr! it looks pretty good, I just left some comments on a few things. most notably, make sure to avoid unwrap and expect. also it looks like clippy left some warnings, so make sure to run `cargo clippy` locally and fix anything it points out.

it looks like the site also has a "Currently reading" page that might be worthwhile to implement as a listing. if you're up for it, you could also implement the filtering, e.g. by genre, on the catalogue page, and then use the `hidesFiltersWhenSearching` config option since it looks like search isn't supported in addition to any filters. it's not strictly necessary though.

sources/en.batcave/src/lib.rs
w-sanches: It worked great. Just found a single weird case within X-Men (1991):
From site's chapter list:

```
X-Men (1991) Issue #-1
X-Men (1991) Issue #1
X-Men (1991) Issue #0.5
X-Men (1991) Annual 2000
X-Men (1991) Annual 1999
```

For some reason there is a `-1` issue which ends up placed before Annual issues, which get `None` as chapter number if I understood the code.

Not sure how to handle it anyway, so I am keeping same approach for now.

sources/en.batcave/res/icon.png
w-sanches: Done in 685e75c
I did not notice it was not part of the SVG they are using for the logo.

sources/en.batcave/src/lib.rs
w-sanches: Done in dee8291

sources/en.batcave/src/lib.rs
w-sanches: Done in 6468eb6

sources/en.batcave/src/lib.rs
w-sanches: Done in e5d710d

sources/en.batcave/res/source.json
w-sanches: Done in ca82a45

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+
+			manga.cover = html
+				.select_first(".page__poster img")
+				.and_then(|x| x.attr("abs:src"));
+
+			manga.artists = html
+				.select_first("ul > li:has(div:contains(Artist))")
+				.and_then(|x| x.text())
+				.map(|x| x.strip_prefix("Artist: ").unwrap_or_default().to_string())
+				.map(|x| vec![x.to_string()]);
```
kkantan: it looks like you're `to_string`ing twice here, so `x` should already be a `String`, right?
```suggestion
				.map(|x| x.strip_prefix("Artist: ").unwrap_or_default().to_string())
				.map(|x| vec![x]);
```

sources/en.batcave/src/lib.rs
w-sanches: Included more sections in 80cbfbd 

Quite challenging, very fun :D

sources/en.batcave/src/lib.rs
w-sanches: Yes, that is correct, removed in 7eafb14

At first I thought I would be able to split this into various artists, but all examples I found used `Various` when multiple artists or writes were included.

sources/en.batcave/src/lib.rs
w-sanches: By the way, is there any tip on developing sources?

So far I've been doing:
- make changes to source;
- package;
- serve with `aidoku serve package.aix`
- go to my phone;
- remove source and add again;
- check for changes;

I feel like there should be a way to update instead remove+add, but if that is the case I could not figure it out

sources/en.batcave/src/lib.rs
w-sanches: Remove remaining expects and unwraps in f0a1b69

Not sure my error handling is the best though, I did not get 100% how these errors should be returned to be properly formatted in the UI.

w-sanches:
> it looks like the site also has a "Currently reading" page that might be worthwhile to implement as a listing.

What exactly is a listing?

I see many home components receive a listing parameter, but the few examples I saw pass `None` to those and it works fine so far.
There are also sources which do implement a `ListingProvider`, but I also could not graps what it means nor when it should be used. Again, my source worked fine without it so I simply skipped.

Even aidoku-rs source code did not make it much more clear to me, maybe I missed some important file.

sources/en.batcave/src/lib.rs
w-sanches: Parsing tags in d3f8184

I think it requires extra work somewhere because my tags are not links which would show a list of genres/X comics and I am not sure what is missing in my source to make it work.

sources/en.batcave/src/lib.rs
```diff
@@ -132,50 +140,53 @@ impl Source for BatCave {
...
+			let script_data = html
+				.select_first(".page__chapters-list > script")
+				.and_then(|x| x.data())
+				.ok_or(error!("No script data"))?;
+
+			let json_str = script_data
+				.strip_prefix("window.__DATA__ = ")
+				.ok_or(error!("Wrong script format"))?
+				.strip_suffix(";")
+				.ok_or(error!("Wrong script format"))?;
```
kkantan: this can be simplified a little bit
```suggestion
				.strip_prefix("window.__DATA__ = ")
				.and_then(|d| d.strip_suffix(";"))
				.ok_or(error!("Wrong script format"))?;
```

sources/en.batcave/src/lib.rs
```diff
@@ -132,50 +140,53 @@ impl Source for BatCave {
...
+						.strip_prefix(&manga.title)
+						.map(str::trim)
+						.map(String::from)
+						.unwrap_or_else(|| chapter.title);
+
+					let chapter_number = title
+						.find('#')
+						.and_then(|idx| title[idx + 1..].parse::<f32>().ok());
+
+					let date_uploaded = parse_date(&chapter.date, "%-d.%-m.%Y");
```
kkantan: are we using an older version of aidoku-rs here? I believe the date format string should use the [DateFormatter](https://www.nsdateformatter.com/) format, so something like `dd.MM.yyyy`.

sources/en.batcave/src/lib.rs
```diff
@@ -249,29 +388,51 @@ impl Home for BatCave {
...
 		})
+	})
+}
+
+impl Home for BatCave {
+	fn get_home(&self) -> Result<HomeLayout> {
+		let html = Request::get(BASE_URL)?.html()?;
+		let mut components = Vec::new();
+
+		let component_fns: Vec<ComponentBuilderFn> = vec![
```
kkantan: I don't think you need an allocated vector here, you should be able to get away with just a slice.
```suggestion
		let component_fns: &[ComponentBuilderFn; 4] = &[
```

sources/en.batcave/src/lib.rs
kkantan: to do that you can add a `select` type filter in the `filters.json` file with the `isGenre` tag set to true and then handle it in the `get_search_manga_list` function.

sources/en.batcave/src/lib.rs
kkantan: home page looks good!

what I personally do is run aidoku though xcode so I can do all development on my mac without having to switch devices. I drag and drop the aix file onto the simulator to install it quickly. if you're using `aidoku serve`, you could try bumping the version every time you make a change and rebuild, and then you should be able to pull to refresh the browse tab and update the source.

kkantan (changes requested):
> What exactly is a listing?

if you look at other sources, they're the "tabs" in the source you can switch to. for example, in the mangadex source you can see the "Popular" and "Recently Added" tabs, but you can also see that some of the home components, such as "Latest Updates" have a link to a page with more items on it. that's what you get if you specify a listing for a `HomeComponent`.

to add listings to your source you can add them in the `source.json` file, then implement the `ListingProvider` trait.

the last thing I would suggest is to try implementing the `DeepLinkHandler` trait, since I believe every other source on here has it. all you need to do is parse the manga key from the url that's given, if it's a valid one. to test, you can go to the website in safari and just swap the `https://` with `aidoku://`, and aidoku should open and automatically go to the correct series.

sources/en.batcave/src/lib.rs
w-sanches: You are right, this works, did not know it was possible, done in f217a70

sources/en.batcave/src/lib.rs
w-sanches: Got it.
Not using a Mac for development right now, so bumping version works better for me, ty :)
Some bash scripting and I got version bump on every repackaging.

sources/en.batcave/src/lib.rs
w-sanches: I was already parsing page lists with a simpler approach, changed here too in 7965d6f

sources/en.batcave/src/lib.rs
w-sanches: Interesting, I assumed it was working since I got no errors, but after updating the format string I see it never properly worked before :man_facepalming: 

Fixed in 3d61ba3

sources/en.batcave/src/lib.rs
w-sanches: Website is a bit all over the place on genre handling.
If you go to a specific comic and click its genre, you end up in `https://batcave.biz/genres/Action/`, but you can also multi-select genres in `https://batcave.biz/comix/`, which actually redirects you to `https://batcave.biz/ComicList/g=1,2/`.

Added genre filtering the best I could in c25839b

Just noticed you can also sort by some other fields like `date` and `title`, but testing them I don´t get how it is actually being sorted because I only get new comics and nothing from 1990s.
Decided to simply skip those for now.

I will add year of issue filtering too in a following commit as that looks useful and works properly.

sources/en.batcave/res/filters.json
kkantan, line 1: this file should use tags for indentation

sources/en.batcave/res/filters.json
```diff
@@ -0,0 +1,117 @@
+[
+  {
+		"id": "year_of_issue",
+		"title": "Yearn of Issue",
```
kkantan:
```suggestion
		"title": "Year of Issue",
```

sources/en.batcave/res/source.json
```diff
@@ -0,0 +1,15 @@
+{
+	"info": {
+		"id": "en.batcave",
+		"name": "BatCave",
+		"version": 27,
```
kkantan: this should still be v1 when adding to the main repo
```suggestion
		"version": 1,
```

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+				query.unwrap_or_default()
+			)
+		} else {
+			format!(
+				"{BASE_URL}/ComicList/{}/page/{page}/",
+				filters_vec.join("/")
+			)
+		};
+
+		println!("{}", url);
```
kkantan: this should be removed for public builds

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+					.filter_map(|element| {
+						let cover = element.select_first("img")?.attr("abs:data-src");
+						let url = element.select_first("a")?.attr("abs:href");
+						let title = element
+							.select_first("div > h2")
+							.and_then(|x| x.text())
+							.unwrap_or_default();
+
+						Some(Manga {
+							key: url.clone().unwrap_or_default(),
```
kkantan: we don't want to end up with items with invalid keys, so you can use `?` to this to filter it out.

however, it's best not to use entire urls as the manga key, since if the domain ever changes it will break everything. you can do `url.strip_prefix(BASE_URL).into()` to get the path, and then in `get_manga_update` you can construct the url like `format!("{BASE_URL}{}", manga.key)`.

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+			manga.description = html.select_first(".page__text").and_then(|x| x.text());
+
+			manga.cover = html
+				.select_first(".page__poster img")
+				.and_then(|x| x.attr("abs:src"));
+
+			manga.artists = html
+				.select_first("ul > li:has(div:contains(Artist))")
+				.and_then(|x| x.text())
+				.map(|x| x.strip_prefix("Artist: ").unwrap_or_default().to_string())
```
kkantan: so you don't end up with empty strings:
```suggestion
				.and_then(|x| x.strip_prefix("Artist: ").map(|s| s.to_string()))
```

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+				.ok_or(error!("No script data"))?;
+
+			let json_str = script_data
+				.strip_prefix("window.__DATA__ = ")
+				.and_then(|x| x.strip_suffix(";"))
+				.unwrap_or_default();
+
+			let chapter_list = match serde_json::from_str::<ChapterList>(json_str) {
+				Ok(list) => list,
+				Err(_) => return Err(error!("Unable to parse chapter list")),
```
kkantan: there's a shorthand function for this, `bail!`, but instead of `match` you could use `.map_err(|e| AidokuError::JsonParseError(e))?` on the result I think.

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+						.unwrap_or_default();
+
+					let cover = element
+						.select_first("img")
+						.and_then(|x| x.attr("abs:data-src"));
+
+					let url = element.attr("abs:href");
+
+					Manga {
+						key: url.clone().unwrap_or_default(),
```
kkantan: you should use `filter_map` and `?`

sources/en.batcave/src/lib.rs
```diff
@@ -0,0 +1,502 @@
...
+		value: aidoku::HomeComponentValue::MangaChapterList {
+			page_size: Some(6),
+			entries,
+			listing: None,
+		},
+		..Default::default()
+	})
+}
+
+fn get_side_block(index: i32) -> ComponentBuilderFn {
```
kkantan: I think you should put all of these home-related functions inside of the `get_home` function, then move the entire `Home` implementation to a separate `home.rs` file. there are other sources that do this, like mangadex.

kkantan (changes requested):
there are also a few clippy errors you can see with `cargo clippy`

sources/en.batcave/src/lib.rs
w-sanches: Moved in [a161dbb](https://github.com/Aidoku-Community/sources/pull/140/commits/a161dbb6af8754135a9d00a977935da13c449b74)

Mangadex had me learning a lot of new things :eyes:

sources/en.batcave/res/filters.json
w-sanches: I think you meant tabs? Done in [d72ea3c](https://github.com/Aidoku-Community/sources/pull/140/commits/d72ea3c3388f30337fdfe868255531f3ba73f3c9)

sources/en.batcave/res/filters.json
w-sanches: Fixed typo in [7fe2cbe](https://github.com/Aidoku-Community/sources/pull/140/commits/7fe2cbe613c0a610eb604081e28c61d8d092249a)

sources/en.batcave/res/source.json
w-sanches: I think I did a `git add .` before turning off computer and committed unwanted changes

[a898067](https://github.com/Aidoku-Community/sources/pull/140/commits/a89806732ae0c83b208177123d2b08505e47d7b0)

sources/en.batcave/src/lib.rs
w-sanches: I think I did a `git add .` before turning off computer and committed unwanted changes

[a898067](https://github.com/Aidoku-Community/sources/pull/140/commits/a89806732ae0c83b208177123d2b08505e47d7b0)

sources/en.batcave/src/lib.rs
w-sanches: Done in [bad3514](https://github.com/Aidoku-Community/sources/pull/140/commits/bad35144ca73d78987ff540cbafbf9aeb1b0e9c3)

Same with writers

sources/en.batcave/src/lib.rs
w-sanches: Changed in [8afae88](https://github.com/Aidoku-Community/sources/pull/140/commits/8afae88a7616ea6d09486596969047a25d862773)

Also touched page_list decoding to be with a similar idiom

sources/en.batcave/src/lib.rs
w-sanches: Should be handled in [4e6754d](https://github.com/Aidoku-Community/sources/pull/140/commits/4e6754d19410012c2c53d8ed0966017978174dbc)

sources/en.batcave/src/lib.rs
w-sanches: Should be handled in [4e6754d](https://github.com/Aidoku-Community/sources/pull/140/commits/4e6754d19410012c2c53d8ed0966017978174dbc)

kkantan:
it looks like chapter 100 of 100 bullets and the 1984 comic, among others, no longer load, and don't have urls. I think you're setting the url as the path now instead of the full url with the BASE_URL attached.

w-sanches:
Their image data is a bit weird...

For some comics they do require Referer header, while others in the same domain do not, and there are even those comics which are hosted in a different domain.
Also I found some cases like "Mother Teresa of Calcutta" in which their images are indeed just a path...

I think I got they all working now, but there might be some other case I don´t know about yet.

PR #142, opened 2025-10-18
feat: add zh.bilimanga

dyphire:


sources/zh.bilimanga/src/html/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+		};
+		manga.url = Some(Url::manga(manga.key.clone()).to_string());
+		Ok(())
+	}
+	fn manga_page_result(&self) -> Result<MangaPageResult> {
+		let mut entries: Vec<Manga> = Vec::new();
+
+		let alternate_url = self
+			.select("link[rel='alternate']")
+			.and_then(|link| link.first())
```
kkantan: I don't think there's any reason not to use `select_first` for these

dyphire: Done

kkantan (approved):

PR #144, opened 2025-10-19
feat: add zh.godamanga

dyphire:


sources/zh.godamanga/res/filters.json
```diff
@@ -0,0 +1,88 @@
...
+			"dushi",
+			"yixing",
+			"qingchun",
+			"mori",
+			"xuanyi",
+			"xiuxian",
+			"zhandou"
+		]
+	}
+]
```
kkantan:
```suggestion
]

```

sources/zh.godamanga/src/html/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+pub trait MangaPage {
+	fn update_details(&self, manga: &mut Manga) -> Result<()>;
+	fn manga_page_result(&self) -> Result<MangaPageResult>;
+}
+
+impl MangaPage for Document {
+	fn update_details(&self, manga: &mut Manga) -> Result<()> {
+		let ids = manga.key.split("/").collect::<Vec<&str>>();
+		let url = format!("{}/manga/{}", BASE_URL, ids[0]);
+		let html = Request::new(url.clone(), HttpMethod::Get)?
```
kkantan:
```suggestion
		let html = Request::get(url.clone())?
```

sources/zh.godamanga/src/html/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+		manga.url = Some(url);
+
+		Ok(())
+	}
+
+	fn manga_page_result(&self) -> Result<MangaPageResult> {
+		let mut entries: Vec<Manga> = Vec::new();
+
+		for item in self.try_select(".pb-2>a")? {
+			let href = item.attr("href").unwrap_or_else(String::new);
```
kkantan: this should do the same thing but is more conventional
```suggestion
			let href = item.attr("href").unwrap_or_default();
```

sources/zh.godamanga/src/json/chapter_list/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+		let ids = manga_id.split("/").collect::<Vec<&str>>();
+		let url = format!("{}/api/manga/get?mid={}&mode=all", API_URL, ids[1]);
+		let json: serde_json::Value = Request::new(url.clone(), HttpMethod::Get)?
+			.header("Origin", BASE_URL)
+			.header("Referer", BASE_URL)
+			.send()?
+			.get_json()?;
+		let data = json
+			.as_object()
+			.ok_or_else(|| AidokuError::message("Expected JSON object"))?;
```
kkantan:
```suggestion
			.ok_or_else(|| error!("Expected JSON object"))?;
```

sources/zh.godamanga/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+pub const API_URL: &str = "https://api-get-v2.mgsearcher.com";
+pub const IMG_URL: &str = "https://f40-1-4.g-mh.online";
+
+fn handle_cover_url(url: String) -> String {
+	if url.contains("url=") {
+		url.split("url=")
+			.map(|a| a.to_string())
+			.collect::<Vec<String>>()
+			.pop()
+			.unwrap()
```
kkantan: you shouldn't need to do all this. I think `last()` should work on the split iterator, and then you could map it and `unwrap_or(url)`.

kkantan (changes requested):
I only commented on the first instance of each of the things, but the suggestions should be applied everywhere.

it looks like most of the items on the source are manhwa/manhua, so it might be worth defaulting the manga viewer type to `Webtoon`, if you can't reliably parse it from the site info.

sources/zh.godamanga/src/html/mod.rs
dyphire: Done

sources/zh.godamanga/src/json/chapter_list/mod.rs
dyphire: Done

sources/zh.godamanga/src/lib.rs
dyphire: Done

sources/zh.godamanga/src/html/mod.rs
dyphire: Done

sources/zh.godamanga/src/json/page_list.rs
```diff
@@ -0,0 +1,65 @@
...
+			.and_then(|v| v.as_array())
+			.ok_or_else(|| error!("Expected images array"))?;
+		let mut pages: Vec<Page> = Vec::new();
+
+		for (index, item) in list.iter().enumerate() {
+			let item = match item.as_object() {
+				Some(item) => item,
+				None => continue,
+			};
+			let _index = index as i32;
```
kkantan: you shouldn't need to `enumerate()` the list if this isn't used

kkantan (changes requested):

sources/zh.godamanga/src/json/page_list.rs
dyphire: Done

sources/zh.godamanga/src/html/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+pub trait MangaPage {
+	fn update_details(&self, manga: &mut Manga) -> Result<()>;
+	fn manga_page_result(&self) -> Result<MangaPageResult>;
+}
+
+impl MangaPage for Document {
+	fn update_details(&self, manga: &mut Manga) -> Result<()> {
+		let ids = manga.key.split("/").collect::<Vec<&str>>();
+		let url = format!("{}/manga/{}", BASE_URL, ids[0]);
+		let html = Request::get(url.clone())?
```
kkantan: url doesn't need to be cloned
```suggestion
		let html = Request::get(&url)?
```

sources/zh.godamanga/src/html/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+								.replace("#", "")
+								.replace("热门推荐", "")
+								.trim()
+								.to_string()
+						})
+					})
+					.filter(|a| !a.is_empty())
+					.collect::<Vec<String>>()
+			})
+			.unwrap_or_else(Vec::new);
```
kkantan: you don't need to unwrap this since you're using it like `Some(categories)` below. same with the description.

sources/zh.godamanga/src/html/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+	fn manga_page_result(&self) -> Result<MangaPageResult> {
+		let mut entries: Vec<Manga> = Vec::new();
+
+		for item in self.try_select(".pb-2>a")? {
+			let href = item.attr("href").unwrap_or_default();
+			let id = href
+				.split("/")
+				.filter(|s| !s.is_empty())
+				.last()
+				.unwrap_or("")
```
kkantan: you shouldn't need to do this. you could `unwrap_or_default`, but what situation would you want an empty id? you should use `filter_map`

sources/zh.godamanga/src/html/mod.rs
dyphire: Done

sources/zh.godamanga/src/html/mod.rs
dyphire: Done

kkantan (approved):

PR #145, opened 2025-10-19
feat: add zh.happymh

dyphire:


dyphire:
> same comments as other pr

Tried implementing all the suggestions, except for the suggestion to default the manga viewer type to Webtoon all, which I don't really like.

sources/zh.happymh/src/html/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+		let description = html
+			.select_first("#showmore")
+			.and_then(|e| e.text())
+			.map(|t| t.trim().to_string())
+			.unwrap_or_default();
+		let categories = html
+			.select(".mg-cate>a")
+			.map(|elements| elements.filter_map(|a| a.text()).collect::<Vec<String>>())
+			.unwrap_or_else(Vec::new);
+		let status = MangaStatus::Unknown;
```
kkantan: you don't parse the status so you don't need to set it

sources/zh.happymh/src/html/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+			.unwrap_or_default();
+		let description = html
+			.select_first("#showmore")
+			.and_then(|e| e.text())
+			.map(|t| t.trim().to_string())
+			.unwrap_or_default();
+		let categories = html
+			.select(".mg-cate>a")
+			.map(|elements| elements.filter_map(|a| a.text()).collect::<Vec<String>>())
+			.unwrap_or_else(Vec::new);
```
kkantan:
```suggestion
			.unwrap_or_default();
```

sources/zh.happymh/src/html/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+				.and_then(|e| e.attr("src"))
+				.map(handle_cover_url)
+				.unwrap_or_default();
+			let title = item
+				.select_first(".manga-title")
+				.and_then(|e| e.text())
+				.map(|t| t.trim().to_string())
+				.unwrap_or_default();
+
+			if !id.is_empty() && !title.is_empty() {
```
kkantan: instead of checking if they're empty you should use `Option<String>` and check if they're `Some` to avoid extra allocations. this is easier if you use `filter_map` on the elements and collect into a `Vec`, since then you can use `?` on `Option` values to skip `None`s.

sources/zh.happymh/src/json/chapter_list/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+					.unwrap_or_default()
+					.to_string();
+				let title = item
+					.get("chapterName")
+					.and_then(|v| v.as_str())
+					.unwrap_or_default()
+					.to_string();
+				let chapter_or_volume =
+					extract_chapter_number(&title).unwrap_or((all_chapters.len() + 1) as f32);
+				let url = format!("{}/mangaread/{}", BASE_URL, id.clone());
```
kkantan: you shouldn't need to clone the `id` here.

sources/zh.happymh/src/json/chapter_list/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+				let scanlator = if vo > -1.0 {
+					"单行本".to_string()
+				} else {
+					"默认".to_string()
+				};
+
+				all_chapters.push(aidoku::Chapter {
+					key: id,
+					title: Some(title),
+					volume_number: if vo >= 0.0 { Some(vo) } else { None },
```
kkantan: you could use `then_some` like your other sources.

sources/zh.happymh/src/json/page_list.rs
```diff
@@ -0,0 +1,59 @@
...
+};
+
+pub struct PageList;
+
+impl PageList {
+	pub fn get_pages(_manga_id: String, chapter_id: String) -> Result<Vec<Page>> {
+		let url = format!(
+			"{}/v2.0/apis/manga/reading?code={}&v=v3.1818134",
+			BASE_URL,
+			chapter_id.clone()
```
kkantan: no need to clone anything in `format!`.

sources/zh.happymh/src/json/page_list.rs
```diff
@@ -0,0 +1,59 @@
...
+				Some(item) => item,
+				None => continue,
+			};
+			let url = item
+				.get("url")
+				.and_then(|v| v.as_str())
+				.unwrap_or_default()
+				.to_string();
+			pages.push(Page {
+				content: aidoku::PageContent::Url(url, None),
```
kkantan:
```suggestion
				content: aidoku::PageContent::url(url),
```

sources/zh.happymh/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+	ListingProvider, Manga, MangaPageResult, Page, Result, Source,
+};
+use html::MangaPage as _;
+use net::Url;
+
+pub const BASE_URL: &str = "https://m.happymh.com";
+
+fn handle_cover_url(url: String) -> String {
+	url
+}
```
kkantan: you can remove this.

sources/zh.happymh/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+				cover: Some(cover),
+				title,
+				..Default::default()
+			});
+		}
+
+		Ok(MangaPageResult {
+			entries: mangas,
+			// There is no page information in the response, assuming it is always true
+			// The API returns null results for non-existent page requests
```
kkantan: you should probably check that and return an empty result with `has_next_page` set to `false`, since otherwise if a user reaches the end it will throw an error, which hides the entire manga list

sources/zh.happymh/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+		let url = match listing.id.as_str() {
+			"day" => format!("{}/rank/day?page={}", BASE_URL, page),
+			"week" => format!("{}/rank/week?page={}", BASE_URL, page),
+			"month" => format!("{}/rank/month?page={}", BASE_URL, page),
+			"dayBookcasesOne" => format!("{}/rank/dayBookcasesOne?page={}", BASE_URL, page),
+			"weekBookcasesOne" => format!("{}/rank/weekBookcasesOne?page={}", BASE_URL, page),
+			"monthBookcasesOne" => format!("{}/rank/monthBookcasesOne?page={}", BASE_URL, page),
+			"voteNumMonthRank" => format!("{}/rank/voteNumMonthRank?page={}", BASE_URL, page),
+			"voteRank" => format!("{}/rank/voteRank?page={}", BASE_URL, page),
+			_ => return self.get_search_manga_list(None, page, Vec::new()),
```
kkantan: I think it makes more sense to return an error, even though this shouldn't be hit during normal usage.
```suggestion
			_ => bail!("Invalid listing"),
```

sources/zh.happymh/src/html/mod.rs
dyphire: Done

sources/zh.happymh/src/lib.rs
dyphire: Done

sources/zh.happymh/src/lib.rs
dyphire: Done

sources/zh.happymh/src/lib.rs
dyphire: This is actually used for the "latest" listing, let me fix it.

kkantan (approved):

PR #147, opened 2025-10-20
feat: set viewer mode for zh.bilimanga, zh.manhuagui, zh.mycomic

dyphire:


sources/zh.bilimanga/src/html/mod.rs
```diff
@@ -98,7 +98,7 @@ impl MangaPage for Document {
 			.try_select(".tag-small-group>.tag-small>a")?
 			.filter_map(|a| a.text())
 			.collect::<Vec<String>>();
-		manga.tags = Some(tags);
+		manga.tags = Some(tags.clone());
```
kkantan: ideally you shouldn't need to clone the entire vector of tags.
for example, if you've already assigned `manga.tags`, you can get them with:
```rs
let tags = manga.tags.as_deref().unwrap_or(&[]);
manga.viewer = if tags
	.iter()
	.any(|e| matches!(e.as_str(), "大陸" | "韓國"))
{
	Viewer::Webtoon
} else {
	...
};
```

dyphire: Done

kkantan (approved):

PR #148, opened 2025-10-20
feat: rewrite zh.baozimanhua with new api

### LINKED ISSUE #28: Mandarin Chinese manhua source
https://www.baozimh.com/

https://m.happymh.com/

I noticed that there aren’t a lot of Mandarin Chinese sources, and for those who want to read - like myself - we aren’t always with Wi-Fi, so I think it would be a nice addition to have it added, given with the download option that comes with Aidoku.

Thanks

dyphire:
Close #28


sources/zh.baozimanhua/res/source.json
```diff
@@ -0,0 +1,29 @@
+{
+	"info": {
+		"id": "zh.baozimanhua",
+		"name": "包子漫畫",
+		"version": 4,
+		"url": "https://www.baozimh.com",
+		"urls": [
```
kkantan: you shouldn't need both fields. I think it defaults the main url to the first one in the `urls` field, and since you already have it included in both, you can remove the `url` one.

sources/zh.baozimanhua/src/html/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+			let key = url
+				.split('/')
+				.filter(|s| !s.is_empty())
+				.next_back()
+				.unwrap_or("")
+				.to_string();
+
+			if key.is_empty() {
+				continue;
+			}
```
kkantan:
```suggestion
			let Some(key) = url
				.split('/')
				.filter(|s| !s.is_empty())
				.next_back()
				.map(|s| s.to_string())
			else {
				continue;
			};
```

sources/zh.baozimanhua/src/html/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	fn manga_page_result(&self) -> Result<MangaPageResult> {
+		let mut entries: Vec<Manga> = Vec::new();
+
+		for item in self.try_select("div.comics-card")? {
+			let url = item
+				.select_first("a.comics-card__poster")
+				.and_then(|a| a.attr("href"))
+				.unwrap_or_else(String::new);
```
kkantan:
```suggestion
				.unwrap_or_default();
```

sources/zh.baozimanhua/src/html/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+impl ChapterPage for Document {
+	fn chapters(&self, _manga_id: &str) -> Result<Vec<aidoku::Chapter>> {
+		let mut chapters: Vec<aidoku::Chapter> = Vec::new();
+
+		// Try to get chapters from the main list, fallback to alternative selector
+		let chapter_items = self
+			.try_select("div.pure-g[id] a.comics-chapters__item")
+			.or_else(|_| self.try_select("a.comics-chapters__item"))?;
```
kkantan: I don't think this fallback does anything. the `try` for the select should only fail if the selector is invalid, which in this case it is not.

sources/zh.baozimanhua/src/html/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+impl PageList for Document {
+	fn pages(&self) -> Result<Vec<Page>> {
+		let mut pages: Vec<Page> = Vec::new();
+
+		for item in self.try_select("amp-img.comic-contain__item")? {
+			let url = item.attr("data-src").unwrap_or_else(String::new);
+
+			if !url.is_empty() {
+				pages.push(Page {
+					content: aidoku::PageContent::Url(url, None),
```
kkantan: there should be a shorthand function for this
```suggestion
					content: aidoku::PageContent::url(url),
```

sources/zh.baozimanhua/src/html/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+						let url = item.attr("data-src").unwrap_or_else(String::new);
+
+						if !url.is_empty() {
+							pages.push(Page {
+								content: aidoku::PageContent::Url(url, None),
+								..Default::default()
+							});
+						}
+					}
+				}
```
kkantan, line 307: is this the correct `cargo fmt` result for this? I've never seen formatting like this before

sources/zh.baozimanhua/src/json/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+				let title = item.name;
+
+				// Deduplicate authors
+				let artists: Vec<String> = item
+					.author
+					.split(',')
+					.map(|s| s.trim().to_string())
+					.filter(|s| !s.is_empty())
+					.collect();
+				let mut unique = artists.clone();
```
kkantan: why does `artists` need to be cloned? can't it just be mutable?

kkantan (changes requested):

sources/zh.baozimanhua/src/html/mod.rs
dyphire: Done

sources/zh.baozimanhua/src/html/mod.rs
dyphire: Done

sources/zh.baozimanhua/src/json/mod.rs
dyphire: Done

dyphire:
> I tried opening the first manga, martial peak, and was unable access chapters 3862-3868. is there a reason for this?

It was found that it was a url splicing problem on the next page, and it was fixed.

PR #160, opened 2025-10-23
feat: add madtheme template

### LINKED ISSUE #120: toonily.me
NSFW

toonily.me source request

kkantan:
adds the following sources:
- toonily.me (closes #120)
- novelbuddy.com

the main source that uses this template, mangabuddy.com, seems to have ddos guard protection in place that is non-trivial to bypass. this may require a handler inside the app like cloudflare for the captcha. there are also a number of other sources using this template that don't use ddos guard, which should be easy to add if requested.


Hiro1404:
Toonily and novelbuddy already there, Just waiting for mangak.io(Mangabuddy.com) to be available too

PR #165, opened 2025-10-24
feat: zh.wnacg

dyphire:


sources/zh.wnacg/src/home.rs
```diff
@@ -0,0 +1,256 @@
...
+		}));
+
+		let responses: [core::result::Result<Response, RequestError>; 8] = Request::send_all([
+			// daily ranking
+			Request::get(format!(
+				"{}/albums-favorite_ranking-page-1-type-day",
+				BASE_URL
+			))?
+			.header("User-Agent", USER_AGENT)
+			.header("Origin", BASE_URL),
```
kkantan: looks like every request uses the same format and headers, so you could create an inline helper function to format the request given a path and it would reduce the code duplication a lot.

sources/zh.wnacg/src/html/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+
+impl MangaPage for Document {
+	fn manga_page_result(&self) -> Result<MangaPageResult> {
+		let mut entries: Vec<Manga> = Vec::new();
+
+		for item in self.try_select(".gallary_item")? {
+			let href = item
+				.select_first(".pic_box>a")
+				.and_then(|a| a.attr("href"))
+				.unwrap_or_default();
```
kkantan: you shouldn't allocate an empty `String` when it's never used. this is the case for `href`, `id`, and `title`. you can either continue the loop or use `filter_map` and `?`.

sources/zh.wnacg/src/html/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+				.select_first(".info>.title>a")
+				.and_then(|a| a.text())
+				.unwrap_or_default()
+				.trim()
+				.to_string();
+
+			let description = item
+				.select_first(".info>.info_col")
+				.and_then(|div| div.text())
+				.unwrap_or_default()
```
kkantan: it also doesn't make sense to turn this into an empty string either, because you still turn it back in to an `Option<String>` when passing it to the description.

sources/zh.wnacg/src/html/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+			.text()
+			.unwrap_or_default();
+
+		let full_description = if !description.is_empty() {
+			format!("{}  \n簡介：{}", page_info, description)
+		} else {
+			page_info
+		};
+
+		manga.title = title.clone();
```
kkantan: you don't need to clone the title.

sources/zh.wnacg/src/html/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+			.flat_map(|a| a.split("&"))
+			.map(|a| a.trim().to_string())
+			.collect::<Vec<String>>();
+
+		let tags = self
+			.try_select("#bodywrap>div>.uwconn>.addtags>.tagshow")?
+			.filter_map(|tag| tag.text())
+			.collect::<Vec<String>>();
+
+		let mut all_tags = categories.clone();
```
kkantan: you shouldn't need to clone `categories`.

sources/zh.wnacg/src/html/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+			entries,
+			has_next_page,
+		})
+	}
+
+	fn manga_details(&self, manga: &mut Manga) -> Result<()> {
+		let cover = self
+			.try_select_first("#bodywrap>div>.uwthumb>img")?
+			.attr("src")
+			.unwrap_or_default()
```
kkantan: why `unwrap_or_default`? if it's ever called, it will end up with `https://` being set as the manga cover, which doesn't make sense.

sources/zh.wnacg/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "wnacg"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+aes = "0.8.4"
+aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", version = "0.3.0", features = [
+    "json",
```
kkantan: do you ever use the json feature? you should remove it if it isn't necessary.

sources/zh.wnacg/src/html/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+		// Try to get upload date from the page
+		let date_uploaded = self.try_select_first(".info_col")?.text().and_then(|text| {
+			// Parse "上傳於2025-10-15" format
+			if text.contains("上傳於") {
+				let date_str = text.replace("上傳於", "").trim().to_string();
+				// Convert to RFC3339 format for parsing
+				let rfc3339 = format!("{}T00:00:00Z", date_str);
+				DateTime::parse_from_rfc3339(&rfc3339)
+					.ok()
+					.map(|dt| dt.timestamp())
```
kkantan: I don't think there's any reason to use chrono here instead of the built-in date parsing api from aidoku-rs (`parse_date`).

sources/zh.wnacg/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+		if needs_chapters {
+			let html = Url::manga(manga.key.clone()).request()?.html()?;
+			manga.chapters = Some(html.chapters(&manga.key)?);
+		}
+
+		Ok(manga)
+	}
+
+	fn get_page_list(&self, manga: Manga, _chapter: Chapter) -> Result<Vec<Page>> {
+		let url = format!("{}", Url::chapter(manga.key));
```
kkantan: can't you use `request()` here too, like the other functions?

sources/zh.wnacg/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		if needs_details {
+			let html = Url::manga(manga.key.clone()).request()?.html()?;
+			html.manga_details(&mut manga)?;
+		}
+
+		if needs_chapters {
+			let html = Url::manga(manga.key.clone()).request()?.html()?;
```
kkantan: you use the same `Url::manga(manga.key.clone())` for both branches, and `request()` doesn't consume it, so you could put it in a variable at the start and avoid cloning the key twice.

kkantan (changes requested):

sources/zh.wnacg/src/html/mod.rs
dyphire: Done

sources/zh.wnacg/src/html/mod.rs
dyphire: Done

sources/zh.wnacg/src/html/mod.rs
dyphire: Done

sources/zh.wnacg/src/lib.rs
dyphire: Done

sources/zh.wnacg/src/lib.rs
dyphire: Done

sources/zh.wnacg/src/html/mod.rs
dyphire: Done

sources/zh.wnacg/src/home.rs
dyphire: Done

sources/zh.wnacg/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+	) -> Result<Manga> {
+		let url = Url::manga(manga.key.clone());
+
+		if needs_details {
+			let html = url.request()?.html()?;
+			html.manga_details(&mut manga)?;
+		}
+
+		if needs_chapters {
+			let html = url.request()?.html()?;
```
kkantan: it looks like you're performing the same request twice if this is called with both `needs_details` and `needs_chapters` set to true.

sources/zh.wnacg/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+		let urls = text
+			.split("\\\"")
+			.filter(|s| s.starts_with("//"))
+			.map(|s| format!("https:{}", s))
+			.collect::<Vec<String>>();
+
+		let mut pages: Vec<Page> = Vec::new();
+		for url in urls.into_iter() {
+			pages.push(Page {
+				content: aidoku::PageContent::Url(url, None),
```
kkantan:
```suggestion
				content: aidoku::PageContent::url(url),
```

sources/zh.wnacg/src/lib.rs
dyphire: Fix

sources/zh.wnacg/src/lib.rs
dyphire: Done

sources/zh.wnacg/src/html/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+					title,
+					description: Some(description),
+					..Default::default()
+				})
+			})
+			.collect();
+
+		let has_next_page = self
+			.select_first(".bot_toolbar .paginator")
+			.and_then(|paginator| paginator.select_first(".next"))
```
kkantan: is there a reason this is separated?
```suggestion
			.select_first(".bot_toolbar .paginator .next")
```

sources/zh.wnacg/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+		html.manga_page_result()
+	}
+
+	fn get_manga_update(
+		&self,
+		mut manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		if needs_details || needs_chapters {
```
kkantan: this function shouldn't be called without one of these being true so I don't think this check is necessary.

kkantan (changes requested):

sources/zh.wnacg/src/lib.rs
dyphire: Done

PR #166, opened 2025-10-24
feat: add zh.picacomic

dyphire:
Test account:
username: iamabot
password: password


sources/zh.picacomic/src/crypto.rs
```diff
@@ -0,0 +1,12 @@
+use aidoku::{alloc::String, prelude::*};
+use hmac::{Hmac, Mac};
+use sha2::Sha256;
+
+type HmacSha256 = Hmac<Sha256>;
+
+pub fn encrypt(data: &[u8], key: &[u8]) -> String {
+	let mut mac = HmacSha256::new_from_slice(key).expect("");
```
kkantan: `.expect` shouldn't be used if there's any real reasonable possibility of it failing.

sources/zh.picacomic/src/net/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+		&gen_signature(&url, &gen_time(), &gen_nonce(), method),
+	);
+	request = request.header("Accept", "application/vnd.picacomic.com.v1+json");
+	if !token.is_empty() {
+		request = request.header("Authorization", &token);
+	}
+	request = request.header("Content-Type", "application/json; charset=UTF-8");
+	request = request.header("User-Agent", "okhttp/3.8.1");
+
+	if !body.is_empty() {
```
kkantan: this function shouldn't take an allocated string if you're just checking if it's not empty. you can probably use an `Option<&str>`.

sources/zh.picacomic/src/net/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+	);
+
+	Ok(token_str.to_string())
+}
+
+pub fn gen_login_url() -> String {
+	format!("{}/{}", API_URL, "auth/sign-in")
+}
+
+pub fn create_request(url: String, method: &str, body: String) -> Result<Request> {
```
kkantan: why is method a `&str` when you convert it to the `HttpMethod` enum anyways? it would be better to just use that directly.

sources/zh.picacomic/src/net/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+	let mut token = defaults_get::<String>("token").unwrap_or_default();
```
kkantan: this can stay as an `Option<String>`.

sources/zh.picacomic/src/home.rs
```diff
@@ -0,0 +1,364 @@
...
+						docs: comics_obj
+							.get("docs")
+							.and_then(|d| d.as_array())
+							.ok_or(error!("No docs in comics"))?
+							.iter()
+							.map(|item| {
+								serde_json::from_value::<ComicItem>(serde_json::Value::Object(
+									item.as_object()
+										.ok_or_else(|| error!("Invalid item format"))?
+										.clone(),
```
kkantan: does this need to be cloned?

sources/zh.picacomic/src/json/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+		} else {
+			MangaStatus::Ongoing
+		};
+		let url = format!("https://manhuabika.com/pcomicview/?cid={}", item._id);
+
+		let mut all_tags = Vec::new();
+		if let Some(tags) = &item.tags {
+			all_tags.extend(tags.iter().cloned());
+		}
+		all_tags.extend(item.categories.iter().cloned());
```
kkantan: I think since `item` is being consumed, you shouldn't need to clone these iterators.

sources/zh.picacomic/res/settings.json
```diff
@@ -0,0 +1,79 @@
...
+			},
+			{
+				"type": "text",
+				"key": "password",
+				"title": "密码",
+				"placeholder": "哔咔漫画密码",
+				"secure": true,
+				"refreshes": ["content", "listings"]
+			}
+		]
```
kkantan, line 26: can you use the `login` setting type with the basic method instead?

kkantan (changes requested):

sources/zh.picacomic/src/crypto.rs
dyphire: Done

sources/zh.picacomic/src/net/mod.rs
dyphire: Done

sources/zh.picacomic/src/home.rs
dyphire: Done

sources/zh.picacomic/src/json/mod.rs
dyphire: Done

sources/zh.picacomic/res/settings.json
dyphire: Done

sources/zh.picacomic/src/net/mod.rs
dyphire: Done

sources/zh.picacomic/src/net/mod.rs
dyphire: Done

sources/zh.picacomic/src/home.rs
```diff
@@ -0,0 +1,356 @@
...
+			net::create_request(
+				net::gen_explore_url("".into(), "dd".into(), 1),
+				HttpMethod::Get,
+				None,
+			),
+		];
+
+		let requests: [Request; 8] = requests
+			.into_iter()
+			.map(|r| r.map_err(|_| error!("Failed to create request, please check login status",)))
```
kkantan:
```suggestion
			.map(|r| r.map_err(|_| error!("Failed to create request, please check login status")))
```

sources/zh.picacomic/src/home.rs
```diff
@@ -0,0 +1,356 @@
...
+				// Explore format
+				let explore_data = ExploreData {
+					comics: ComicsData {
+						docs: comics_obj
+							.get("docs")
+							.and_then(|d| d.as_array())
+							.ok_or(error!("No docs in comics"))?
+							.iter()
+							.map(|item| {
+								serde_json::from_value::<ComicItem>(item.clone())
```
kkantan: you're still cloning `item` here; you can maybe use `into_iter` instead to avoid this.

sources/zh.picacomic/src/home.rs
```diff
@@ -0,0 +1,356 @@
...
+			} else {
+				// Rank format
+				let rank_data = RankData {
+					comics: data
+						.get("comics")
+						.and_then(|c| c.as_array())
+						.ok_or(error!("No comics in data"))?
+						.iter()
+						.map(|item| {
+							serde_json::from_value::<ComicItem>(item.clone())
```
kkantan: same here

sources/zh.picacomic/src/json/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+	if blocklist.is_empty() {
+		return false;
+	}
+
+	let mut all_tags = Vec::new();
+	if let Some(ref categories) = manga.tags {
+		all_tags.extend(categories.iter().cloned());
+	}
+
+	all_tags.iter().any(|tag| blocklist.contains(tag))
```
kkantan: you shouldn't need to clone categories into a new vector when it's not combined with anything. just use the categories iter directly to check against `blocklist`.

sources/zh.picacomic/src/json/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+			..Default::default()
+		}
+	}
+}
+
+impl From<PageItem> for aidoku::Page {
+	fn from(item: PageItem) -> Self {
+		let url = format!("{}/static/{}", item.media.file_server, item.media.path);
+		aidoku::Page {
+			content: aidoku::PageContent::Url(url, None),
```
kkantan:
```suggestion
			content: aidoku::PageContent::url(url),
```

sources/zh.picacomic/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+		.pages
+		.docs
+		.into_iter()
+		.map(Into::into)
+		.collect())
+}
+
+impl ListingProvider for Picacomic {
+	fn get_manga_list(&self, listing: Listing, page: i32) -> Result<MangaPageResult> {
+		let mut rank_time = String::new();
```
kkantan: you can use an `Option<String>` instead for these to avoid unnecessary allocation.

sources/zh.picacomic/src/net/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+							sort: sort.clone(),
+							page,
+						});
+					}
+					_ => {}
+				},
+				FilterValue::Sort {
+					id,
+					index,
+					ascending: _,
```
kkantan:
```suggestion
					..
```

sources/zh.picacomic/src/net/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+		}
+	}
+}
+
+pub fn login() -> Result<String> {
+	let username = crate::settings::get_username()?;
+	let password = crate::settings::get_password()?;
+
+	if username.is_empty() || password.is_empty() {
+		return Err(error!("Username or password not set"));
```
kkantan: you can use the shorthand macro for this.
```suggestion
		bail!("Username or password not set");
```

sources/zh.picacomic/src/settings.rs
```diff
@@ -0,0 +1,40 @@
...
+pub fn get_blocklist() -> Vec<String> {
+	defaults_get::<Vec<String>>("blockGenres").unwrap_or_default()
+}
+
+pub fn get_list_viewer() -> bool {
+	defaults_get("isListView").unwrap_or(false)
+}
+
+pub fn get_username() -> Result<String> {
+	defaults_get::<String>("username").ok_or_else(|| error!("Please login first"))
```
kkantan: technically it should be "log in" as a verb, but does it make more sense to have error messages in chinese since it's a chinese source?
```suggestion
	defaults_get::<String>("username").ok_or_else(|| error!("Please log in first"))
```

sources/zh.picacomic/src/lib.rs
dyphire: Done

sources/zh.picacomic/src/settings.rs
dyphire: Fix, I have no intention of maintaining the Chinese error message of the Chinese source. So keep it as it is.

sources/zh.picacomic/src/home.rs
dyphire: Rewrite this part of the process.

sources/zh.picacomic/src/json/mod.rs
kkantan: this wasn't changed?

sources/zh.picacomic/src/net/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+		let mut category = String::new();
+		let mut sort = String::from("dd");
+
+		for filter in filters {
+			match filter {
+				FilterValue::Text { id, value } => match id.as_str() {
+					"author" => {
+						return Ok(Self::Author {
+							author: value.to_string(),
+							sort: sort.clone(),
```
kkantan: you shouldn't need to clone these sort values since you're returning. applies to the genre filter too.
```suggestion
							sort,
```

sources/zh.picacomic/src/net/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	let json: serde_json::Value = response.get_json()?;
+	let data = json.get("data").ok_or(error!("No data in response"))?;
+	let token = data.get("token").ok_or(error!("No token in response"))?;
+	let token_str = token.as_str().ok_or(error!("Token is not a string"))?;
+
+	defaults_set(
+		"token",
+		aidoku::imports::defaults::DefaultValue::String(token_str.to_string()),
```
kkantan: you should probably just import `DefaultValue`, e.g. like you have `defaults_set`, instead of using the full namespace.

sources/zh.picacomic/src/net/mod.rs
dyphire: Done

sources/zh.picacomic/src/net/mod.rs
dyphire: Done

PR #179, opened 2025-11-03
copy over asurascans from old repo

AlokSwaminathan:


kkantan:
I think it's better to leave this repo as strictly 0.7 sources only, and it doesn't make sense to add old or unmaintained sources. the other community sources repo can still be added as a source list.

Issue #216, opened 2025-12-07
Suwayomi Extension - isit possible to add categories?


kasharn:
Hello, i am loving the Suwayomi Extension on Aidoku, and was just wondering if you was able to show the categories from Suwayomi in Aidoku when browsing the extenstion? (Similar to how Kavita works with different libraries)

At the moment it only shows 'Library'(everything together) and yes, you could argue that you should most likely do the categorising in the Aidoku app itself(when you add a manga to your library), but it would be nice to be able to also see the categories in the Suwayomi extension first, especially if you have a large library that you want to add to Aidoku in a smart way.

The scenario you would have to account for is the default on Suwayomi, if there are no categories created/made, and one where Suwayomi has categories, and so the 'Default' category doesn't have all the manga's in it(since they may only be seen in a different category).
Also it would be cool if you could choose which category is the default when viewing the extension, but i understand that might get too complex.

Thanks :)

PR #218, opened 2025-12-08
[Suwayomi]: Add ability to list out and show manga per category

gh-adrl:
Addresses: https://github.com/Aidoku-Community/sources/issues/216

1. Use `DynamicListing` to fetch all the categories in suwayomi and remove the static listing defined in source.json
2. When querying the manga list, add a filter for `categoryId` (with a special case for id 0)

Also update to latest aidoku-rs and do some minor refactors

To test out this feature, you must:

1. Create a category in suwayomi (Settings > Library > Edit Categories)
2. Move a manga into a category
4. Go to the suwayomi extension in aidoku
5. Verify manga added to the category shows up and manga in a category don't show up in Default

### DIFF
--- a/sources/multi.suwayomi/src/lib.rs
+++ b/sources/multi.suwayomi/src/lib.rs
@@ -115,9 +115,9 @@ impl Source for Suwayomi {
 			"variables": json_value,
 		});
 
-		let base_url = settings::get_base_url()?;
 		let response = self.graphql_request::<MultipleMangas>(body)?;
 
+		let base_url = settings::get_base_url()?;
 		Ok(MangaPageResult {
 			entries: response
 				.data
@@ -137,7 +137,6 @@ impl Source for Suwayomi {
 		needs_chapters: bool,
 	) -> Result<Manga> {
 		let manga_id = manga.key.parse::<i32>().expect("Invalid number");
-		let base_url = settings::get_base_url()?;
 		if needs_details {
 			let gql = graphql::GraphQLQuery::MANGA_DESCRIPTION;
 			let variables = serde_json::json!({
@@ -172,6 +171,7 @@ impl Source for Suwayomi {
 
 			let response = self.graphql_request::<MultipleChapters>(body)?;
 
+			let base_url = settings::get_base_url()?;
 			manga.chapters = Some(
 				response
 					.data
@@ -202,9 +202,9 @@ impl Source for Suwayomi {
 			"variables": variables,
 		});
 
-		let base_url = settings::get_base_url()?;
 		let response = self.graphql_request::<FetchChapterPagesResponse>(body)?;
 
+		let base_url = settings::get_base_url()?;
 		Ok(response
 			.data
 			.fetch_chapter_pages


sources/multi.suwayomi/src/lib.rs
```diff
@@ -55,6 +57,18 @@ impl Source for Suwayomi {
 						"byType": if ascending { "ASC" } else { "DESC" }
 					}));
 				}
+				FilterValue::Check { id, value } => {
```
gh-adrl, line 73: I wasn't sure which `FilterValue` to use. The `equalTo` expects an int

sources/multi.suwayomi/src/lib.rs
```diff
@@ -223,18 +237,53 @@ impl Source for Suwayomi {
...
+		});
+
+		let base_url = settings::get_base_url()?;
+		let data = Request::post(format!("{base_url}/api/graphql"))?
+			.header("Content-Type", "application/json")
+			.body(body.to_string())
+			.data()?;
+
+		let response = serde_json::from_slice::<GraphQLResponse<MultipleCategories>>(&data)
+			.map_err(|_| AidokuError::JsonParseError)?;
```
gh-adrl: I'll probably refactor the request generation into a helper function. Or better yet, if there was a graphql library but I guess I can take a look at that in the future.

sources/multi.suwayomi/src/lib.rs
```diff
@@ -55,6 +57,18 @@ impl Source for Suwayomi {
 						"byType": if ascending { "ASC" } else { "DESC" }
 					}));
 				}
+				FilterValue::Check { id, value } => {
+					if id == CATEGORY_FILTER_ID {
+						// This is special cased since the "Default" category means you don't have
+						// any categories attached to the manga.
+						let filter_value = if value == 0 {
+							serde_json::json!({"isNull": true})
```
gh-adrl, line 78: Doing an `equalTo: 0` doesn't work since there is not category associated in the manga.

This is also special cased in suwayomi-server https://github.com/Suwayomi/Suwayomi-Server/blob/b6e79532a99d8065cd6d82d7d31bd768ad4cdcd5/server/src/main/kotlin/suwayomi/tachidesk/graphql/dataLoaders/MangaDataLoader.kt#L62 when selecting a category in the Library tab

sources/multi.suwayomi/src/lib.rs
kkantan: this is a bit of an odd use of filter values, but I suppose it works fine.

sources/multi.suwayomi/src/lib.rs
kkantan: I believe aidoku-rs has been updated since this source was created, since the `AidokuError::JsonParseError` should contain the `serde_json::Error`. you should update the aidoku-rs dependency and then you can use `json()` or `json_owned()` directly on the request instead of converting to data first.

gh-adrl: Sounds good, I added the json feature to the dep, updated cargo lock, refactored the graphql request code, and used the json_owned

kkantan (approved):

PR #222, opened 2025-12-11
chore(multi.suwayomi): Minor code clean up + consistency

gh-adrl:
Minor cleanups

Contains these changes:

1. renaming graphql mutation name to be inline with the variable name
2. add a helper function to execute the graphql query that takes in the graphql query struct and variables (if applicable)
3. Replace `expect` with `map_err AidokuError::DeserializeError`
4. use i32 instead of u32 to be consistent when serializing/deserialing
5. simplify scanlator name logic


kkantan (approved):
lgtm

PR #223, opened 2025-12-11
feat(multi.suwayomi): add user and password support

### LINKED ISSUE #224: Suwayomi: add authentication support
Hey guys, I realized the Suwayomi source didn't have authentication support, and I wanted to access my library securely. I implemented a fix for this and submitted a PR in case it's useful for others too.

please let me know if anything needs adjustment 🙃🎀

Cruellest:
Adding support for simple login on suwayomi

It includes:

- Adding new configurations parameters including `User` and `Pass`
- Updating the app version
- Interacting with suwayomi login

### DIFF
--- a/sources/multi.suwayomi/res/settings.json
+++ b/sources/multi.suwayomi/res/settings.json
@@ -11,6 +11,21 @@
 				"returnKeyType": 9,
 				"autocorrectionDisabled": true,
 				"refreshes": ["content"]
+			},
+			{
+				"type": "text",
+				"placeholder": "User",
+				"key": "user",
+				"autocorrectionDisabled": true,
+				"refreshes": ["content"]
+			},
+			{
+				"type": "text",
+				"placeholder": "Password",
+				"key": "pass",
+				"secureTextEntry": true,
+				"autocorrectionDisabled": true,
+				"refreshes": ["content"]
 			}
 		]
 	}
--- a/sources/multi.suwayomi/res/settings.json
+++ b/sources/multi.suwayomi/res/settings.json
@@ -13,20 +13,28 @@
 				"refreshes": ["content"]
 			},
 			{
-				"type": "text",
-				"placeholder": "User",
-				"key": "user",
-				"autocorrectionDisabled": true,
+				"type": "select",
+				"key": "authMode",
+				"title": "Auth Mode",
+				"values": ["auto", "none", "basic_auth", "simple_login"],
+				"titles": [
+					"Auto",
+					"None",
+					"Basic Auth",
+					"Simple Login"
+				],
+				"default": "auto",
 				"refreshes": ["content"]
 			},
 			{
-				"type": "text",
-				"placeholder": "Password",
-				"key": "pass",
-				"secureTextEntry": true,
-				"autocorrectionDisabled": true,
+				"type": "login",
+				"method": "basic",
+				"key": "credentials",
+				"title": "Login",
+				"requires": "baseUrl",
 				"refreshes": ["content"]
 			}
-		]
+		],
+		"footer": "Enter your Suwayomi server URL and optionally login with your credentials."
 	}
 ]
--- a/sources/multi.suwayomi/src/lib.rs
+++ b/sources/multi.suwayomi/src/lib.rs
@@ -30,6 +30,16 @@ impl Suwayomi {
 		T: serde::de::DeserializeOwned,
 	{
 		let base_url = settings::get_base_url()?;
+
+		if let Some((user, pass)) = settings::get_credentials() {
+			let form = format!("user={}&pass={}", user, pass);
+			let _ = Request::post(format!("{}/login.html", base_url))?
+				.header("Content-Type", "application/x-www-form-urlencoded")
+				.body(form)
+				.send()
+				.ok();
+		}
+
 		Request::post(format!("{base_url}/api/graphql"))?
 			.header("Content-Type", "application/json")
 			.body(body.to_string())
--- a/sources/multi.suwayomi/src/lib.rs
+++ b/sources/multi.suwayomi/src/lib.rs
@@ -21,6 +21,52 @@ use aidoku::{
 };
 use alloc::string::ToString;
 use alloc::vec;
+use core::fmt::Write;
+
+fn url_encode(input: &str) -> String {
+    let mut output = String::new();
+    for byte in input.bytes() {
+        match byte {
+            b'A'..=b'Z' | b'a'..=b'z' | b'0'..=b'9' | b'-' | b'_' | b'.' | b'~' => {
+                output.push(byte as char);
+            }
+            _ => {
+                let _ = write!(output, "%{:02X}", byte);
+            }
+        }
+    }
+    output
+}
+
+fn base64_encode(input: &str) -> String {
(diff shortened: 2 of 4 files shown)


sources/multi.suwayomi/src/lib.rs
```diff
@@ -30,6 +30,16 @@ impl Suwayomi {
 		T: serde::de::DeserializeOwned,
 	{
 		let base_url = settings::get_base_url()?;
+
+		if let Some((user, pass)) = settings::get_credentials() {
```
gh-adrl: I'm not too familiar with aidoku auth patterns but looking at the aidoku-rs source code, this is probably what you want: https://github.com/Aidoku/aidoku-rs/blob/405b91d74c95e8099217b30ede333f852e6d9c5c/crates/lib/src/structs/source.rs#L116 instead of POSTing to login at every request.

If this is not the pattern, looking at the paperback implementation, an authorization header should be set https://github.com/Suwayomi/tachidesk-paperback-ext/blob/813586cdecd6094461b48bca95cc84220320c621/src/TachiDesk/TachiDesk.ts#L96

I'll leave it to @kkantan for guidance

kkantan: yes, I would agree with this. @Cruellest there is a specific "login" setting type that can also be combined with a `BasicLoginHandler` if necessary. I briefly checked the [suwayomi wiki](https://github.com/Suwayomi/Suwayomi-Server/wiki/Configuring-Suwayomi%E2%80%90Server), and it looks like you should just be able to send an `Authorization` header with each request that contains the base64-encoded user name and password, thus not requiring a handler.

sources/multi.suwayomi/src/settings.rs
kkantan, line 1: it looks like spaces are being used here for indentation instead of tabs. make sure to run `cargo fmt` and it should fix this.

kkantan (changes requested):

sources/multi.suwayomi/src/settings.rs
Cruellest: yeah i didin`t saw that, and first time using rust thx for the help

sources/multi.suwayomi/src/lib.rs
Cruellest: The login component its really better, thanks for the recomendation, i implemented diferent login logic for diferent authentication types, when using basic authentication it really is as simple as sending an authorization header, but the other login type is a little more tricky, please tell me if there is any problem with it, or if i didn't understood any recomendation

Syer10:
> Wow thx for the sugestion, i will look at the Mihon implementation and see what i could improve, On that note i am implementing the UI Login logic but i couldn't find any instance of official UI that implements it (at least in my instance of suwayomi it doesn't exists) is there any recomendation of UI that uses it or any method that i could use to debug it?

The Mihon extension does implement it. And the WebUI does as well. I would suggest you mainly reference the Mihon extension though since it is a much more centralized implementation. WebUI handles different things in different places. Such as the login(get refresh token) request in the [Login page](https://github.com/Suwayomi/Suwayomi-WebUI/blob/84cfa6bea52a4d22834d6d336e393159856c24c5/src/features/authentication/screens/LoginPage.tsx#L41), and the refresh access token in the [base client interface](https://github.com/Suwayomi/Suwayomi-WebUI/blob/84cfa6bea52a4d22834d6d336e393159856c24c5/src/lib/requests/client/BaseClient.ts#L44).

A lot of the testing for the UI_LOGIN can be done in Graphql Playground(found at /api/graphql).
Using these mutations
```gql
mutation Login($password: String!, $username: String!) {
  login(input: {password: $password, username: $username}) {
    refreshToken
    accessToken
  }
}

mutation RefreshAccessToken($refreshToken: String!) {
  refreshToken(input: {refreshToken: $refreshToken}) {
    accessToken
  }
}
```

Then you can test normal queries by changing the headers field at the bottom of the GQL Playground to something like
```json
{
  "authorization": "Bearer MY_ACCESS_TOKEN"
}
```

sources/multi.suwayomi/src/lib.rs
```diff
@@ -15,12 +15,61 @@ use aidoku::imports::std::send_partial_result;
...
+		if chunk.len() > 2 {
+			output.push(CHARSET[(triple & 0x3F) as usize] as char);
+		} else {
+			output.push('=');
+		}
+	}
+	output
+}
+
+fn url_encode(input: &str) -> String {
```
kkantan, line 55: there's an `encode_uri_component` in aidoku-rs helpers you can use instead, unless this is a specific implementation?

sources/multi.suwayomi/src/lib.rs
```diff
@@ -15,12 +15,61 @@ use aidoku::imports::std::send_partial_result;
...
-	alloc::{String, Vec},
+	alloc::{String, Vec, vec},
 	imports::net::Request,
 	prelude::*,
 };
 use alloc::string::ToString;
-use alloc::vec;
+use core::fmt::Write;
+
+fn base64_encode(input: &str) -> String {
```
kkantan, line 25: you can use the `base64`, with default features disabled, instead of implementing it yourself. I believe there are other sources that do this.

sources/multi.suwayomi/src/settings.rs
```diff
@@ -1,11 +1,23 @@
-use aidoku::{AidokuError, alloc::string::String, imports::defaults::defaults_get, prelude::bail};
+use aidoku::{Result, alloc::String, imports::defaults::defaults_get, prelude::*};
 
-const BASE_URL_KEY: &str = "baseUrl";
```
kkantan, line 3: no reason to remove this. you can also add a `AUTH_MODE_KEY` const.

sources/multi.suwayomi/src/lib.rs
```diff
@@ -30,10 +79,54 @@ impl Suwayomi {
...
-			.header("Content-Type", "application/json")
-			.body(body.to_string())
-			.json_owned::<GraphQLResponse<T>>()
+		let auth_mode = settings::get_auth_mode();
+		let body_str = body.to_string();
+
+		let send_req = |with_basic: bool| {
+			let mut request = Request::post(format!("{base_url}/api/graphql"))?
+				.header("Content-Type", "application/json")
+				.body(body_str.clone());
```
kkantan, line 88: this shouldn't need to be cloned

sources/multi.suwayomi/res/settings.json
```diff
@@ -11,7 +11,30 @@
...
+				"values": ["auto", "none", "basic_auth", "simple_login"],
+				"titles": [
+					"Auto",
+					"None",
+					"Basic Auth",
+					"Simple Login"
+				],
+				"default": "auto",
+				"refreshes": ["content"]
+			},
```
kkantan, line 28: is it not possible to determine this automatically via some request to suwayomi? I would expect it to be, but if not that's okay.

kkantan (changes requested):

sources/multi.suwayomi/res/settings.json
Syer10: It is not. Most clients ask the user to input the auth mode.

PR #226, opened 2025-12-14
fix(fr.bigsolo): adapt to new apis

apix0n:
## content of this pr

i updated the source to use the new api implementation of BigSolo, and i added some home components to match the new homepage.


sources/fr.bigsolo/src/lib.rs
```diff
@@ -31,66 +28,73 @@ impl Source for BigSolo {
...
+				continue;
+			}
 
-			// extract filename from file_path to use as manga entry key
-			let file_name = file_path
-				.split('/')
-				.next_back()
-				.unwrap_or("unknown")
-				.replace(".json", "");
+			let slug = series.slug.clone();
```
kkantan: you shouldn't need to clone this, since it's only used in a `format!`

sources/fr.bigsolo/src/lib.rs
```diff
@@ -31,66 +28,73 @@ impl Source for BigSolo {
...
+			let manga = Manga {
+				key: series.slug.clone(),
+				title: series.title.clone(),
+				cover: Some(series.cover.url_lq.clone()),
+				authors: Some(vec![series.author]),
+				artists: Some(vec![series.artist]),
+				description: Some(series.description),
+				status: MangaStatus::Unknown,
```
kkantan: no need to specify this since it's included in the `..Default::default()`

sources/fr.bigsolo/src/lib.rs
```diff
@@ -31,66 +28,73 @@ impl Source for BigSolo {
 		_page: i32,
 		_filters: Vec<FilterValue>,
 	) -> Result<MangaPageResult> {
-		let manga_list = BigSolo::get_bigsolo_manga_files();
+		let response = Request::get("https://bigsolo.org/data/series")?.string()?;
```
kkantan: usually it's good to define a `const BASE_URL: &str` so that it's easy to update in the future if the domain changes

sources/fr.bigsolo/src/lib.rs
```diff
@@ -31,66 +28,73 @@ impl Source for BigSolo {
...
+		let series_list: SeriesList =
+			serde_json::from_str(&response).map_err(AidokuError::message)?;
```
kkantan: you should update the aidoku-rs dependency, add the "json" feature, and then use `.json()` or `.json_owned()` instead

sources/fr.bigsolo/src/lib.rs
```diff
@@ -246,31 +213,68 @@ impl Home for BigSolo {
...
+		let response = Request::get("https://bigsolo.org/data/series")?.string()?;
+		let home_data: SeriesList =
+			serde_json::from_str(&response).map_err(AidokuError::message)?;
+
+		let reco_entries = home_data
+			.reco
+			.into_iter()
+			.map(|series| Manga {
+				key: series.slug.clone(),
+				title: series.title.clone(),
```
kkantan: you shouldn't need to clone this since the `series` object is consumed. it also might be worthwhile to just implement `From<...> for Manga` so you can simply do `.map(Into::into)`. you can check other sources that do this for the json response models.

kkantan (changes requested):

apix0n:
fixed that, thanks for the comments

sources/fr.bigsolo/src/lib.rs
```diff
@@ -242,35 +183,76 @@ impl BigSolo {
 	}
 }
 
+impl From<Series> for Manga {
```
kkantan: this should be put in `models.rs` along with the `Series` struct definition.

sources/fr.bigsolo/src/lib.rs
```diff
@@ -31,70 +30,53 @@ impl Source for BigSolo {
...
-				authors: Some(vec![series_data.author]),
-				artists: Some(vec![series_data.artist]),
-				description: Some(series_data.description),
-				status: BigSolo::map_bigsolo_status(&series_data.release_status),
-				tags: Some(tags),
-				url: Some(format!("https://bigsolo.org/{slug}")),
-				..Default::default()
-			};
+			let mut manga: Manga = series.into();
+			manga.status = MangaStatus::Unknown;
```
kkantan: shouldn't this be marked as completed for one-shots? what does the site data have?

sources/fr.bigsolo/src/lib.rs
```diff
@@ -151,29 +130,19 @@ impl Source for BigSolo {
...
-							.groups
-							.values()
-							.next()
-							.and_then(|url| url.split('/').next_back())
-							.unwrap_or("unknown");
-
+						let chapter_key_str = chapter_key.as_str();
 						Chapter {
-							key: String::from(chapter_id),
+							key: chapter_key.clone(),
```
kkantan: you can also avoid cloning this by defining the variables that use its value (`chapter_number` and `url`) above the struct initialization.

sources/fr.bigsolo/src/lib.rs
```diff
@@ -151,29 +130,19 @@ impl Source for BigSolo {
...
-							},
-							locked: chapter.licencied.as_ref().is_some_and(|v| *v),
+							chapter_number: chapter_key_str.parse().ok(),
+							volume_number: chapter.volume.unwrap_or_default().parse().ok(),
+							date_uploaded: Some(chapter.timestamp),
+							url: Some(format!(
+								"{BASE_URL}/{}/{}",
+								series_data.slug, chapter_key_str
+							)),
+							scanlators: Some(chapter.teams.clone()),
```
kkantan: does this need to be cloned?

sources/fr.bigsolo/src/lib.rs
```diff
@@ -242,35 +183,76 @@ impl BigSolo {
...
+		latest_chapters.truncate(10);
+
+		let latest_chapters: Vec<MangaWithChapter> = latest_chapters
+			.into_iter()
+			.filter_map(|(series_slug, chapter_key, chapter_data)| {
+				// Find the manga for this chapter
+				let manga = all_entries.iter().find(|m| m.key == series_slug)?.clone();
+
+				let chapter = Chapter {
+					key: chapter_key.clone(),
```
kkantan: like above, you shouldn't need to clone anything here.

sources/fr.bigsolo/src/lib.rs
```diff
@@ -377,12 +336,89 @@ impl AlternateCoverProvider for BigSolo {
...
+		if listing.id == "catalogue" {
+			// we need to sort by title because the catalogue is not sorted by title
+			let mut mangas = self.get_search_manga_list(None, page, Vec::new())?.entries;
+			mangas.sort_by(|a, b| a.title.cmp(&b.title));
+			Ok(MangaPageResult {
+				entries: mangas,
+				has_next_page: false,
+			})
+		} else {
+			Err(AidokuError::message("Unknown listing"))
```
kkantan: there's a shorthand for this. you could also use `Err(error!("..."))`.
```suggestion
			bail!("Unknown listing")
```

sources/fr.bigsolo/src/lib.rs
```diff
@@ -377,12 +336,89 @@ impl AlternateCoverProvider for BigSolo {
...
+			Request::get(format!("{BASE_URL}/data/series"))?.json_owned::<SeriesList>()?;
+
+		// if old_key matches, return the slug as the new identifier
+		for series in series_list.series.iter().chain(series_list.os.iter()) {
+			if series.key == old_key {
+				return Ok(series.slug.clone());
+			}
+		}
+
+		Err(AidokuError::message(format!(
```
kkantan: you can also use `bail!` here without needing the `format!`.

sources/fr.bigsolo/src/lib.rs
```diff
@@ -377,12 +336,89 @@ impl AlternateCoverProvider for BigSolo {
...
+impl MigrationHandler for BigSolo {
+	fn handle_manga_migration(&self, old_key: String) -> Result<String> {
+		// fetch /data/series to find the series
+		let series_list: SeriesList =
+			Request::get(format!("{BASE_URL}/data/series"))?.json_owned::<SeriesList>()?;
+
+		// if old_key matches, return the slug as the new identifier
+		for series in series_list.series.iter().chain(series_list.os.iter()) {
+			if series.key == old_key {
+				return Ok(series.slug.clone());
```
kkantan:
```suggestion
				return Ok(series.slug);
```

sources/fr.bigsolo/src/lib.rs
```diff
@@ -377,12 +336,89 @@ impl AlternateCoverProvider for BigSolo {
...
+		Err(AidokuError::message(format!(
+			"Could not find manga with old key '{}' for migration",
+			old_key
+		)))
+	}
+
+	fn handle_chapter_migration(&self, _manga_key: String, chapter_key: String) -> Result<String> {
+		// the old chapter ids are not available anymore,
+		// we cannot do a match like in handle_manga_migration.
+		Ok(chapter_key)
```
kkantan: does this mess anything up? if you don't implement the migrationhandler, then the app will automatically prompt the user to migrate data using the app's built-in migration view, which may be more complete, but slower.

apix0n: it actually _does_ work better through aidoku's built-in migration view, i just didn't know that we could do sources migrations using it

apix0n:
fixed again, thanks

PR #229, opened 2025-12-17
fix(libgroup): switch to web login and update API endpoints

### LINKED ISSUE #118: HentaiLib.me Failed to Load Chapters
The chapters won't open. I think it's because of the banner on the website that warns that the manga is 18+

image
image

remytomme:
Replaces the broken OAuth flow with Web Login to successfully pass DDOS-Guard checks.
- Implements LocalStorage token extraction from Aidoku's get_local_storage (via `auth` key).
- Updates API URLs.
- Bumps versions for all LibGroup sources.
- Fixed wrong url used on chapters fetch

Closes #118

### DIFF
--- a/sources/ru.mangalib/res/source.json
+++ b/sources/ru.mangalib/res/source.json
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "ru.mangalib",
 		"name": "MangaLib",
-		"version": 7,
+		"version": 8,
 		"url": "https://mangalib.me",
 		"contentRating": 1,
 		"minAppVersion": "0.7.1",
--- a/templates/libgroup/src/auth/mod.rs
+++ b/templates/libgroup/src/auth/mod.rs
@@ -11,11 +11,12 @@ use aidoku::{
 use crate::{
 	context::Context,
 	endpoints::Url,
-	models::responses::{TokenResponse, UserResponse},
+	models::responses::{LocalStorageWrapper, TokenResponse, UserResponse},
 	settings::get_api_url,
 };
 
 const TOKEN_KEY: &str = "login";
+const AUTH_KEY: &str = "login.ls.auth";
 const USER_ID_KEY: &str = "user_id";
 
 pub const USER_AGENT: &str = "Mozilla/5.0 (iPhone; CPU iPhone OS 18_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.0 Mobile/15E148 Safari/604.1";
@@ -43,9 +44,10 @@ impl AuthRequest for Request {
 			.header("User-Agent", USER_AGENT);
 
 		if let Ok(token) = get_token()
-			&& let Some(access_token) = token.access_token
+			&& let Some(access_token) = &token.access_token
 		{
-			self = self.header(HEADER_AUTH, &format!("{AUTH_SCHEME} {access_token}"));
+			let scheme = token.token_type.as_deref().unwrap_or(AUTH_SCHEME);
+			self = self.header(HEADER_AUTH, &format!("{scheme} {access_token}"));
 		}
 
 		let response = self.send()?;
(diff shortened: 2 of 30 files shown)


sources/ru.mangalib/res/source.json
```diff
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "ru.mangalib",
 		"name": "MangaLib",
-		"version": 7,
+		"version": 8,
 		"url": "https://mangalib.me",
 		"contentRating": 1,
 		"minAppVersion": "0.7.1",
```
kkantan, line 8: this needs to be updated for all the sources.
```suggestion
		"minAppVersion": "0.8",
```

templates/libgroup/src/auth/mod.rs
```diff
@@ -72,8 +75,13 @@ impl AuthRequest for Request {
...
 fn get_token() -> Result<TokenResponse> {
-	defaults_get_json::<TokenResponse>(TOKEN_KEY)
-		.map_err(|_| AidokuError::Message("No token".into()))
+	if let Ok(token) = defaults_get_json::<TokenResponse>(TOKEN_KEY) {
+		return Ok(token);
+	}
+	if let Ok(wrapper) = defaults_get_json::<LocalStorageWrapper>(AUTH_KEY) {
+		return Ok(wrapper.token);
+	}
+	Err(AidokuError::Message("No token".into()))
```
kkantan:
```suggestion
	bail!("No token")
```

templates/libgroup/src/chapters/test.rs
```diff
@@ -13,12 +14,16 @@ fn test_context() -> Context {
 	}
 }
 
-fn fake_now() -> i64 {
-	1_000_000
+// Thread-safe atomic mock time
+static MOCK_TIME: AtomicI64 = AtomicI64::new(1_000_000);
```
kkantan: the compiled wasm is run single threaded, so you don't need to use atomic types.

templates/libgroup/src/auth/mod.rs
```diff
@@ -43,9 +44,10 @@ impl AuthRequest for Request {
 			.header("User-Agent", USER_AGENT);
 
 		if let Ok(token) = get_token()
-			&& let Some(access_token) = token.access_token
+			&& let Some(access_token) = token.access_token.clone()
```
kkantan: does this need to be cloned? you should be able to use it in your `format!` as a reference instead.

sources/ru.mangalib/res/source.json
remytomme: The authentication logic is backward compatible. It attempts to retrieve the token from `defaults`, which the new v0.8 WebView integration populates from `localStorage`. On the current app version, this key is simply absent, so the source skips the login flow without errors.

Since public sources like MangaLib function correctly without a user session, keeping them at `0.7.1` allows users to benefit from the fixed API endpoints immediately. HentaiLib is the only source bumped to `0.8` because it is auth-walled and entirely non-functional without the token extraction.

Given that the URL fixes restore functionality for the public sources on the current app version, should I leave the versions as they are to merge now, or would you prefer I bump `minAppVersion` for all of them and wait for the 0.8 release?

templates/libgroup/src/chapters/mod.rs
```diff
@@ -50,33 +50,47 @@ impl ChaptersCache {
...
+				return Ok(entry.data.clone());
+			}
+		}
+
+		// 2. Slow path: write lock
+		// Only one thread enters here
 		let mut guard = self.cache.write();
 
-		// Check cache hit and TTL
+		// Double-check: another thread might have inserted it while we waited for the write lock
```
kkantan: this comment is somewhat misleading since technically this is impossible since there's no multithreading happening. it's not really strictly necessary to change since the logic still works, but I believe it is a little overcomplicated.

sources/ru.mangalib/res/source.json
kkantan: oh I see, sorry for missing that in my review. in that case, it should be fine to leave them since they're not linking any newer functions that would prevent running them on older versions.

kkantan (approved):

templates/libgroup/src/chapters/mod.rs
remytomme: Done. It's ready for review and merge now.
Thanks for reviewing, and happy new year or forthcoming one 🎄

PR #235, opened 2025-12-18
feat: add proxy http for cuutruyen

tachibana-shin:
All the domains of this source have been blocked and they will not buy new domains so we need proxies

### DIFF
--- a/sources/vi.cuutruyen/src/lib.rs
+++ b/sources/vi.cuutruyen/src/lib.rs
@@ -353,7 +353,7 @@ impl BaseUrlProvider for CuuTruyen {
 		if proxy_url.is_empty() {
 			Ok(url)
 		} else {
-			Ok(format!("{}/?url={}", proxy_url, url))
+			Ok(format!("{proxy_url}/?url={url}"))
 		}
 	}
 }


sources/vi.cuutruyen/res/settings.json
kkantan, line 1: this file should use tabs for indentation

kkantan (changes requested):
to mark the source as updated you'll need to increment the version in the `source.json` file.

however, does this actually fix the source? it doesn't seem to load for me with this change. it's also worth noting that the source still works for me on the main branch, and the domains appear to still load for me in a web browser as well.

sources/vi.cuutruyen/res/settings.json
```diff
@@ -1,9 +1,21 @@
...
-  {
-    "type": "text",
-    "key": "proxy_url",
-    "placeholder": "HTTP Proxy URL",
-    "title": "Proxy URL",
-    "default": "https://hard-emu-60.deno.dev"
-  }
+	{
+		"type": "group",
+		"title": "Cài đặt",
```
kkantan: you can use the app's localization.
```suggestion
		"title": "SETTINGS",
```

sources/vi.cuutruyen/src/lib.rs
```diff
@@ -347,7 +347,14 @@ impl DeepLinkHandler for CuuTruyen {
 
 impl BaseUrlProvider for CuuTruyen {
 	fn get_base_url(&self) -> Result<String> {
-		Ok(defaults_get::<String>("url").unwrap_or_default())
+		let url = defaults_get::<String>("url").unwrap_or_default();
+		let proxy_url = defaults_get::<String>("proxy_url").unwrap_or("".to_string());
```
kkantan:
```suggestion
		let proxy_url = defaults_get::<String>("proxy_url").unwrap_or_default();
```

sources/vi.cuutruyen/src/lib.rs
```diff
@@ -347,7 +347,14 @@ impl DeepLinkHandler for CuuTruyen {
...
+
+		if proxy_url.is_empty() {
+			Ok(url)
+		} else {
+			Ok(format!("{}/?url={}", proxy_url, url))
```
kkantan:
```suggestion
			Ok(format!("{proxy_url}/?url={url}"))
```

kkantan (changes requested):
only issue with this is that I believe the app will show the proxy url as one of the selectable base urls, which could mess things up, but I'm not sure if that's fixable or if it's something the app should handle.

PR #241, opened 2025-12-19
Add `vi.foxtruyen`, `vi.nettruyen`, `vi.truyen3q`, `vi.truyenqq`, `vi.truyenqq` and template `wpcomics`

tachibana-shin:


sources/vi.nettruyen/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "vi.nettruyen",
+		"name": "Net truyện",
```
kkantan: this is how they label the titles on the website.
```suggestion
	"info": {
		"id": "vi.nettruyen",
		"name": "NetTruyen",
```

sources/vi.nettruyen/res/source.json
```diff
@@ -0,0 +1,12 @@
...
+		"version": 2,
```
kkantan: if you're replacing the existing source on the legacy repo, it should be one version incremented.
```suggestion
		"version": 16,
```

sources/vi.nettruyen/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "netruyen"
```
kkantan:
```suggestion
name = "nettruyen"
```

sources/vi.foxtruyen/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "vi.foxtruyen",
+		"name": "Fox truyện",
+		"version": 2,
```
kkantan, line 5: this one is new.
```suggestion
		"version": 1,
```

sources/vi.foxtruyen/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "vi.foxtruyen",
+		"name": "Fox truyện",
```
kkantan:
```suggestion
		"name": "FoxTruyen",
```

sources/vi.truyen3q/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "truyenqq"
```
kkantan:
```suggestion
name = "truyen3q"
```

sources/vi.nettruyen/res/icon.png
kkantan, line 1: this isn't that great of an icon since the app will apply a different corner radius to the icon.

sources/vi.truyen3q/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "vi.truyen3q",
+		"name": "Truyện 3Q",
+		"version": 2,
```
kkantan:
```suggestion
		"name": "TruyenTranh3Q",
		"version": 1,
```

sources/vi.truyenqq/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "vi.truyenqq",
+		"name": "Truyện QQ",
+		"version": 2,
```
kkantan:
```suggestion
		"name": "TruyenQQ",
		"version": 7,
```

sources/vi.truyenqq/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+use wpcomics::{Impl, Params, WpComics, helper::urlencode};
+
+const USER_AGENT: &str = "Mozilla/5.0 (iPhone; CPU iPhone OS 17_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) GSA/300.0.598994205 Mobile/15E148 Safari/604";
+const BASE_URL: &str = "https://truyenqqno.com";
+
+fn get_visit_read_id() -> Result<String> {
+	Ok(defaults_get::<String>("visitReadId")
+		.map(|v| v.trim_end_matches('/').to_string())
+		.unwrap_or_default())
+}
```
kkantan: there's no point in returning a `Result` if you're just unwrapping it anyways. just return a `String`.

sources/vi.truyenqq/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+impl Impl for TruyenQQ {
+	fn new() -> Self {
+		Self
+	}
+
+	fn params(&self) -> Params {
+		let cookie = Some(format!(
+			"visit-read={}",
+			get_visit_read_id().unwrap_or_default()
+		));
```
kkantan: it might be worth checking if the result is empty, and if it is then using `None` instead.

sources/vi.truyenqq/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+					.nth(1)
+					.and_then(|s| s.split('/').next())
+					.unwrap_or_default()
+					.into()
+			},
+			chapter_parse_id: |url| {
+				String::from(
+					url.rsplit_once("-chap-")
+						.map(|(_, tail)| tail.trim_end_matches(".html"))
+						.unwrap(),
```
kkantan: try to avoid `unwrap` when possible.

sources/vi.truyenqq/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+					"{}/truyen-tranh/{}-chap-{}",
+					params.base_url, manga.key, chapter.key
+				)
+			},
+
+			get_search_url: |params, q, page, filters| {
+				let mut excluded_tags: Vec<String> = Vec::new();
+				let mut included_tags: Vec<String> = Vec::new();
+				let mut query = QueryParameters::new();
+				query.push("q", Some(&q.unwrap_or_default()));
```
kkantan: I think you can do this.
```suggestion
				query.push("q", q.as_deref());
```

sources/vi.truyenqq/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+						_ => {}
+					}
+				}
+
+				Ok(format!(
+					"{}/tim-kiem-nang-cao/trang-{}.html?category={}&notcategory={}&{}",
+					params.base_url,
+					page,
+					included_tags.join(","),
+					excluded_tags.join(","),
```
kkantan: why are you handling tags separately? you should be able to just add them as a parameter to `query` inside the filter match.

templates/wpcomics/src/helpers.rs
```diff
@@ -0,0 +1,358 @@
...
+		.collect::<String>()
+		.split(' ')
+		.collect::<Vec<&str>>()
+		.into_iter()
+		.map(|a: &str| a.parse::<f32>().unwrap_or(-1.0))
+		.filter(|a| *a >= 0.0)
+		.collect::<Vec<f32>>()
+}
+
+pub fn urlencode(string: String) -> String {
```
kkantan: can you use `encode_uri_component` from the aidoku-rs helpers instead?

sources/vi.truyenqq2/res/source.json
kkantan, line 1: why is this separate from `vi.truyenqq`? could you just add the domain as a mirror instead that's selectable as a base url? I'm not sure if having two different sources with the same icon and everything is very good practice.

templates/wpcomics/scripts/get-options-id.ts
kkantan, line 1: ideally all files should have an extra blank newline at the end, and it might be worth adding some explanation for which page to run this script on. I would just make a `README.md` file like some of the other templates with these two code blocks.

templates/wpcomics/src/helper.rs
```diff
@@ -0,0 +1,358 @@
...
+	if temp.find("cimoC") == Some(0) {
+		temp
+			.replacen("cimoC", "", 1)
+			.chars()
+			.rev()
+			.collect::<String>()
+	} else {
+		temp.chars().rev().collect::<String>()
+	}
+}
```
kkantan: this seems like a bit of an odd implementation to me, and rather inefficient since you're reversing a string twice.

something like this could be done instead:
```rs
pub fn trunc_trailing_comic(title: &str) -> &str {
    title.strip_suffix("Comic").unwrap_or(title)
}
```
note that taking and returning string slices is better than allocated strings when possible.

templates/wpcomics/src/helpers.rs
```diff
@@ -0,0 +1,358 @@
...
+		50 => "vertigo",
+		51 => "video-games",
+		52 => "war",
+		53 => "western",
+		54 => "zombies",
+		_ => "",
+	})
+}
+
+pub fn text_with_newlines(node: Element) -> String {
```
kkantan, line 87: would this work instead? https://github.com/Aidoku/aidoku-rs/blob/96d7d8c784bfb1327dabe1bef5cc0ffcaa334f62/crates/lib/src/helpers/element.rs#L10

templates/wpcomics/src/helpers.rs
kkantan, line 1: I think it would be more consistent if this module was named `helpers` instead, to match other sources and templates.

templates/wpcomics/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+};
+use core::cell::RefCell;
+
+pub mod helper;
+mod imp;
+
+pub use imp::Impl;
+
+pub struct Params {
+	pub base_url: String,
```
kkantan: I think it's better to use `Cow<'static, str>` instead of `String` or `&'static str` for all of the plain string parameters, because then either can be used while still avoiding allocation where possible.

templates/wpcomics/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+			user_agent: None,
+
+			datetime_format: "MMMM dd, yyyy",
+			datetime_locale: "en_US_POSIX",
+			datetime_timezone: "current",
+
+			genre_endpoint: "tim-kiem-nang-cao.html",
+
+			cache_manga_id: None,
+			cache_manga_value: None,
```
kkantan: I don't think putting these in the params is a good idea. Ideally, the params should be static and just what are passed in for configuration from the sources using the template. Instead, you should probably create a new `CacheStore` (name up to you) struct that can be held in a `RefCell` on the `WpComics` struct and passed into the `Impl` functions.

templates/wpcomics/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+
+	fn get_page_list(&self, manga: Manga, chapter: Chapter) -> Result<Vec<Page>> {
+		let mut params = self.params.borrow_mut();
+		self.inner.get_page_list(&mut params, manga, chapter)
+	}
+}
+
+impl<T: Impl> ListingProvider for WpComics<T> {
+	fn get_manga_list(&self, _listing: aidoku::Listing, _page: i32) -> Result<MangaPageResult> {
+		todo!()
```
kkantan: panics should be avoided, since I believe they still cause the source to fail until the app is restarted.
```suggestion
		AidokuError::Unimplemented
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+	fn get_cache_manga_id(&self, params: &mut Params) -> Option<String> {
+		params.cache_manga_id.clone()
+	}
+	fn set_cache_manga_id(&self, params: &mut Params, value: Option<String>) {
+		params.cache_manga_id = value;
+	}
+
+	fn get_cache_manga_value(&self, params: &mut Params) -> Option<Vec<u8>> {
+		params.cache_manga_value.clone()
+	}
```
kkantan: could you instead get a reference to a slice of bytes, instead of cloning the entire vec?

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+};
+
+pub trait Impl {
+	fn new() -> Self;
+
+	fn params(&self) -> Params;
+
+	fn get_cache_manga_id(&self, params: &mut Params) -> Option<String> {
+		params.cache_manga_id.clone()
+	}
```
kkantan: you should be able to use `as_deref` to get a slice instead.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+	fn set_cache_manga_value(&self, params: &mut Params, value: Option<Vec<u8>>) {
+		params.cache_manga_value = value;
+	}
+
+	fn cache_manga_page(&self, params: &mut Params, url: &str) -> Result<()> {
+		let cached_id = self.get_cache_manga_id(params).clone();
+
+		if cached_id == Some(url.to_string()) {
+			return Ok(());
+		}
```
kkantan: you're cloning the cached id twice and then allocating a new string for the url, but none of that should be necessary. just compare string slices.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+	}
+
+	fn create_request(
+		&self,
+		params: &mut Params,
+		url: &str,
+		headers: Option<&[(&str, &str)]>,
+	) -> Result<Request> {
+		// 通常のリクエスト
+		let mut req = Request::new(url, HttpMethod::Get)?;
```
kkantan: the request shorthand functions are preferred.
```suggestion
		let mut req = Request::get(url)?;
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+		#[allow(clippy::needless_match)]
+		let mut viewer = params.viewer;
+		for category in categories {
+			match category.to_ascii_lowercase().as_str() {
+				"smut" | "mature" | "18+" | "adult" => nsfw = ContentRating::NSFW,
+				"ecchi" | "16+" => {
+					nsfw = match nsfw {
+						ContentRating::NSFW => ContentRating::NSFW,
+						_ => ContentRating::Suggestive,
+					}
```
kkantan:
```suggestion
					if nsfw != ContentRating::NSFW {
						nsfw = ContentRating::Suggestive
					}
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+
+		let html = self.create_request(params, &search_url, headers)?.html()?;
+
+		let Some(elems) = html.select(params.manga_cell) else {
+			return Ok(MangaPageResult {
+				entries: vec![],
+				has_next_page: false,
+			});
+		};
+		let mut entries: Vec<Manga> = Vec::with_capacity(elems.size());
```
kkantan: you can `iter` the `elems` and collect into a vector instead of allocating one with the correct size and pushing to it. most other sources do this.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+		for item_node in elems {
+			let title = item_node
+				.select(params.manga_cell_title)
+				.and_then(|node| node.first())
+				.and_then(|n| n.text());
+			let url = item_node
+				.select(params.manga_cell_url)
+				.and_then(|node| node.first())
+				.and_then(|n| n.attr("abs:href"))
+				.unwrap_or("".to_string());
```
kkantan:
```suggestion
				.unwrap_or_default();
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+					.and_then(|v| v.first())
+					.and_then(|n| n.attr(params.manga_cell_image_attr))
+			} else {
+				None
+			};
+			entries.push(Manga {
+				key: (params.manga_parse_id)(url).to_string(),
+				cover,
+				title: (params.manga_details_title_transformer)(title.unwrap_or("".to_string()))
+					.to_string(),
```
kkantan: doesn't this return a `String` already?

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+				..Default::default()
+			});
+		}
+		if !params.next_page.is_empty() {
+			has_next_page = html
```
kkantan: this doesn't need to be a mutable variable, you can just assign to it here with an else branch.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+			.and_then(|n| n.first())
+			.and_then(|n| n.attr(params.manga_details_cover_attr));
+
+		let authors = Some((params.manga_details_authors_transformer)(
+			details
+				.select(params.manga_details_authors)
+				.map(|l| {
+					l.map(|node| String::from(node.text().unwrap_or("".to_string()).trim()))
+						.filter(|s| !s.is_empty())
+						.collect()
```
kkantan: you can use `filter_map` with `node.text()` to simplify this and avoid some allocations.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+		Ok(MangaPageResult {
+			entries,
+			has_next_page,
+		})
+	}
+
+	fn parse_manga_element(&self, params: &mut Params, url: String) -> Result<Manga> {
+		self.cache_manga_page(params, url.as_str())?;
+
+		let html = self.get_cache_manga_value(params).unwrap();
```
kkantan: it might be worth making the `cache_manga_page` return the value as well as caching it, since it appears that's the only time `get_cache_manga_value` is used anyways and it would avoid an `unwrap`.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+		let details = Html::parse_with_url(html, url.clone())?;
```
kkantan: the url shouldn't need to be cloned
```suggestion
		let details = Html::parse_with_url(html, &url)?;
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+				})
+				.unwrap_or_default(),
+		));
+		let description = details
+			.select(params.manga_details_description)
+			.and_then(|l| l.first())
+			.map(text_with_newlines);
+		let mut tags = Vec::new();
+
+		if !params.manga_details_tags.is_empty() {
```
kkantan: usually when there are checks like this it would make more sense to use an `Option` instead of just a string.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+			(params.manga_details_status_transformer)(
+				details
+					.select(params.manga_details_status)
+					.and_then(|v| v.text())
+					.unwrap_or("".to_string()),
+			)
+			.to_string(),
+		);
+		Ok(Manga {
+			key: (params.manga_parse_id)(url.clone()).to_string(),
```
kkantan: you should make this function take a string slice so you don't need to clone the url to send it through. it also already returns an allocated string so you shouldn't need `to_string`.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+			status,
+			content_rating,
+			viewer,
+			..Default::default()
+		})
+	}
+
+	fn get_chapter_list(&self, params: &mut Params, url: String) -> Result<Vec<Chapter>> {
+		let mut skipped_first = false;
+		let mut chapters: Vec<Chapter> = Vec::new();
```
kkantan: you should just collect an iterator here as well instead of pushing to a vec.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+		);
+		let title = title_untrimmed.trim();
+		for chapter_node in html.select(params.manga_details_chapters).unwrap() {
+			if params.chapter_skip_first && !skipped_first {
+				skipped_first = true;
+				continue;
+			}
+			let chapter_url = chapter_node
+				.select_first(params.chapter_anchor_selector)
+				.unwrap()
```
kkantan: don't unwrap. you can use a `filter_map` to have it automatically filter out with `?`.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+					.unwrap()
+					.text()
+					.unwrap_or_default(),
+			);
+
+			chapter_title = chapter_title.trim().to_string();
+			chapter_title = String::from(if chapter_title.is_empty() {
+				title_raw.trim()
+			} else {
+				&chapter_title
```
kkantan: double allocation

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+			});
+
+			chapters.push(Chapter {
+				key: chapter_id.to_string(),
+				title: if chapter_title.is_empty() {
+					None
+				} else {
+					Some(chapter_title)
+				},
```
kkantan:
```suggestion
				title: (!chapter_title.is_empty()).then_some(chapter_title),
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+			if page_url.is_none() {
+				page_url = if page_node.has_attr("src") {
+					page_node.attr("src")
+				} else {
+					page_url
+				}
+			}
+
+			pages.push(Page {
+				content: PageContent::Url(
```
kkantan: use `PageContent::url`

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+	}
+
+	fn get_search_manga_list(
+		&self,
+		params: &mut Params,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		let url = { (params.get_search_url)(params, query, page, filters)? };
```
kkantan: why is this in curly braces?

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+		let html = Html::parse_with_url(html, url.clone())?;
+		if html.select(params.manga_viewer_page).is_none() {
+			let breadcrumbs = html.select(".breadcrumb li").unwrap();
+			let manga_id = breadcrumbs
+				.get(breadcrumbs.size() / 2 - 2)
+				.expect("node array")
+				.select_first("a")
+				.unwrap()
+				.attr("abs:href")
+				.unwrap_or_default();
```
kkantan: don't unwrap or expect.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,620 @@
...
+					page_url
+				}
+			}
+			if page_url.is_none() {
+				page_url = if page_node.has_attr("src") {
+					page_node.attr("src")
+				} else {
+					page_url
+				}
+			}
```
kkantan: this logic looks like it can be simplified. one option would just be a single conditional expression with else if branches.

sources/vi.nettruyen/res/source.json
tachibana-shin: I don't like this; it messes up the versions of the sources I'm maintaining. Doesn't Aidoku have a way to differentiate between them?

sources/vi.foxtruyen/res/source.json
tachibana-shin: It needs to be 2 to avoid conflicts with my power supply.

sources/vi.nettruyen/res/icon.png
tachibana-shin: I wouldn't want to redraw a website icon.

sources/vi.truyen3q/res/source.json
tachibana-shin: It was renamed "Truyen3Q" because that's the group's official name.

templates/wpcomics/src/imp.rs
tachibana-shin: This code doesn't seem easy to read.

templates/wpcomics/src/imp.rs
tachibana-shin: Why does the cookie think it's `::url` when I see `:Url` in the source code?

templates/wpcomics/src/imp.rs
tachibana-shin: Scope borrow check

templates/wpcomics/src/imp.rs
tachibana-shin: I like this one; it's clearer and easier to read than chain.

sources/vi.truyenqq/src/lib.rs
tachibana-shin: No, it doesn't follow the RC39 standard. It requires a stringified query like `?foo=1,2,3,4`. If I append it, it will be `?foo=1&foo==2...`

sources/vi.truyenqq2/res/source.json
tachibana-shin: No, they are different sources; you shouldn't consider them as one. They are just built from a common template.

templates/wpcomics/src/lib.rs
tachibana-shin: I can't use RefCell because I'm using the latest version of Rust, which doesn't like RefCell and will refuse to compile.

templates/wpcomics/src/helpers.rs
tachibana-shin: I want to avoid `set_text` as much as possible

templates/wpcomics/src/imp.rs
tachibana-shin: you mean Option for tags?

templates/wpcomics/src/imp.rs
tachibana-shin: Iterator doesn't make much sense here, I don't think it's necessary to complicate things

README.md
kkantan, line 1: this shouldn't be changed.

sources/vi.nettruyen/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+	alloc::{string::ToString, *},
+	helpers::uri::QueryParameters,
+	prelude::*,
+};
+use wpcomics::{Impl, Params, WpComics};
+
+const USER_AGENT: &str = "Mozilla/5.0 (iPhone; CPU iPhone OS 17_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) GSA/300.0.598994205 Mobile/15E148 Safari/604";
+const BASE_URL: &str = "https://nettruyenviet1.com";
+
+struct FoxTruyen;
```
kkantan:
```suggestion
struct NetTruyen;
```

sources/vi.nettruyen/res/source.json
kkantan: aidoku will display whatever the highest version source is from all combined source lists, so the version needs to be at least one above the previous highest version, which was 15 on the legacy community repo.

sources/vi.truyen3q/res/source.json
kkantan: sources are meant to be a reflection of the websites they scrape (at least, in my eyes), and the site says "TruyenTranh3Q" everywhere: the page titles, the header, the domain, etc. I can't find any mention of "Truyen3Q".

sources/vi.truyenqq/src/lib.rs
kkantan:
```rs
query.push("foo", Some(&included.join(","))); // "foo=1,2,3"
```

sources/vi.truyenqq2/res/source.json
kkantan: I would suggest renaming to something like "TruyenQQ.online", in order to have some way of differentiating the two sources. since keiyoushi doesn't have this source I'm not sure. ideally it should be apparent from looking at the source list which source corresponds to which website.

templates/wpcomics/src/lib.rs
kkantan: `RefCell` is already being used on line 240 though?

templates/wpcomics/src/helpers.rs
kkantan: you should just use the function directly instead of calling a wrapper `urlencode` function.

templates/wpcomics/src/helpers.rs
```diff
@@ -0,0 +1,330 @@
...
+	let html = node.html().unwrap_or("".to_string());
+	if !String::from(html.trim()).is_empty() {
```
kkantan: this shouldn't need to be allocated

templates/wpcomics/src/helpers.rs
```diff
@@ -0,0 +1,330 @@
...
+			)
+			.as_bytes(),
+		) {
+			node.select_first("div")
+				.and_then(|v| v.text())
+				.map(|t| t.replace("{{ .LINEBREAK }}", "\n"))
+				.unwrap_or("".to_string())
+		} else {
+			String::new()
+		}
```
kkantan: use `and_then` on the parsed result, then `unwrap_or_default` at the end to avoid an else branch.

templates/wpcomics/src/helpers.rs
```diff
@@ -0,0 +1,330 @@
...
+		51 => "video-games",
+		52 => "war",
+		53 => "western",
+		54 => "zombies",
+		_ => "",
+	})
+}
+
+pub fn text_with_newlines(node: Element) -> String {
+	let html = node.html().unwrap_or("".to_string());
```
kkantan: `unwrap_or_default`

templates/wpcomics/src/helpers.rs
```diff
@@ -0,0 +1,330 @@
...
+				.unwrap_or("".to_string())
+		} else {
+			String::new()
+		}
+	} else {
+		String::new()
+	}
+}
+
+pub fn category_parser(
```
kkantan: this function looks unused; you have it duplicated in `imp.rs`.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+		}
+		self.modify_request(params, req)
+	}
+
+	fn category_parser(
+		&self,
+		params: &mut Params,
+		categories: &Vec<String>,
+	) -> (ContentRating, Viewer) {
+		#[allow(clippy::needless_match)]
```
kkantan: I don't think these are necessary.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+		&self,
+		params: &mut Params,
+		search_url: String,
+		headers: Option<&[(&str, &str)]>,
+	) -> Result<MangaPageResult> {
+		let html = self.create_request(params, &search_url, headers)?.html()?;
+
+		let Some(elems) = html.select(params.manga_cell) else {
+			return Ok(MangaPageResult {
+				entries: vec![],
```
kkantan:
```suggestion
				entries: Vec::new(),
```

templates/wpcomics/src/imp.rs
kkantan: for any of the string params that you conditionally use only if they're not empty.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+
+		if !params.manga_details_tags.is_empty() {
+			if params.manga_details_tags_splitter.is_empty() {
+				tags = details
+					.select(params.manga_details_tags)
+					.map(|list| {
+						list.map(|elem| elem.text().unwrap_or_default())
+							.collect::<Vec<_>>()
+					})
+					.unwrap_or(vec![]);
```
kkantan:
```suggestion
					.unwrap_or_default();
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+						.unwrap_or_default()
+						.split(params.manga_details_tags_splitter)
+						.map(|s| s.trim().to_string())
+						.filter(|s| !s.is_empty())
+						.collect::<Vec<String>>()
+				});
+
+				if let Some(split) = split {
+					for node in split {
+						tags.push(node);
```
kkantan: you can do the same this as the if branch and just assign to `tags` instead of pushing stuff individually, and then you don't need it to be mutable.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+			content_rating,
+			viewer,
+			..Default::default()
+		})
+	}
+
+	fn get_chapter_list(&self, params: &mut Params, url: String) -> Result<Vec<Chapter>> {
+		let html = self.cache_manga_page(params, url.as_str())?;
+
+		let html = Html::parse_with_url(html, url)?;
```
kkantan:
```suggestion
		let html_data = self.cache_manga_page(params, &url)?;
		let html = Html::parse_with_url(html_data, url)?;
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+		let title_untrimmed = (params.manga_details_title_transformer)(
+			html.select(params.manga_details_title)
+				.and_then(|v| v.text())
+				.unwrap_or_default(),
+		);
+		let title = title_untrimmed.trim();
+		let mut skipped_first = false;
+
+		let Some(chapters_iter) = html.select(params.manga_details_chapters) else {
+			return Ok(vec![]);
```
kkantan:
```suggestion
			return Ok(Vec::new());
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+						.text()
+						.unwrap_or_default(),
+				);
+
+				chapter_title = chapter_title.trim().to_string();
+				chapter_title = String::from(if chapter_title.is_empty() {
+					title_raw.trim()
+				} else {
+					&chapter_title
+				});
```
kkantan: still double allocating unnecessarily

templates/wpcomics/src/imp.rs
kkantan: `PageContent::Url` is the enum case, `PageContent::url` is the shorthand function you can use so you don't need to provide the `None` for the page context.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+				..Default::default()
+			});
+		}
+
+		Ok(pages)
+	}
+
+	fn handle_deep_link(&self, params: &mut Params, url: String) -> Result<Option<DeepLinkResult>> {
+		let html = self.cache_manga_page(params, url.as_str())?;
+		let html = Html::parse_with_url(html, url.clone())?;
```
kkantan:
```suggestion
		let html_data = self.cache_manga_page(params, &url)?;
		let html = Html::parse_with_url(html_data, &url)?;
```

templates/wpcomics/src/imp.rs
kkantan: ?

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+		}
+		let (content_rating, viewer) = self.category_parser(params, &tags);
+		let status = (params.status_mapping)((params.manga_details_status_transformer)(
+			details
+				.select(params.manga_details_status)
+				.and_then(|v| v.text())
+				.unwrap_or_default(),
+		));
+		Ok(Manga {
+			key: (params.manga_parse_id)(url.clone()),
```
kkantan: these functions should still take `&str` instead of `String` so you don't need to clone so much extra stuff.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+
+			manga.cover = new_manga.cover;
+			manga.title = new_manga.title;
+			manga.authors = new_manga.authors;
+			manga.description = new_manga.description;
+			manga.url = new_manga.url;
+			manga.tags = new_manga.tags;
+			manga.status = new_manga.status;
+			manga.content_rating = new_manga.content_rating;
+			manga.viewer = new_manga.viewer;
```
kkantan: I think you can use `manga.copy_from(new_manga);` instead.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,619 @@
...
+		if needs_chapters {
+			manga.chapters = Some(self.get_chapter_list(params, url)?);
+		}
+
+		Ok(manga)
+	}
+
+	fn get_home(&self, params: &mut Params) -> Result<HomeLayout> {
+		let base_url = &params.base_url.clone();
+		let html = { self.create_request(params, base_url, None)?.html()? };
```
kkantan: you shouldn't need the extra clone and scoping here.

templates/wpcomics/src/lib.rs
kkantan: this applies to all the other `&'static str` parameters too.

sources/vi.foxtruyen/src/lib.rs
```diff
@@ -0,0 +1,160 @@
+#![no_std]
+use aidoku::{
+	alloc::{string::ToString, *},
+	helpers::uri::QueryParameters,
+	prelude::*,
+	FilterValue, Source, Viewer,
+};
+use wpcomics::{helpers::urlencode, Impl, Params, WpComics};
+
+const USER_AGENT: &str = "Mozilla/5.0 (iPhone; CPU iPhone OS 17_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) GSA/300.0.598994205 Mobile/15E148 Safari/604";
```
kkantan, line 10: does not providing this user agent break the cloudflare handling? usually it's best to use the default user agent, since then it matches the device. for example, sometimes user agents like this don't work on ipad and macOS, where cloudflare can detect that the device isn't an iPhone without the user agent.

sources/vi.foxtruyen/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+			manga_cell_image_attr: "data-src",
+
+			manga_details_title: ".title_tale h1",
+			manga_details_cover: ".thumbblock img",
+			manga_details_chapters: ".item_chap",
+			chapter_anchor_selector: "a",
+			chapter_date_selector: "em",
+
+			manga_parse_id: |url| {
+				String::from(
```
kkantan: I think you can just do `.into()` at the end instead of using `String::from`.

sources/vi.foxtruyen/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+			manga_viewer_page: ".content_detail_manga img",
+
+			manga_details_authors: ".org",
+			manga_details_description: "div.story-detail-info.detail-content",
+			manga_details_tags: "ul.list01 > li",
+			manga_details_tags_splitter: "",
+			manga_details_status: "li.status.row p.col-xs-9",
+
+			chapter_skip_first: false,
+			page_url_transformer: |url| url,
```
kkantan: isn't this the default?

templates/wpcomics/src/helpers.rs
```diff
@@ -0,0 +1,330 @@
...
+	alloc::{String, Vec, str, string::ToString},
+	helpers::uri::QueryParameters,
+	imports::html::{Element, Html},
+	prelude::format,
+};
+use chrono::{NaiveDate, NaiveDateTime};
+
+use crate::Params;
+
+pub fn extract_f32_from_string(title: String, text: String) -> Vec<f32> {
```
kkantan: this should take `&str`s instead of `String`s.

sources/vi.foxtruyen/src/lib.rs
tachibana-shin: I don't think that's necessary. Aidoku seems to handle Cloudflare requests very openly.

tachibana-shin: > does not providing this user agent break the cloudflare handling? usually it's best to use the default user agent, since then it matches the device. for example, sometimes user agents like this don't work on ipad and macOS, where cloudflare can detect that the device isn't an iPhone without the user agent.

Anyway, don't worry, I've tested all the cases.

sources/vi.truyen3q/res/source.json
tachibana-shin: I know their last name and their name is `truyen3q` anyway if you like it I don't like the new name because it's too long

templates/wpcomics/src/lib.rs
tachibana-shin: No, I wouldn't change the arguments and returns to static, it creates redundant %.into()% in sources that use the template, I won't use it, and it doesn't make much sense there either.

templates/wpcomics/src/lib.rs
tachibana-shin: i wouldn't change it i wouldn't have the patience to add 2 parameters to all functions that use cache

templates/wpcomics/src/imp.rs
tachibana-shin: clone is need because borrow checker

templates/wpcomics/src/imp.rs
tachibana-shin: good fn

sources/vi.truyenqq2/res/source.json
tachibana-shin: I think 'TuyenQQ2' is enough

templates/wpcomics/src/lib.rs
tachibana-shin: thanks chat gpt 4.1 edit code

sources/vi.truyenqq/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+					},
+				);
+				query.push(
+					"notcategory",
+					if excluded_tags.is_empty() {
+						None
+					} else {
+						Some(&exclude)
+					},
+				);
```
kkantan: there's no reason to push to a separate vec and then have these outside of the match on `FilterValue::MultiSelect`. also, should the query parameter still exist if there are no tags?

this also applies to the other sources.

sources/vi.foxtruyen/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+			home_sliders_selector: ".homepage_suggest",
+			home_sliders_title_selector: "h2",
+			home_sliders_item_selector: "li",
+
+			home_grids_selector: "section > div > .col-md-6, .container > section:nth-child(1)",
+			home_grids_title_selector: ".title_cate",
+			home_grids_item_selector: ".item_home",
+
+			home_manga_cover_attr: "abs:data-src",
+			time_formats: Some(["%d/%m/%Y", "%m-%d-%Y", "%Y-%d-%m"].to_vec()),
```
kkantan:
```suggestion
			time_formats: Some(vec!["%d/%m/%Y", "%m-%d-%Y", "%Y-%d-%m"]),
```

sources/vi.nettruyen/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+			manga_parse_id: |url| {
+				String::from(
+					url.split("truyen-tranh/")
+						.nth(1)
+						.and_then(|s| s.split('/').next())
+						.unwrap_or_default(),
+				)
+			},
+			chapter_parse_id: |url| {
+				String::from(url.trim_end_matches('/').rsplit('/').next().unwrap())
```
kkantan: don't use `unwrap()`.

sources/vi.nettruyen/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+						_ => {}
+					}
+				}
+
+				Ok(format!(
+					"{}/tim-truyen/{}{}{query}",
+					params.base_url,
+					tag,
+					if query.is_empty() { "" } else { "?" }
+				))
```
kkantan:
```suggestion
				let mut tag = None;

				for filter in filters {
					match filter {
						FilterValue::Select { id, value } => {
							if id == "tag" {
								tag = Some(value);
							} else {
								query.push(&id, Some(&value));
							}
						}
						FilterValue::Sort { id, index, .. } => {
							query.push(&id, Some(&index.to_string()));
						}
						_ => {}
					}
				}

				Ok(format!(
					"{}/tim-truyen/{}{}{query}",
					params.base_url,
					tag.as_deref().unwrap_or_default(),
					if query.is_empty() { "" } else { "?" }
				))
```

sources/vi.truyenqq2/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+	helpers::uri::QueryParameters,
+	imports::defaults::defaults_get,
+	prelude::*,
+};
+use wpcomics::{Impl, Params, WpComics};
+
+const USER_AGENT: &str = "Mozilla/5.0 (iPhone; CPU iPhone OS 17_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) GSA/300.0.598994205 Mobile/15E148 Safari/604";
+const BASE_URL: &str = "https://truyenqq.online";
+
+fn get_visit_read_id() -> Result<String> {
```
kkantan: this should match the one in `vi.truyenqq`.

sources/vi.truyenqq2/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+				url.split("truyen-tranh/")
+					.nth(1)
+					.and_then(|s| s.split('/').next())
+					.unwrap_or_default()
+					.into()
+			},
+			chapter_parse_id: |url| {
+				url.rsplit_once("chapter/")
+					.map(|(_, tail)| tail.trim_end_matches(".html"))
+					.unwrap()
```
kkantan: don't use `unwrap()`.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,628 @@
...
+	prelude::*,
+};
+
+pub trait Impl {
+	fn new() -> Self;
+
+	fn params(&self) -> Params;
+	fn cache_manga_page(
+		&self,
+		cache: &mut Option<crate::Cache>,
```
kkantan: just add this to the namespace with `use crate::Cache`.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,628 @@
...
+		params: &Params,
+		url: &str,
+	) -> Result<Vec<u8>> {
+		if let Some(cached) = cache.as_ref()
+			&& cached.manga_id == url
+		{
+			// need .clone() avoid lifetime issues. not request change
+			return Ok(cached.manga_value.clone());
```
kkantan: what? your previous version of this function was able to return a `u8` slice instead of a vec just fine. the data is stored on the `Cache` object, which looks you have a reference to in all the places this function is called, so there should be no lifetime issues.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,628 @@
...
+		}
+
+		let req = self.create_request(cache, params, url, None)?;
+		let data = req.data()?;
+		*cache = Some(crate::Cache {
+			manga_id: url.to_string(),
+			// need .clone() avoid lifetime issues. not request change
+			manga_value: data.clone(),
```
kkantan: also not necessary. the value can be consumed here, then referenced as a slice for the return value.

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,628 @@
...
+		})
+	}
+
+	fn parse_manga_element(
+		&self,
+		cache: &mut Option<crate::Cache>,
+		params: &Params,
+		url: String,
+	) -> Result<Manga> {
+		let html = self.cache_manga_page(cache, params, url.as_str())?;
```
kkantan:
```suggestion
		let html = self.cache_manga_page(cache, params, &url)?;
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,628 @@
...
+	fn get_search_manga_list(
+		&self,
+		cache: &mut Option<crate::Cache>,
+		params: &Params,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		let url = (params.get_search_url)(params, query, page, filters)?;
+		println!("url = {url}");
```
kkantan:
```suggestion
```

templates/wpcomics/src/imp.rs
```diff
@@ -0,0 +1,628 @@
...
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		let url = (params.manga_page)(params, &manga);
+
+		if needs_details {
+			let new_manga = self.parse_manga_element(cache, params, url.clone())?;
+
+			manga.copy_from(new_manga);
+
+			send_partial_result(&manga);
```
kkantan:
```suggestion
			if needs_chapters {
				send_partial_result(&manga);
			}
```

sources/vi.nettruyen/src/lib.rs
tachibana-shin: It's pointless to remove the initial value of the string as empty and then unwrap it back to empty.

templates/wpcomics/src/imp.rs
tachibana-shin: Try changing it; it has many complex constraints regarding lifespan, we don't need to waste time dealing with this.

sources/vi.truyenqq/src/lib.rs
tachibana-shin: So what's your idea? Concatenating is slower than joining vectors.

sources/vi.truyenqq2/src/lib.rs
tachibana-shin: No problem at all.

templates/wpcomics/src/imp.rs
kkantan: I would argue that this is the most important part *to* deal with: the html pages you're fetching from the requests are likely some of the largest objects the source will handle, and cloning them will double the memory usage.

tachibana-shin: Why do you think `.clone() vec` increases memory usage? That's not true.

templates/wpcomics/src/imp.rs
```diff
@@ -109,38 +109,40 @@ pub trait Impl {
...
+				Some(Manga {
+					key: (params.manga_parse_id)(&url),
+					cover,
+					title: (params.manga_details_title_transformer)(title.unwrap_or_default()),
+					..Default::default()
+				})
 			})
-		});
-		let entries: Vec<Manga> = entries.collect();
+			.collect();
```
tachibana-shin, line 144: We should separate it to reduce the value by 1 when fmt formatting.

tachibana-shin:
> * the filters in `filters.json` should use the "title" field instead of "name", so they actually have titles in the filters ui.
> * the selectors for `vi.foxtruyen` look wrong, since tags and descriptions don't appear. e.g. you should have `a.clblue` instead of `ul.list01 > li` for tags.

We should add the JSON schema to the manual JSON formatting check; otherwise, the JSON format should be omitted.

tachibana-shin:
@kkantan What is the role of the `refreshes` field in filters.json? I didn't find its role in Aidoku Runner

kkantan:
> What is the role of the refreshes field in filters.json? I didn't find its role in Aidoku Runner

when changing the setting in the app, aidoku will refresh the corresponding thing. for example, adding a "content" refresh will reload the home and/or search page, depending on whichever is open, when the setting is changed. likewise, for filters and listings I believe it just calls the dynamic provider functions again to update them in the ui.

> Can this repo host such a hybrid source or 1 pure novel source?

if the source is usable in aidoku it's good to have here, but if it uses some apis or anything that aidoku doesn't support then I don't think it would make sense to have, at least until the changes land in a public build and can be locked behind `minAppVersion`. you might want to make a rakuyomi-only repo instead.

I believe there's also another source on here that supports novels, but it's just simple text without any formatting.

templates/wpcomics/src/imp.rs
kkantan: I'm not sure what you mean by this

kkantan (approved):

PR #247, opened 2025-12-22
Add setting for deduplicated chapter and poster quality select settings

dangobums:


sources/en.comix/res/settings.json
kkantan, line 1: json files should use tabs for indentation for consistency.

sources/en.comix/src/lib.rs
```diff
@@ -24,6 +24,45 @@ const NSFW_GENRE_IDS: &[&str] = &["87264", "8", "87265", "13", "87266", "87268"]
...
+		None => {
+			map.insert(key, ch);
+		}
+		Some(current) => {
+			if is_better(&ch, current) {
+				map.insert(key, ch);
+			}
+		}
+	}
+}
```
kkantan: I think it would be better to move these functions into a `helpers.rs` file.

sources/en.comix/src/lib.rs
```diff
@@ -151,31 +190,53 @@ impl Source for Comix {
...
-						.into_chapters(&manga.key),
+			let limit = 100;
+			let mut page = 1;
+			let deduplicate = settings::get_dedupchapter();
+			let mut chapter_map: HashMap<String, ComixChapter> = HashMap::new();
+			let mut chapter_list: Vec<ComixChapter> = Vec::new();
+			loop {
+				let url = format!(
+					"{API_URL}/manga/{}/chapters?limit={}&page={}&order[number]=desc",
+					manga.key, limit, page
```
kkantan:
```suggestion
					"{API_URL}/manga/{}/chapters?limit={limit}&page={page}&order[number]=desc",
					manga.key
```

sources/en.comix/src/lib.rs
```diff
@@ -151,31 +190,53 @@ impl Source for Comix {
...
+			let raw_chapters = if deduplicate {
+				chapter_map.into_values().collect::<Vec<_>>()
+			} else {
+				chapter_list
+			};
+
+			let mut chapters: Vec<Chapter> = raw_chapters
+				.into_iter()
+				.map(|item| item.into_chapter(&manga.key))
+				.collect();
```
kkantan: I think we can avoid a collection by keeping it as an iterator.
```suggestion
			let raw_chapters = if deduplicate {
				chapter_map.into_values()
			} else {
				chapter_list.into_iter()
			};

			let mut chapters: Vec<Chapter> = raw_chapters
				.map(|item| item.into_chapter(&manga.key))
				.collect();
```

sources/en.comix/src/lib.rs
```diff
@@ -151,31 +190,53 @@ impl Source for Comix {
...
+
+			chapters.sort_by(|a, b| {
+				b.chapter_number
+					.partial_cmp(&a.chapter_number)
+					.unwrap_or(core::cmp::Ordering::Equal)
+			});
```
kkantan: this is only necessary when using the `chapter_map`, correct? you should put it in a `if deduplicate` condition.

kkantan (approved):

kkantan (approved):

PR #251, opened 2025-12-23
feat(nhentai): Dynamic filters, tag translation & favorites

yozurukt:
Switched from static JSON to `DynamicFilters` to allow better customization and localization.

**Key Changes:**
* **Tag Translation:** Supports translating tags and metadata. It handles bidirectional lookup, so you can use localized tags in Blocklist and Favorites directly.
* **Performance:** used `phf` (Perfect Hash Function) for translations to keep overhead minimal.
* **Refactor:** `update_tags.py` now generates a strict typed `tags.rs` instead of JSON.
* **UX:** Added "Favorite Tags" and sorting options (Popularity vs Alphabetical).

**Note:**
Enabling translation might slightly affect performance, but I've optimized it with PHF.
Some settings may require **re-opening the source** to take effect.

### DIFF
--- a/sources/multi.nhentai/res/source.json
+++ b/sources/multi.nhentai/res/source.json
@@ -1,8 +1,8 @@
 {
 	"info": {
-		"id": "multi.nhentai-test",
+		"id": "multi.nhentai",
 		"name": "nhentai",
-		"version": 2,
+		"version": 16,
 		"url": "https://nhentai.net",
 		"contentRating": 2,
 		"languages": ["All", "en", "ja", "zh"]


yozurukt:
I completely understand your concerns. I'm still new to Rust and this type of development environment, so I underestimated the performance and binary size implications. I agree that a 10x increase in binary size and the additional interpreter overhead is too high a cost for the main repository, especially since the majority of users won't need the localization.
I'll follow your advice and maintain this as a separate fork for users who specifically need Chinese support.
I'll close this PR now. Thanks for the feedback!

PR #253, opened 2025-12-28
Add latest listing for comix

dangobums:


sources/en.comix/src/lib.rs
```diff
@@ -366,9 +366,22 @@ impl ListingProvider for Comix {
...
+			self.get_search_manga_list(
+				None,
+				page,
+				vec![FilterValue::Sort {
+					id: "order".into(),
+					index: 1, // recent update
+					ascending: false,
+				}],
+			)
+		};
```
kkantan, line 379: this will fetch the results for every listing, regardless of whether it's the latest listing or not, resulting in extra network requests.

sources/en.comix/res/source.json
kkantan, line 1: the source version needs to be incremented for every update.

PR #255, opened 2025-12-29
vi.cmanga

tachibana-shin:


sources/vi.cmanga/res/filters.json
```diff
@@ -0,0 +1,720 @@
...
+			"Adult",
+			"Ma Pháp",
+			"Bệnh Kiều",
+			"Tình Yêu Công Sở",
+			"Girl Love",
+			"Ác Nữ",
+			"Ảo Tưởng",
+			"Idol"
+		],
+		"ids": [
```
kkantan: are the ids all the same as the options? in that case, they could be left out.

sources/vi.cmanga/res/icon.png
kkantan, line 1: this icon is too low quality; it looks like there's one [here](https://cmangax9.com/assets/img/brand/cmanga/logo-mobile.png) that could be used.

sources/vi.cmanga/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+				.collect::<Vec<_>>(),
+				listing: None,
+			},
+		}));
+		send_partial_result(&HomePartialResult::Component(HomeComponent {
+			title: Some(String::from("Hoàn Thành")),
+			subtitle: None,
+			value: aidoku::HomeComponentValue::Scroller {
+				entries: Request::get(format!(
+					"{}/api/home_album_list?file=image&sort=update&limit=14&page=1",
```
kkantan, line 201: it would probably be better to parallelize all these requests with `Request::send_all` instead.

sources/vi.cmanga/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+			content: PageContent::url(p),
+			..Default::default()
+		})
+		.collect();
+
+		Ok(pages)
+	}
+}
+
+register_source!(CManga, Home);
```
kkantan, line 358: it would be nice to implement `DeepLinkHandler` as well, since most other sources do.

sources/vi.cmanga/src/models.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+impl From<MangaItem> for MangaWithChapter {
+	fn from(value: MangaItem) -> Self {
+		let id = value.info.chapter.id.clone();
+		let last = value.info.chapter.last;
+
+		Self {
+			manga: value.into(),
+			chapter: Chapter {
+				key: id.to_string(),
```
kkantan: isn't `id` already a `String`?

sources/vi.cmanga/src/models.rs
```diff
@@ -0,0 +1,280 @@
...
+			key: value.info.id,
+			title: if value.info.name.is_empty() {
+				None
+			} else {
+				Some(value.info.name)
+			},
+			chapter_number: value.info.num.parse::<f32>().ok(),
+			date_uploaded: parse_datetime_to_timestamp(&value.info.last_update),
+			url: None,
+			language: Some("vi".to_owned()),
```
kkantan: I don't believe providing this is necessary for single-language sources.

kkantan (changes requested):
accidental approval

sources/vi.cmanga/res/icon.png
tachibana-shin: We shouldn't use an aspect ratio that's too different.

sources/vi.cmanga/src/lib.rs
tachibana-shin: The loading time is very long, so loading in parts is the correct approach since the website also does that.

sources/vi.cmanga/src/lib.rs
tachibana-shin: This source cannot implement deep linking because its referencing URLs are almost impossible to reverse engineer from the original URL (url detail != URL chapter).

sources/vi.cmanga/res/icon.png
kkantan: it's a transparent png that can be placed on a 128x128 image with a white background.

tachibana-shin: And it would be super small, I don't think I'd do that.

tachibana-shin: use picsart to upscale quality

PR #257, opened 2025-12-31
Vi.zettruyen

tachibana-shin:


sources/vi.zettruyen/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "vi.zettruyen",
+		"name": "ZetTruyen",
+		"version": 1,
+		"url": "https://wwww.zettruyen.com",
```
kkantan:
```suggestion
		"url": "https://www.zettruyen.com",
```

sources/vi.zettruyen/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+
+			page += 1;
+		}
+
+		Ok(chapters)
+	}
+}
+
+#[derive(Deserialize)]
+struct Related {
```
kkantan: it would be nicer to separate these serde structs into a `models.rs` file.

tachibana-shin: There are only a few structs I don't want to separate.

PR #264, opened 2026-01-02
New Source: es.manhwaweb

Llavesuke:
New Source: Es.manhwaweb

# Description
This PR adds a new source for **ManhwaWeb** (Spanish web).

## Features
- **Listing**: Latest Chapters, New Works.
- **Search**: Supports text search and filtering by genre, status, and sort order.
- **Manga Details**: Fetches cover, synopsis, status, authors, and chapter list.
- **Chapter Images**: correctly resolves and displays chapter pages.

## Verification
I have manually built the package and tested it on an **iPhone**:
- [x] Source installs and loads correctly.
- [x] Home page sections populate.
- [x] Search and filters work.
- [x] Chapter images load successfully.

This is my first contribution to Aidoku! I'm happy to help the community. I'll be waiting for any feedback or requested changes. Thanks for all the great work all has been doing


sources/es.manhwaweb/res/filters.json
kkantan, line 1: all the json files should use tabs for indentation and end with a single newline, for convention.

sources/es.manhwaweb/res/icon.png
kkantan, line 1: the corner radius on this icon should be removed if possible, since the icon could be displayed at any corner radius by aidoku. you could fill in the edges with the background color.

sources/es.manhwaweb/src/helper.rs
```diff
@@ -0,0 +1,15 @@
+use aidoku::{
+    imports::{net::Request, std::sleep},
+    Result,
+};
+
+/// Enforces a rate limit by sleeping before returning the request object.
+/// This prevents spamming the server and getting IP banned.
+pub fn request_with_limits(url: &str, method: &str) -> Result<Request> {
+    // Sleep for 1 second to respect rate limits.
+    sleep(1);
```
kkantan: instead, could you use `set_rate_limit`? this sets a specific rate limit in a given period. for example, 1 request allowed every 2 seconds. for reference, the mangadex source uses this.

sources/es.manhwaweb/src/helper.rs
```diff
@@ -0,0 +1,15 @@
...
+    match method {
+        "POST" => Request::post(url).map_err(|e| e.into()),
```
kkantan: I think you can do this.
```suggestion
        "POST" => Request::post(url).map_err(Into::into),
```

sources/es.manhwaweb/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", version = "0.3.0", features = ["json"] }
+iken = { path = "../../templates/iken" }
+serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
+serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+chrono = { version = "0.4", default-features = false, features = ["alloc"] }
+
+[dev-dependencies]
+aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", features = ["test"] }
+aidoku-test = { git = "https://github.com/Aidoku/aidoku-rs.git" }
+
```
kkantan, line 14:
```suggestion
```

sources/es.manhwaweb/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+name = "manhwaweb"
+version = "1.0.0"
+edition = "2024"
+
+[dependencies]
+aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", version = "0.3.0", features = ["json"] }
+iken = { path = "../../templates/iken" }
+serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
+serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+chrono = { version = "0.4", default-features = false, features = ["alloc"] }
```
kkantan: you're linking the iken template and chrono, but it doesn't look like you use them.

sources/es.manhwaweb/src/imp.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    fn get_search_manga_list(
+        &self,
+        query: Option<String>,
+        page: i32,
+        filters: Vec<FilterValue>,
+    ) -> Result<MangaPageResult> {
+        // API is 0-based, Aidoku is 1-based.
+        let api_page = if page > 0 { page - 1 } else { 0 };
```
kkantan: the page should never be 0, so you could just do `page - 1` without the branching.

sources/es.manhwaweb/src/imp.rs
```diff
@@ -0,0 +1,294 @@
...
+        // Default to "no" erotic content if the filter wasn't explicitly set
+        if !erotic_filter_set {
+            qs.push("erotico", Some("no"));
+        }
+        
+        // Ensure qs is appended with '&' prefix because base URL might have query params already.
+        let qs_str = format!("{}", qs);
+        if !qs_str.is_empty() {
+            url.push_str(&format!("&{}", qs_str));
+        }
```
kkantan: you can check if the params are empty first without allocating an extra string.
```suggestion
        if !qs.is_empty() {
            url.push_str(&format!("&{}", qs));
        }
```

sources/es.manhwaweb/src/imp.rs
```diff
@@ -0,0 +1,294 @@
...
+        &self,
+        mut manga: Manga,
+        needs_details: bool,
+        needs_chapters: bool,
+    ) -> Result<Manga> {
+        let url = format!("{BACKEND_URL}/manhwa/see/{}", manga.key);
+        let mut response = crate::helper::request_with_limits(&url, "GET")?
+            .header("Referer", &format!("{}/", BASE_URL))
+            .send()?;
+        let data = response.get_json::<SeeResponse>()?;
```
kkantan: the `Request` struct has a function for this you could use.
```suggestion
        let data = crate::helper::request_with_limits(&url, "GET")?
            .header("Referer", &format!("{}/", BASE_URL))
            .json_owned::<SeeResponse>()()?;
```

sources/es.manhwaweb/src/imp.rs
kkantan, line 1: it looks like some of the indentation is a little off. you can run `cargo fmt` to format everything correctly.

sources/es.manhwaweb/src/imp.rs
```diff
@@ -0,0 +1,294 @@
...
+                return Ok(Some(DeepLinkResult::Manga {
+                    key: id.into(),
+                }));
+            }
+         }
+         Ok(None)
+    }
+}
+
+use aidoku::alloc::string::ToString;
```
kkantan: this should be up at the top of the file along with the rest of the namespaces.

sources/es.manhwaweb/src/imp.rs
```diff
@@ -0,0 +1,294 @@
...
+        "3" => "Acción", "29" => "Aventura", "18" => "Comedia", "1" => "Drama",
+        "42" => "Recuentos de la vida", "2" => "Romance", "5" => "Venganza", "6" => "Harem",
+        "23" => "Fantasía", "31" => "Sobrenatural", "25" => "Tragedia", "43" => "Psicológico",
+        "32" => "Horror", "44" => "Thriller", "28" => "Historias cortas", "30" => "Ecchi",
+        "34" => "Gore", "37" => "Girls love", "27" => "Boys love", "45" => "Reencarnación",
+        "41" => "Sistema de niveles", "33" => "Ciencia ficción", "38" => "Apocalíptico",
+        "39" => "Artes marciales", "40" => "Superpoderes", "35" => "Cultivación (cultivo)",
+        "8" => "Milf", _ => "Desconocido",
+    }.to_string()
+}
```
kkantan: this should probably be moved into the helpers module.
```suggestion
/// Helper function to map genre IDs to Spanish names.
fn get_genre_name(id: &str) -> String {
    match id {
        "3" => "Acción", "29" => "Aventura", "18" => "Comedia", "1" => "Drama",
        "42" => "Recuentos de la vida", "2" => "Romance", "5" => "Venganza", "6" => "Harem",
        "23" => "Fantasía", "31" => "Sobrenatural", "25" => "Tragedia", "43" => "Psicológico",
        "32" => "Horror", "44" => "Thriller", "28" => "Historias cortas", "30" => "Ecchi",
        "34" => "Gore", "37" => "Girls love", "27" => "Boys love", "45" => "Reencarnación",
        "41" => "Sistema de niveles", "33" => "Ciencia ficción", "38" => "Apocalíptico",
        "39" => "Artes marciales", "40" => "Superpoderes", "35" => "Cultivación (cultivo)",
        "8" => "Milf", _ => "Desconocido",
    }.into()
}
```

sources/es.manhwaweb/src/helper.rs
kkantan, line 1: you should rename this to `helpers.rs` and move the other helper function(s) here.

sources/es.manhwaweb/src/imp.rs
```diff
@@ -0,0 +1,294 @@
...
+                    let latest_entries: Vec<Manga> = data.manhwas.spanish_manhwas.iter().map(|m| {
+                        let group = m.gru_name.clone().unwrap_or_default();
+                        let subtitle = format!("Cap. {} • {}", m.chapter, group);
+
+                         Manga {
+                            key: m.id_manhwa.clone().into(),
+                            title: m.name_manhwa.clone(),
+                            authors: Some(vec![subtitle]), 
+                            cover: m.img.clone().map(|s| s.into()),
+                            url: Some(format!("{}/manhwa/{}", BASE_URL, m.id_manhwa)),
```
kkantan: you shouldn't need to clone anything here. if you define the url above, before you use the key, that clone can be avoided, and if you use `into_iter` instead of `iter` that should allow the other two to be avoided.

sources/es.manhwaweb/src/imp.rs
```diff
@@ -0,0 +1,294 @@
...
+impl ListingProvider for ManhwaWeb {
+    fn get_manga_list(&self, listing: Listing, page: i32) -> Result<MangaPageResult> {
+        let name = listing.id.as_str(); 
+        let api_page = if page > 0 { page - 1 } else { 0 };
+
+        if name.starts_with("Genre:") {
+             let id = name.split(':').nth(1).unwrap_or("");
+             // Ensure 'creacion' (creation date) sort is properly applied for these listings too
+             let url = format!("{BACKEND_URL}/manhwa/library?page={}&perPage={}&erotico=no&generes={}&order_item=creacion&order_dir=desc", api_page, PER_PAGE, id);
+             let resp = crate::helper::request_with_limits(&url, "GET")?.send()?.get_json::<LibraryResponse>()?;
```
kkantan:
```suggestion
             let resp = crate::helper::request_with_limits(&url, "GET")?.json_owned::<LibraryResponse>()?;
```

sources/es.manhwaweb/src/models.rs
```diff
@@ -0,0 +1,175 @@
...
+    #[serde(rename = "_status")]
+    pub status: Option<String>,
+    #[serde(rename = "_erotico")]
+    pub erotic: Option<String>,
+    #[serde(rename = "_categoris")]
+    pub categories: Option<Vec<i32>>,
+}
+
+impl LibraryManga {
+    pub fn to_manga(&self, base_url: &str) -> Manga {
```
kkantan: this function should probably be consuming, and then you can avoid the `clone()`s. just make sure to use `into_iter` when collecting.
```suggestion
    pub fn to_manga(self, base_url: &str) -> Manga {
```

sources/es.manhwaweb/src/models.rs
```diff
@@ -0,0 +1,189 @@
...
+	#[serde(rename = "_status")]
+	pub status: Option<String>,
+	#[serde(rename = "_erotico")]
+	pub erotic: Option<String>,
+	#[serde(rename = "_categoris")]
+	pub categories: Option<Vec<i32>>,
+}
+
+impl LibraryManga {
+	pub fn to_manga(self, base_url: &str) -> Manga {
```
kkantan, line 59: I believe that prefixing this with `into_` instead will fix the clippy warning.
```suggestion
	pub fn into_manga(self, base_url: &str) -> Manga {
```

sources/es.manhwaweb/src/imp.rs
```diff
@@ -0,0 +1,339 @@
...
+				.into_iter()
+				.map(|m| {
+					let group = m.gru_name.unwrap_or_default();
+					let subtitle = format!("Cap. {} - {}", m.chapter, group);
+					let url = Some(format!("{}/manhwa/{}", BASE_URL, m.id_manhwa));
+
+					Manga {
+						key: m.id_manhwa.into(),
+						title: m.name_manhwa,
+						authors: Some(vec![subtitle]),
```
kkantan, line 186: I think this is best to put as the description instead.

sources/es.manhwaweb/src/imp.rs
```diff
@@ -0,0 +1,339 @@
...
+				}
+				FilterValue::MultiSelect { id, included, .. } => {
+					// For genres
+					if id == "genres" || id == "genreIds" || id == "generes" {
+						for val in included {
+							genre_values.push(val);
+						}
+					}
+				}
+				FilterValue::Sort { index, .. } => {
```
kkantan, line 68: since you don't use the ascending value, you should set `canAscend` to false in the `filters.json`.

sources/es.manhwaweb/src/imp.rs
```diff
@@ -0,0 +1,339 @@
...
+
+		// Handle Genres: joined by 'a' (e.g., "1a2a3")
+		if !genre_values.is_empty() {
+			let joined_genres = genre_values.join("a");
+			qs.push("generes", Some(&joined_genres));
+		}
+
+		// Default to "no" erotic content if the filter wasn't explicitly set
+		if !erotic_filter_set {
+			qs.push("erotico", Some("no"));
```
kkantan, line 88: this is the same thing for both the "no" option and the "ver todo" option.

since the source rating is set to 18+, and there's an 18+ listing, it seems like the default option should be "ver todo", if it can be made to work.

Llavesuke: I'd still prefer not to show 18+ content by default and keep it as an optional choice for the user. If we go with that approach, should we adjust how the source rating is handled or labeled to make it more consistent with a 'hidden by default' policy?

kkantan: in that case, you should remove the 18+ listing and then change the `contentRating` in the source.json to 1 instead of 2.

PR #273, opened 2026-01-04
fix(en.mangapark): add 'nsfw' cookie header to HTTP requests

### LINKED ISSUE #189: MangaPark +18 problem
Hello,

I have a question about the MangaPark source on Aidoku. On the MangaPark website, there are NSFW (+18) titles that become accessible when you enable the corresponding option in the site settings. However, in Aidoku, I can’t find any way to view or enable these NSFW contents within the MangaPark source.

Is it possible to add an option or setting in the source so that NSFW (+18) titles can be displayed in Aidoku as they are on the website? Or, if this is already supported, could you explain how to enable it?

Thank you!

vikas5914:
Fix #189

### DIFF
--- a/sources/en.mangapark/res/source.json
+++ b/sources/en.mangapark/res/source.json
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "en.mangapark",
 		"name": "MangaPark",
-		"version": 2,
+		"version": 3,
 		"url": "https://mangapark.com",
 		"contentRating": 1,
 		"languages": ["en"]


kkantan (changes requested):
thank you for the pr! but I think it would probably be better to match [the mihon source](https://github.com/keiyoushi/extensions-source/blob/8dffd09b08d4d4a2eb0b1875189f02ec3be1c2ec/src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaPark.kt#L58-L61) and conditionally have the header when the user turns on an "Enable NSFW Content" source setting.

also, you'll need to increment the version number in `source.json` to mark this as an update to the source.

sources/en.mangapark/Cargo.toml
kkantan, line 1: this version number doesn't matter, it's the version number in the `source.json` file that needs to be incremented by one.

PR #286, opened 2026-01-09
feat: add zh.zaimanhua

Aiko3993:
Add Zaimanhua (再漫画) Chinese manga source.

Features:
- Home page with editorial picks, rankings, and audience categories
- Filters: status, categories, demographic, rank mode
- User login with auto check-in


sources/zh.zaimanhua/res/filters.json
kkantan, line 1: json files should use tabs for indentation and have an extra trailing newline for consistency.

sources/zh.zaimanhua/src/json.rs
kkantan, line 1: personally, I find using the `serde::Deserialize` trait and implementing `From<...> for Manga` and so on for the model structs to be a bit easier and less repetitive, but this should still work fine.

sources/zh.zaimanhua/src/net.rs
```diff
@@ -0,0 +1,82 @@
...
+    let digest = md5::compute(input.as_bytes());
+    format!("{:x}", digest)
+}
+
+pub fn get_request(url: &str) -> Result<Request> {
+    Ok(Request::get(url)?.header("User-Agent", USER_AGENT))
+}
+
+pub fn post_request(url: &str) -> Result<Request> {
+    Ok(Request::new(url, HttpMethod::Post)?
```
kkantan: you can use the shorthand function for post as well.
```suggestion
    Ok(Request::post(url)?
```

sources/zh.zaimanhua/src/settings.rs
```diff
@@ -0,0 +1,73 @@
...
+pub fn get_enhanced_mode() -> bool {
+    defaults_get::<bool>(ENHANCED_MODE_KEY).unwrap_or(false)
+}
+
+// === Clear Data ===
+
+/// Clear all user data - called on logout
+/// Note: Due to Aidoku framework limitations, this sets values to empty/false
+/// rather than truly deleting them. User must use "Reset Source Settings" 
+/// in Aidoku for complete removal.
```
kkantan: does setting `DefaultValue::Null` not clear the setting? I assumed that's what it would do, but I don't actually know for certain.

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+
+pub const BASE_URL: &str = "https://www.zaimanhua.com/";
+const V4_API_URL: &str = "https://v4api.zaimanhua.com/app/v1";
+
+
+/// Helper function to create a GET request with optional authentication
+/// Authentication is only added if:
+/// 1. User is logged in (has token)
+/// 2. Enhanced mode is enabled (user explicitly turned it on)
+fn get_api_request(url: &str) -> Result<Request> {
```
kkantan: I would personally put these extra helper functions in a `helpers.rs` module, as is the convention with most other sources.

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+/// Search manga by keyword
+fn search_by_keyword(keyword: &str, page: i32) -> Result<MangaPageResult> {
+    let encoded = encode_uri_component(keyword);
+    let url = format!(
+        "{}/search/index?keyword={}&source=0&page={}&size=20",
+        V4_API_URL, encoded, page
+    );
+
+    let mut response = get_api_request(&url)?.send()?;
+    let json_val: serde_json::Value = response.get_json()?;
```
kkantan: you could just use `.json_owned()` on the request instead of `.send()`, as it should do the same thing.

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+    
+    for filter in filters {
+        if let FilterValue::Select { id, value } = filter {
+            match id.as_str() {
+                "排序" => sort_type = value.clone(),
+                "地区" => zone = value.clone(),
+                "状态" => status = value.clone(),
+                "受众" => cate = value.clone(),
+                "题材" => theme = value.clone(),
+                "榜单" => rank_mode = value.clone(),
```
kkantan: I don't believe you should need to clone these values. it looks like `filters` isn't used after this function is called anyways, so you could consume the `Vec<FilterValue>` instead of a slice.

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+            }
+        }
+    }
+    
+    // If rank mode is selected, use rank API (ignores other filters)
+    let url = match rank_mode.as_str() {
+        "1" => format!("{}/comic/rank/list?rank_type=0&by_time=0&page={}&size=20", V4_API_URL, page), // 日榜
+        "2" => format!("{}/comic/rank/list?rank_type=0&by_time=1&page={}&size=20", V4_API_URL, page), // 周榜
+        "3" => format!("{}/comic/rank/list?rank_type=0&by_time=2&page={}&size=20", V4_API_URL, page), // 月榜
+        "4" => format!("{}/comic/rank/list?rank_type=0&by_time=3&page={}&size=20", V4_API_URL, page), // 总榜
```
kkantan: since you don't use any of the filters in these branches, it'd be better to move the parsing from above into the final branch so the strings are only allocated if necessary.

another optimization to consider would be to use `Option<String>`s instead, and start the variables as `None`, so that if a value wasn't assigned you can then unwrap with the default value instead of always allocating them even when unnecessary.

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+    let mut keyword_manga: Vec<serde_json::Value> = Vec::new();
+    let mut seen_authors: Vec<String> = Vec::new();
+    
+    // Step 1: Search for manga by author name
+    let search_url = format!("{}/search/index?keyword={}&source=0&page=1&size=50", V4_API_URL, encoded);
+    
+    // Use authenticated request to access restricted content
+    if let Ok(mut resp) = get_api_request(&search_url)?.send()
+        && let Ok(json) = resp.get_json::<serde_json::Value>()
+            && let Some(list) = json.get("data").and_then(|d| d.get("list")).and_then(|l| l.as_array()) {
```
kkantan: instead of doing large let chains like this, you could just use `and_then` instead, since you seem to already be doing so.

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+                for manga in list {
+                    let manga_authors = manga.get("authors").and_then(|a| a.as_str()).unwrap_or("");
```
kkantan:
```suggestion
                    let manga_authors = manga.get("authors").and_then(|a| a.as_str()).unwrap_or_default();
```

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+                    
+                    if author_matches(manga_authors) {
+                        keyword_manga.push(manga.clone());
```
kkantan: could you move this to the bottom of the `if` branch so that cloning isn't necessary?

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+        if id > 0 && !seen_ids.contains(&id) {
+            seen_ids.push(id);
+            final_manga.push(m.clone());
+        }
+    }
+    
+    if !final_manga.is_empty() {
+        let fv = serde_json::Value::Array(final_manga.clone());
+        let mut res = json::parse_manga_list(&fv)?;
+        res.has_next_page = if tag_total > 0 { (page * 100) < tag_total } else { final_manga.len() >= 100 };
```
kkantan: if you store the value assigned to `res.has_next_page` above line 257 here, you shouldn't need to clone `final_manga`.

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+        return Ok(res);
+    }
+    
+    Ok(MangaPageResult { entries: Vec::new(), has_next_page: false })
```
kkantan: does `MangaPageResult` have a default implementation? I forget.
```suggestion
    Ok(MangaPageResult::default())
```

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+
+        if password.is_empty() {
+            return Ok(false);
+        }
+
+        // Clear old account data before logging in with new credentials
+        settings::clear_all();
+        
+        settings::set_username(&username);
+        settings::set_password(&password);
```
kkantan: you shouldn't need to manually manage the username and password settings. they're stored in `{loginkey}.username` and `{loginkey}.password` by the app automatically after a successful login, where `{loginkey}` is the key for the login setting.

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,717 @@
...
+    fn handle_basic_login(&self, key: String, username: String, password: String) -> Result<bool> {
+        if key != "login" {
+            bail!("Invalid login key: `{key}`");
+        }
+
+        // Handle logout (empty username means logout)
+        if username.is_empty() {
+            settings::clear_all();
+            return Ok(true);
+        }
```
kkantan: you shouldn't need to clear settings for an empty username. in the first place, doesn't the app disable submitting the login form if the username is left empty? anyways, you could implement `NotificationHandler` to handle a logout; I believe the gigaviewer template does this for reference.

Aiko3993: > you shouldn't need to clear settings for an empty username. in the first place, doesn't the app disable submitting the login form if the username is left empty? anyways, you could implement `NotificationHandler` to handle a logout; I believe the gigaviewer template does this for reference.

This came from over-reliance on AI assistance, and I missed some basic behavior. That’s on me — I’ll fix it.

Aiko3993:
> if you copy the `.cargo/config.toml` directory and file in the other sources, it will properly compile your source without the workflow modifications. you also probably don't need the customized `rustfmt.toml` params just for this source.

Thanks for the review. The unnecessary changes came from not properly studying existing sources first, combined with a messy local worktree during development. I should have aligned with project conventions before starting. Cleaning up now.

PR #289, opened 2026-01-11
feat: add zh.zaimanhua

Aiko3993:
Resubmission after workflow corrections.

P.S. Vibe coding: never again. 🫠

### DIFF
--- a/sources/zh.zaimanhua/res/settings.json
+++ b/sources/zh.zaimanhua/res/settings.json
@@ -1,85 +1,81 @@
 [
-    {
-        "type": "group",
-        "key": "account",
-        "title": "账号",
-        "items": [
-            {
-                "type": "login",
-                "key": "login",
-                "title": "登录",
-                "logoutTitle": "登出",
-                "method": "basic",
-                "notification": "login",
-                "refreshes": [
-                    "settings",
-                    "content",
-                    "listings"
-                ]
-            },
-            {
-                "type": "switch",
-                "key": "autoCheckin",
-                "title": "自动签到",
-                "default": false,
-                "requires": "login"
-            },
-            {
-                "type": "switch",
-                "key": "enhancedMode",
-                "title": "增强浏览",
-                "subtitle": "获取更多内容 (Lv.1)",
-                "default": false,
-                "requires": "login",
-                "refreshes": [
-                    "content",
-                    "listings",
-                    "settings"
-                ]
-            },
-            {
-                "type": "switch",
-                "key": "showHiddenContent",
-                "title": "隐藏内容",
-                "subtitle": "搜索和浏览时包含隐藏漫画 (Lv.3)",
-                "default": false,
-                "requires": "enhancedMode",
-                "refreshes": [
-                    "content",
-                    "listings"
-                ]
-            }
-        ]
-    },
-    {
-        "type": "group",
-        "key": "advanced",
-        "title": "高级",
-        "items": [
-            {
-                "type": "switch",
-                "key": "useProxy",
-                "title": "使用代理",
-                "subtitle": "通过代理中转请求",
-                "default": false,
-                "refreshes": [
-                    "settings"
-                ]
-            },
-            {
-                "type": "text",
-                "key": "proxyUrl",
-                "title": "代理地址",
-                "placeholder": "https://your-worker.workers.dev",
-                "keyboardType": 10,
-                "autocapitalizationType": 0,
-                "autocorrectionDisabled": true,
-                "requires": "useProxy",
-                "refreshes": [
-                    "content",
-                    "listings"
-                ]
-            }
-        ]
-    }
+	{
+		"type": "group",
+		"key": "account",
+		"title": "账号",
+		"items": [
+			{
+				"type": "login",
+				"key": "login",
+				"title": "登录",
+				"logoutTitle": "登出",
+				"method": "basic",
+				"notification": "login",
+				"refreshes": [
+					"settings",
+					"content"
+				]
+			},
+			{
+				"type": "switch",
+				"key": "autoCheckin",
+				"title": "自动签到",
+				"default": false,
+				"requires": "login"
+			},
+			{
+				"type": "switch",
+				"key": "enhancedMode",
+				"title": "增强浏览",
+				"subtitle": "获取更多内容 (Lv.1)",
+				"default": false,
+				"requires": "login",
+				"refreshes": [
+					"content",
+					"settings"
+				]
+			},
+			{
+				"type": "switch",
+				"key": "showHiddenContent",
+				"title": "隐藏内容",
+				"subtitle": "搜索和浏览时包含隐藏漫画 (Lv.3)",
+				"default": false,
+				"requires": "enhancedMode",
+				"refreshes": [
+					"content"
+				]
+			}
+		]
+	},
+	{
+		"type": "group",
+		"key": "advanced",
+		"title": "高级",
+		"items": [
+			{
+				"type": "switch",
+				"key": "useProxy",
+				"title": "使用代理",
+				"subtitle": "通过代理中转请求",
+				"default": false,
+				"refreshes": [
+					"settings"
+				]
+			},
+			{
+				"type": "text",
+				"key": "proxyUrl",
+				"title": "代理地址",
+				"placeholder": "https://your-worker.workers.dev",
+				"keyboardType": 10,
+				"autocapitalizationType": 0,
+				"autocorrectionDisabled": true,
+				"requires": "useProxy",
+				"refreshes": [
+					"content"
+				]
+			}
+		]
+	}
 ]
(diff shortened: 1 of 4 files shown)


sources/zh.zaimanhua/res/source.json
```diff
@@ -0,0 +1,42 @@
...
+		{
+			"id": "subscribe",
+			"name": "订阅"
+		}
+	],
+	"config": {
+		"hidesFiltersWhileSearching": true,
+		"supportsAuthorSearch": true
+	}
+}
```
kkantan: this file needs an extra newline at the end.
```suggestion
}

```

sources/zh.zaimanhua/src/home.rs
```diff
@@ -0,0 +1,390 @@
...
+					authors: Some(vec![item.sub_title.clone().unwrap_or_default()]),
+					cover: Some(item.cover.clone().unwrap_or_default()),
+					status: MangaStatus::Unknown,
+					..Default::default()
+				})
+				.collect();
+
+			if manga_list.is_empty() {
+				continue;
+			}
```
kkantan: I'm confused why this `manga_list` is being created when it doesn't seem to be used. you can more easily check if the `cat.data` is empty, but regardless it should only need to happen if `cat.category_id` is 109, since that's a much less expensive check.

sources/zh.zaimanhua/src/home.rs
```diff
@@ -0,0 +1,390 @@
...
+				big_scroller_manga = cat.data.iter()
+					// Filter only Manga type (1) to avoid Topics/Ads
+					.filter(|item| item.obj_id > 0 && item.item_type == 1)
+					.map(|item| {
+						let mut real_title = item.title.clone();
+						let mut manga_cover = item.cover.clone().unwrap_or_default();
+
+						// Fetch details for high-res assets
+						// Try to update title/cover from detail API
+						if let Ok(req) = Request::get(format!("https://v4api.zaimanhua.com/app/v1/comic/detail/{}", item.obj_id))
```
kkantan: you could use the `V4_API_URL` const.

sources/zh.zaimanhua/src/home.rs
```diff
@@ -0,0 +1,390 @@
...
+				})
+				.collect();
+
+			if manga_list.is_empty() {
+				continue;
+			}
+
+			// Only handle category 109 (Premium Recommend) as BigScroller
+			if cat.category_id == 109 {
+				big_scroller_manga = cat.data.iter()
```
kkantan: use `into_iter()` and avoid cloning.

sources/zh.zaimanhua/src/settings.rs
```diff
@@ -0,0 +1,74 @@
...
+
+// === Enhanced Mode & Hidden Content ===
+
+pub fn get_enhanced_mode() -> bool {
+	defaults_get::<bool>(ENHANCED_MODE_KEY).unwrap_or(false) && get_token().is_some()
+}
+
+pub fn show_hidden_content() -> bool {
+	get_enhanced_mode() && defaults_get::<bool>(SHOW_HIDDEN_KEY).unwrap_or(false)
+}
```
kkantan:
```suggestion
}

```

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+		if needs_chapters {
+			manga.chapters = Some(manga_detail.into_chapters(&manga.key));
+		}
+
+		Ok(manga)
+	}
+
+	fn get_page_list(&self, manga: Manga, chapter: Chapter) -> Result<Vec<Page>> {
+		let parts: Vec<&str> = chapter.key.split('/').collect();
+		let (comic_id, chapter_id) = if parts.len() == 2 {
```
kkantan: you could use `split_once` instead.

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+		}
+	}
+}
+
+// === Notification Logic ===
+// Handle async events like check-in or login state changes.
+impl NotificationHandler for Zaimanhua {
+	fn handle_notification(&self, notification: String) {
+		match notification.as_str() {
+			"checkin" => {
```
kkantan: where does this notification get posted from?

sources/zh.zaimanhua/src/models.rs
```diff
@@ -0,0 +1,443 @@
...
+	pub last_update_chapter_id: Option<i64>,
+	pub last_updatetime: Option<i64>,
+}
+
+impl FilterItem {
+	pub fn into_manga_with_chapter(self) -> MangaWithChapter {
+		// Extract chapter-specific data before consuming self
+		let chapter_id = self.last_update_chapter_id.unwrap_or(0).to_string();
+		let chapter_key = format!("{}/{}", self.id, chapter_id);
+		let chapter_title = self.last_update_chapter_name.clone();
```
kkantan: this shouldn't need a clone.

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+			let details = manga_detail.clone().into_manga(manga.key.clone());
+			manga.title = details.title;
+			manga.cover = details.cover;
+			manga.description = details.description;
+			manga.authors = details.authors;
+			manga.tags = details.tags;
+			manga.status = details.status;
+			manga.content_rating = details.content_rating;
+			manga.viewer = details.viewer;
+			manga.url = details.url;
```
kkantan: you could use the `copy_from` function instead.

sources/zh.zaimanhua/src/home.rs
```diff
@@ -0,0 +1,390 @@
...
+
+	let [resp_recommend, resp_latest, resp_rank, resp_shounen, resp_shoujo, resp_seinen, resp_josei, mut resp_news] = responses;
+
+	let mut components = Vec::new();
+
+	let mut big_scroller_manga: Vec<Manga> = Vec::new(); // For 109
+	let mut banner_links: Vec<aidoku::Link> = Vec::new();
+
+	if let Ok(ref mut resp) = resp_news
+		&& let Ok(html) = resp.get_string()
```
kkantan: you could just use the `.html()` function on the response to parse the html rather than putting it into a string first.

Aiko3993:
> I tried testing the source, but none of the images load for me. it's likely a regional issue though, since the site itself doesn't load for me either. do you know if there are any ways to access the site via vpn?

I've implemented a proxy feature using Cloudflare Workers to help with testing. You can try out the source on this branch:
https://github.com/Aiko3993/sources/tree/add-zaimanhua-test-cfproxy

Here are the credentials for testing:
Username: test_user
Password: password
*(Note: This is a Level 0 account, so Enhanced Mode is unavailable.)*

sources/zh.zaimanhua/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+
+		// Login (with logout notification)
+		account_items.push(
+			LoginSetting {
+				key: "login".into(),
+				title: "登录".into(),
+				logout_title: Some("登出".into()),
+				notification: Some("login".into()), // Fires on login state change (GigaViewer pattern)
+				method: LoginMethod::Basic,
+				refreshes: Some(vec!["settings".into(), "content".into(), "listings".into()]),
```
kkantan: I believe the "listings" refresh is only for if you're dynamically providing listings, like the mangadex source does. you should be able to leave it off.

kkantan (changes requested):
I think the code looks pretty good now, I just have a few thoughts on the source:
- the image banner scroller never loads for me, and leaves an extra space at the top of the home view that doesn't look great; is this just a regional thing?
- there is an app crash that seems to happen occasionally when using the source, but I don't know the trigger or how exactly it would be found. I don't really see anything in the code that would obviously cause this, so my guess would be it's an app issue that needs to be solved.
- sometimes the settings don't load, or are slow to load, which I think is a side effect of using dynamic settings and why they're typically discouraged. I believe it's possible to combine both static and dynamic settings, as the dynamic ones should always just be appended to the bottom. so it might be worth moving some settings, at least the login and checkin settings that don't change, to static versions in a `settings.json`.

sources/zh.zaimanhua/res/settings.json
```diff
@@ -0,0 +1,85 @@
...
+                "type": "login",
+                "key": "login",
+                "title": "登录",
+                "logoutTitle": "登出",
+                "method": "basic",
+                "notification": "login",
+                "refreshes": [
+                    "settings",
+                    "content",
+                    "listings"
```
kkantan: as said [here](https://github.com/Aidoku-Community/sources/pull/289#discussion_r2688230899), I don't believe any of the settings should need to refresh listings. the listing content is refreshed with the "content" refresh.

sources/zh.zaimanhua/res/settings.json
kkantan, line 1: this file should use tabs for indentation, like the other json files.

Aiko3993: > this file should use tabs for indentation, like the other json files.

This looks odd, the editor doesn’t change at all after I use the `Convert Indentations to Tabs`. This is the first time I’ve seen this happen — kind of hilarious, honestly.
I’ll fix them as soon as possible.

Aiko3993: > > this file should use tabs for indentation, like the other json files.
> 
> This looks odd — I did use tabs when writing it. Maybe the indentation was altered by AI autocomplete? I’ll fix them as soon as possible.

Not AI — it turns out my editor rewrites indentation every time I save `settings.json`. This is the first time I’ve seen this happen — kind of hilarious, honestly.

PR #300, opened 2026-01-16
fix(zh.zaimanhua): fix known bugs and reorganize code structure

Aiko3993:
Excited to have this merged—this should be the final piece of the puzzle.

Changes in this update include:

- Fix slow loading issues
- Resolve `get_search_manga_list` broken after refresh
- Correct description in Source Settings
- Partially reorganize code structure

### DIFF
--- a/sources/zh.zaimanhua/src/lib.rs
+++ b/sources/zh.zaimanhua/src/lib.rs
@@ -8,7 +8,7 @@ use aidoku::{
 	Result, Setting, Source,
 	alloc::{String, Vec, format, string::ToString},
 	helpers::uri::QueryParameters,
-	imports::net::{Request, set_rate_limit, TimeUnit},
+	imports::net::Request,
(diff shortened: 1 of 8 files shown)


sources/zh.zaimanhua/src/lib.rs
```diff
@@ -33,7 +33,7 @@ struct Zaimanhua;
 
 impl Source for Zaimanhua {
 	fn new() -> Self {
-		set_rate_limit(5, 1, TimeUnit::Seconds); // 5 requests per second
+		set_rate_limit(100, 1, TimeUnit::Seconds);
```
kkantan: this rate limit is so high that I don't think it's worth setting.

Aiko3993: > this rate limit is so high that I don't think it's worth setting.

I’ll remove it.

PR #323, opened 2026-02-01
Fix issue with comix formatting value from json

dangobums:


kkantan:
oh, my apologies. I didn't see this pr and opened a separate one with a fix: #324. it looks like we both did essentially the same thing, so I'll just close this one.

Issue #350, opened 2026-02-14
MANGA Plus mobile API option


tcK1:
Would it be possible to allow configuring the authentication parameters to see the full chapter list (app exclusive)?

Although it's not easy to get the parameters, it is possible. https://github.com/hurlenko/mloader/issues/36

kkantan:
since it appears as though the app and web api have differing available for chapters, with some being available on the web api and not the mobile api, it would be ideal to have a setting that enables usage of the mobile api in place of the web api.

however, it looks like the mobile api uses protobuf, which I believe isn't the most convenient to work with no_std rust. I probably won't work on this myself.

tcK1:
first contribution, sorry if something is wrong! https://github.com/Aidoku-Community/sources/pull/352

PR #352, opened 2026-02-15
feat(multi.mangaplus): mobile api option

tcK1:
attempt to solve https://github.com/Aidoku-Community/sources/issues/350

### DIFF
--- a/sources/multi.mangaplus/src/lib.rs
+++ b/sources/multi.mangaplus/src/lib.rs
@@ -11,9 +11,9 @@ use aidoku::{
 use core::cell::RefCell;
 use hashbrown::HashSet;
 
+mod helpers;
 mod models;
 mod settings;
-mod helpers;
 
 use models::{MangaPlusResponse, Title};
 
@@ -60,14 +60,22 @@ impl Source for MangaPlus {
 			let url = || {
 				if let Some(query) = query.as_ref() {
 					if let Some(title_id) = query.strip_prefix("id:") {
-						return format!("{}/title_detailV3?title_id={title_id}&format=json{}", helpers::get_api_url(), helpers::build_auth_params());
+						return format!(
+							"{}/title_detailV3?title_id={title_id}&format=json{}",
+							helpers::get_api_url(),
+							helpers::build_auth_params()
+						);
 					} else if let Some(chapter_id) = query.strip_prefix("chapter-id:") {
 						return format!(
 							"{WEB_API_URL}/manga_viewer?chapter_id={chapter_id}&split=no&img_quality=low&format=json"
 						);
 					}
 				}
-				format!("{}/title_list/allV2?format=json{}", helpers::get_api_url(), helpers::build_auth_params())
+				format!(
+					"{}/title_list/allV2?format=json{}",
+					helpers::get_api_url(),
+					helpers::build_auth_params()
+				)
 			};
 
 			let result = Request::get(url())?
--- a/sources/multi.mangaplus/src/models.rs
+++ b/sources/multi.mangaplus/src/models.rs
@@ -151,14 +151,14 @@ impl TitleDetailView {
 			self.chapter_list_v2.iter().collect()
 		} else {
 			self.chapter_list_group
-			.iter()
-			.flat_map(|group| {
-				group
-					.first_chapter_list
-					.iter()
-					.chain(group.last_chapter_list.iter())
-			})
-			.collect()
+				.iter()
+				.flat_map(|group| {
+					group
+						.first_chapter_list
+						.iter()
+						.chain(group.last_chapter_list.iter())
+				})
+				.collect()
 		}
 	}
 


sources/multi.mangaplus/res/settings.json
```diff
@@ -16,6 +16,41 @@
...
+				"placeholder": "app_ver (e.g., 235)",
+				"refreshes": ["listings", "content"]
+			},
+			{
+				"type": "text",
+				"key": "secret",
+				"title": "secret",
+				"placeholder": "secret (hash value)",
+				"refreshes": ["listings", "content"]
 			}
```
kkantan, line 64: these options should all require "mobile", so that they are enabled only if the mobile option is turned on.

sources/multi.mangaplus/src/lib.rs
```diff
@@ -17,10 +17,29 @@ mod settings;
...
+
+fn build_auth_params() -> String {
+	format!(
+		"&os={}&os_ver={}&app_ver={}&secret={}",
+		settings::get_os(),
+		settings::get_os_ver(),
+		settings::get_app_ver(),
+		settings::get_secret()
+	)
+}
```
kkantan: it would be nice to put these two functions inside a `helpers.rs` file, similar to other sources.

also, these parameters are only necessary when using the mobile api, correct? it would be best to only add them to the url if the option is enabled.

sources/multi.mangaplus/res/settings.json
tcK1: https://github.com/Aidoku-Community/sources/pull/352/changes/56ed71202e8386d21d355673116f9741749b53a7

sources/multi.mangaplus/src/lib.rs
tcK1: https://github.com/Aidoku-Community/sources/pull/352/changes/81421085a58165d77a881e73cfc3d3ea11d5721d and https://github.com/Aidoku-Community/sources/pull/352/changes/67d0b36b5429227c6bc5e3172c336a0e3550f564

sources/multi.mangaplus/res/settings.json
```diff
@@ -16,7 +16,52 @@
 				"key": "split",
 				"title": "Split Double Pages",
 				"default": false
-			}
+			},
+            {
+                "type": "group",
+                "title": "MOBILE API",
```
kkantan: a group can't be embedded inside another group. you should move this outside the items array it's in, and then it will work properly.

the "SETTINGS" title is also capitalized because of there being a matching localization, but since there isn't one for this case, you should use title case instead: "Mobile API".

I think it may also be beneficial to have some footer text explaining briefly where the values for the input fields come from.

tcK1: https://github.com/Aidoku-Community/sources/pull/352/changes/6b00d6bfe0145f610806c9660ae36270f297c9e8

tcK1:
yea! I've been using [Rakuyomi](https://github.com/tachibana-shin/rakuyomi) to test both locally and on my kindle (I don't have a MacOS or iOS device to test on the Aidoku app 😢)

I can dm you somewhere the values I'm using if you want to test it!


https://github.com/user-attachments/assets/797b26f8-fa03-423a-9eeb-0daf6e69407b

sources/multi.mangaplus/res/settings.json
```diff
@@ -17,6 +17,53 @@
 				"title": "Split Double Pages",
 				"default": false
 			}
+
```
kkantan:
```suggestion
```

sources/multi.mangaplus/res/settings.json
```diff
@@ -17,6 +17,53 @@
...
+            {
+                "type": "text",
+                "key": "secret",
+                "title": "secret",
+                "placeholder": "secret (hash value)",
+                "requires": "mobile",
+                "refreshes": ["listings", "content"]
+            }
+        ],
+        "footer": "These values are obtained from the Mobile App, for example, by using a network sniffer. They are necessary to access the mobile API endpoints."
```
kkantan:
```suggestion
        "footer": "These values can be obtained from the MANGA Plus mobile app, for example, by using a network sniffer. They are necessary to access the mobile API endpoints."
```

sources/multi.mangaplus/res/settings.json
kkantan, line 1: sorry I didn't catch this before, but the JSON indentation should be using tabs instead of spaces.

kkantan (changes requested):
alright as long as the changes are working for you I think it's good, just a couple minor tweaks to the settings file and it should be good to merge.

sources/multi.mangaplus/res/settings.json
tcK1: ups, my bad! I thought I was using tabs 🤦🏼‍♂️

kkantan (approved):

PR #381, opened 2026-02-25
feat: add zh.dm5

### LINKED ISSUE #154: dm5.com
https://www.dm5.com

Fetching the list of images and loading the images requires `eval` processing, which I am not familiar with. 
Wait for others to achieve it.

dyphire:
Close #154


sources/zh.dm5/res/icon.png
kkantan, line 1: it would be nice to add a little padding to the edges of this icon.

sources/zh.dm5/res/source.json
kkantan, line 1: some spaces are being used for indentation that should be converted to tabs.

sources/zh.dm5/src/html/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+			let cover = banner
+				.select("img")
+				.and_then(|e| e.first())
+				.and_then(|e| e.attr("abs:src"))
+				.unwrap_or_default();
+			let url = banner
+				.select("p.title > a")
+				.and_then(|e| e.first())
+				.and_then(|e| e.attr("href"))
+				.unwrap_or_default();
```
kkantan: instead of unwrapping the `url` and `id` to empty strings, you could combine this into one variable and continue the loop if it's `None`. for example, something like:
```rust
let Ok(id) = banner
	.select("p.title > a")
	.and_then(|e| e.first())
	.and_then(|e| e.attr("href"))
    .and_then(|s| s.split('/').rfind(|s| !s.is_empty()).map(Into::into))
else {
    continue;
};
```

sources/zh.dm5/src/html/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+			let title = banner
+				.select("p.title > a")
+				.and_then(|e| e.first())
+				.and_then(|e| e.text())
+				.unwrap_or_default();
+			let cover = banner
+				.select("img")
+				.and_then(|e| e.first())
+				.and_then(|e| e.attr("abs:src"))
+				.unwrap_or_default();
```
kkantan: the cover doesn't need to be unwrapped, it can be left as an optional to be passed to the `Manga` struct.

sources/zh.dm5/src/html/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+					.select("p.mh-cover")
+					.and_then(|e| e.first())
+					.and_then(|e| e.attr("style"))
+					.unwrap_or_default();
+				let cover = cover_style
+					.split("url(")
+					.nth(1)
+					.and_then(|s| s.split(')').next())
+					.map(|s| s.to_string())
+					.unwrap_or_default();
```
kkantan: same comments as above.

sources/zh.dm5/src/html/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+		let mut entries: Vec<Manga> = Vec::new();
+
+		// Handle banner_detail_form for search
+		if let Some(banner) = self
+			.select("div.banner_detail_form")
+			.and_then(|e| e.first())
+		{
+			let title = banner
+				.select("p.title > a")
+				.and_then(|e| e.first())
```
kkantan: use `select_first` instead, everywhere.

sources/zh.dm5/src/html/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+impl ChapterPage for Document {
+	fn chapters(&self) -> Result<Vec<Chapter>> {
+		let container = self
+			.select("div#chapterlistload")
+			.and_then(|e| e.first())
+			.ok_or_else(|| error!("Chapter list not found"))?;
+
+		let category_buttons = self
+			.select("div.detail-list-title a.block")
+			.map(|els| els.into_iter().collect::<Vec<Element>>())
```
kkantan: is this necessary? I may be wrong, but I thought you could fetch `Element`s from an index in `ElementList`.

sources/zh.dm5/src/html/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+		}
+
+		chapters.reverse();
+		Ok(chapters)
+	}
+}
+
+pub fn fetch_chapters(manga_id: &str) -> Result<Vec<Chapter>> {
+	let url = format!("{}/{}", BASE_URL, manga_id);
+	let html = Request::new(url, HttpMethod::Get)?
```
kkantan: use `Request::get` instead.

sources/zh.dm5/src/html/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+	let mut pages: Vec<Page> = Vec::new();
+
+	// Check for direct images first
+	if let Some(img_list) = html.select("div#barChapter > img.load-src")
+		&& !img_list.is_empty()
+	{
+		for img in img_list {
+			let data_src = img.attr("data-src").unwrap_or_default();
+			pages.push(Page {
+				content: PageContent::Url(data_src, None),
```
kkantan:
```suggestion
				content: PageContent::url(data_src),
```

sources/zh.dm5/src/html/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+		.html()?;
+
+	let mut pages: Vec<Page> = Vec::new();
+
+	// Check for direct images first
+	if let Some(img_list) = html.select("div#barChapter > img.load-src")
+		&& !img_list.is_empty()
+	{
+		for img in img_list {
+			let data_src = img.attr("data-src").unwrap_or_default();
```
kkantan: instead of unwrapping it's better to skip adding the page to the vector.

sources/zh.dm5/src/html/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	// Find the script containing DM5 variables
+	let script = html
+		.select("script")
+		.and_then(|scripts| {
+			scripts
+				.into_iter()
+				.find(|s| s.data().is_some_and(|d| d.contains("DM5_MID")))
+				.and_then(|s| s.data())
```
kkantan: you can use `find_map` instead to not call `data()` twice.

sources/zh.dm5/src/html/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+		.ok_or_else(|| error!("Script not found"))?;
+
+	if !script.contains("DM5_VIEWSIGN_DT") {
+		if let Some(msg) = html
+			.select_first("div.view-pay-form p.subtitle")
+			.and_then(|e| e.text())
+		{
+			return Err(error!("{}", msg));
+		}
+		return Err(error!("Chapter not available"));
```
kkantan: you can use `bail!` instead.

sources/zh.dm5/src/net/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+					}
+					_ => {}
+				},
+				_ => {}
+			}
+		}
+
+		let mut parts: Vec<&str> = Vec::new();
+		for slug in &[area, genre, audience, status, pay, sort] {
+			if !slug.is_empty() {
```
kkantan: I think it's cleaner, or more conventional, to use `Option<&str>` instead of checking the string lengths.

sources/zh.dm5/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		if needs_details {
+			let html = Url::manga(manga.key.clone()).request()?.html()?;
+			html.update_details(&mut manga)?;
+		}
+
+		if needs_chapters {
+			manga.chapters = Some(html::fetch_chapters(&manga.key)?);
```
kkantan: aren't you requesting the same page here twice? ideally, you should only request the html data once, then use it for both the details and chapters.

sources/zh.dm5/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+		_context: Option<aidoku::PageContext>,
+	) -> Result<Request> {
+		let cid = url
+			.split("cid=")
+			.nth(1)
+			.and_then(|s| s.split('&').next())
+			.unwrap_or("");
+
+		let referer = if !cid.is_empty() {
+			format!("{}/m{}", BASE_URL, cid)
```
kkantan: use optionals instead of empty strings.
```suggestion
			.and_then(|s| s.split('&').next());

		let referer = if let Some(cid) = cid {
			format!("{}/m{}", BASE_URL, cid)
```

kkantan (changes requested):

sources/zh.dm5/src/html/mod.rs
dyphire: Done

sources/zh.dm5/src/html/mod.rs
dyphire: Done

sources/zh.dm5/src/html/mod.rs
dyphire: Done

sources/zh.dm5/src/net/mod.rs
dyphire: Done

sources/zh.dm5/res/source.json
```diff
@@ -0,0 +1,18 @@
+{
+	"info": {
+		"id": "zh.dm5",
+		"name": "动漫屋",
+		"version": 1,
+		"urls": [
+		"https://www.dm5.com",
+		"https://www.dm5.cn"
+	],
```
kkantan: the indentation here is a little off.
```suggestion
		"urls": [
			"https://www.dm5.com",
			"https://www.dm5.cn"
		],
```

sources/zh.dm5/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+		html.manga_page_result()
+	}
+
+	fn get_manga_update(
+		&self,
+		mut manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		if needs_details || needs_chapters {
```
kkantan: one of these will always be true when this function is called, so this check isn't necessary.

kkantan (changes requested):
just these two small changes, and the rest looks good.

sources/zh.dm5/src/lib.rs
dyphire: Done

kkantan (approved):

PR #383, opened 2026-02-26
feat: add vi.minotruyen

### LINKED ISSUE #382: minotruyenv5.xyz
https://minotruyenv5.xyz/

tachibana-shin:
Close #382 
@kkantan Don't bother with the unwrap() function; I'm sure it's not None, or it will fail before reaching it, so there's no need to complicate things.

### DIFF
--- a/sources/vi.minotruyen/src/lib.rs
+++ b/sources/vi.minotruyen/src/lib.rs
@@ -1,11 +1,10 @@
 #![no_std]
-use core::ops::Deref;
+use core::{cell::RefCell, ops::Deref};
 
 use aidoku::{
-	AidokuError, Chapter, FilterValue, Home, HomeComponent, HomeLayout, HomePartialResult,
-	ImageResponse, Manga, MangaPageResult, MangaWithChapter, Page, PageContext, PageImageProcessor,
-	Result, Source,
-	alloc::{String, Vec, borrow::ToOwned, string::ToString, sync::Arc, vec},
+	Chapter, FilterValue, Home, HomeComponent, HomeLayout, HomePartialResult, ImageResponse, Manga,
+	MangaPageResult, MangaWithChapter, Page, PageContext, PageImageProcessor, Result, Source,
+	alloc::{String, Vec, borrow::ToOwned, string::ToString, vec},
 	helpers::uri::QueryParameters,
 	imports::{canvas::ImageRef, defaults::defaults_get, net::Request, std::send_partial_result},
 	prelude::*,
@@ -18,7 +17,6 @@ mod models;
 mod utils;
 
 use models::*;
-use spin::Mutex;
 
 use crate::{
 	crypto::decrypt_cryptojs_passphrase,
@@ -31,7 +29,7 @@ pub const BASE_URL: &str = "https://minotruyenv5.xyz";
 pub const API_URL: &str = "https://api.cloudkk.art";
 
 struct MinoTruyen {
-	drm_tool: Arc<Mutex<Option<DrmToolWasm>>>,
+	drm_tool: RefCell<Option<DrmToolWasm>>,
 }
 
 impl Home for MinoTruyen {
@@ -67,8 +65,7 @@ impl Home for MinoTruyen {
 					API_URL,
 					if is_manga { "manga" } else { "comics" }
 				))?
-				.send()?
-				.get_json::<FeaturedRoot>()?
+				.json_owned::<FeaturedRoot>()?
 				.data
 				.books
 				.into_iter()
@@ -87,8 +84,7 @@ impl Home for MinoTruyen {
 					API_URL,
 					if is_manga { "manga" } else { "comics" }
 				))?
-				.send()?
-				.get_json::<SideHomeRoot>()?
+				.json_owned::<SideHomeRoot>()?
 				.top_books_view
 				.into_iter()
 				.map(|v| Manga::from(v).into())
@@ -149,7 +145,7 @@ impl Home for MinoTruyen {
 impl Source for MinoTruyen {
 	fn new() -> Self {
 		Self {
-			drm_tool: Arc::new(Mutex::new(None)),
+			drm_tool: RefCell::new(None),
 		}
 	}
 
@@ -188,8 +184,7 @@ impl Source for MinoTruyen {
 		}
 
 		let (entries, has_next_page) = Request::get(format!("{API_URL}/api/books?{qs}"))?
-			.send()?
-			.get_json::<FeaturedBooksData>()
+			.json_owned::<FeaturedBooksData>()
 			.map(|res| {
 				(
 					res.books.into_iter().map(Manga::from).collect(),
@@ -211,8 +206,7 @@ impl Source for MinoTruyen {
 	) -> Result<Manga> {
 		if needs_details {
 			let html = Request::get(format!("{BASE_URL}/{}", manga.key.replace("/", "/books/")))?
-				.send()?
-				.get_html()?;
+				.html()?;
 
 			let text = html
 				.select("script")
@@ -238,8 +232,7 @@ impl Source for MinoTruyen {
 				"{API_URL}/api/chapters/{}?order=desc&take=5000",
 				manga.key.rsplit('-').next().unwrap_or_default()
 			))?
-			.send()?
-			.get_json::<Chapters>()?
+			.json_owned::<Chapters>()?
 			.chapters
 			.into_iter()
 			.map(|c| VChapterF::to(c, &manga))
@@ -257,8 +250,7 @@ impl Source for MinoTruyen {
 			manga.key.replace("/", "/books/"),
 			chapter.key
 		))?
-		.send()?
-		.get_html()?;
+		.html()?;
 
 		let Some(data) = html.select("script").and_then(|f| {
 			for n in f {
@@ -278,9 +270,7 @@ impl Source for MinoTruyen {
 		)?;
 
 		let server = defaults_get::<String>("server").unwrap_or("0".to_owned());
-		let first_server = servers
-			.first()
-			.ok_or(AidokuError::Message("server not found".to_owned()))?;
+		let first_server = servers.first().ok_or(error!("server not found"))?;
 
 		let selected = servers
 			.iter()
@@ -316,13 +306,11 @@ impl PageImageProcessor for MinoTruyen {
 			return Ok(response.image);
 		}
 
-		let mut drm_tool = self.drm_tool.lock();
+		let mut drm_tool = self.drm_tool.borrow_mut();
 		if drm_tool.is_none() {
-			let mut tool = DrmToolWasm::new()
-				.map_err(|_| AidokuError::Message("drm tool not found".to_owned()))?;
+			let mut tool = DrmToolWasm::new().map_err(|_| error!("drm tool not found"))?;
 
-			tool.start()
-				.map_err(|_| AidokuError::Message("drm tool start error".to_owned()))?;
+			tool.start().map_err(|_| error!("drm tool start error"))?;
 
 			drm_tool.replace(tool);
 		}
@@ -331,7 +319,7 @@ impl PageImageProcessor for MinoTruyen {
 			.as_mut()
 			.unwrap()
 			.render_image(response, context.as_ref())
-			.map_err(|_| AidokuError::Message("drm tool render image error".to_owned()))?;
+			.map_err(|_| error!("drm tool render image error"))?;
 
 		// render_image(response, context)
 		Ok(image)
--- a/sources/vi.minotruyen/src/models.rs
+++ b/sources/vi.minotruyen/src/models.rs
@@ -1,6 +1,6 @@
 use crate::{
 	BASE_URL,
-	utils::{capitalize, get_viewer, parse_datetime_to_timestamp},
+	utils::{capitalize, get_viewer},
 };
 use aidoku::{
 	Chapter, Manga, MangaStatus, MangaWithChapter, Page, PageContent, PageContext,
@@ -128,9 +128,9 @@ impl From<BookItem> for Manga {
 			},
 			content_rating,
 			viewer,
-			chapters: value
-				.chapters
-				.map(|v| v.into_iter().map(|ref c| c.into()).collect::<Vec<_>>()),
+			// chapters: value
+			// 	.chapters
+			// 	.map(|v| v.into_iter().map(|ref c| c.into()).collect::<Vec<_>>()),
 			..Default::default()
 		}
 	}
@@ -164,7 +164,9 @@ pub struct Cover {
 #[derive(Debug, Deserialize, Default, Clone)]
 pub struct VChapter {
 	#[serde(rename = "createdAt")]
-	pub created_at: Option<String>,
+	pub created_at: Option<DateTime<Utc>>,
+	#[serde(rename = "updatedAt")]
+	pub updated_at: Option<DateTime<Utc>>,
 
 	#[serde(rename = "chapterNumber")]
 	pub chapter_number: f32,
@@ -176,10 +178,7 @@ impl From<&VChapter> for Chapter {
 			key: format!("chapter-{}-{}", value.chapter_number, value.chapter_number),
 			chapter_number: Some(value.chapter_number),
 			title: value.num.as_ref().map(|v| format!("Chap {v}")),
-			date_uploaded: value
-				.created_at
-				.as_ref()
-				.and_then(|v| parse_datetime_to_timestamp(v)),
+			date_uploaded: value.updated_at.or(value.created_at).map(|v| v.timestamp()),
 			..Default::default()
 		}
 	}
--- a/sources/vi.minotruyen/src/utils.rs
+++ b/sources/vi.minotruyen/src/utils.rs
@@ -1,9 +1,7 @@
 use aidoku::{
 	ContentRating, Viewer,
 	alloc::string::{String, ToString},
-	imports::std::get_utc_offset,
 };
-use chrono::{FixedOffset, NaiveDateTime, TimeZone};
 
 pub fn extract_data_chapter_block(script: &str) -> Option<String> {
 	// Regex matches: <hex32>:<value>
@@ -48,15 +46,6 @@ pub fn get_viewer(categories: &[String], category: &str) -> (ContentRating, View
 	}
 
 	(nsfw, viewer)
-}
-
-pub fn parse_datetime_to_timestamp(s: &str) -> Option<i64> {
-	// Format "YYYY-MM-DD HH:MM:SS"
-	let naive = NaiveDateTime::parse_from_str(s, "%Y-%m-%d %H:%M:%S").ok()?;
-	let offset = FixedOffset::east_opt(get_utc_offset() as i32)?;
-
-	let dt = offset.from_local_datetime(&naive).single()?;
-	Some(dt.timestamp())
 }
 
 pub fn extract_next_object(input: &str, skip: Option<usize>) -> Option<String> {


sources/vi.minotruyen/res/filters.json
kkantan, line 1: JSON files should use tabs for indentation.

sources/vi.minotruyen/res/icon.png
kkantan, line 1: the corner radius should be remove from the icon ideally, as the app defines its own corner radius. you can overlay the blue background color on top of the white parts.

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+use crate::{
+	crypto::decrypt_cryptojs_passphrase, drm_tool::DrmToolWasm, env::SECRET_DATA_CHAPTER,
+	utils::extract_data_chapter_block,
+};
+
+pub const BASE_URL: &str = "https://minotruyenv5.xyz";
+pub const API_URL: &str = "https://api.cloudkk.art";
+
+struct MinoTruyen {
+	drm_tool: Arc<Mutex<Option<DrmToolWasm>>>,
```
kkantan: `Arc` and `Mutex` shouldn't be necessary for non-concurrent code. you can use `Rc` instead of `Arc`, or `RefCell` for interior mutability like other sources.

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+}
+
+impl Home for MinoTruyen {
+	fn get_home(&self) -> Result<HomeLayout> {
+		let is_manga = defaults_get::<String>("type").unwrap_or_default() == "manga";
```
kkantan: try to avoid allocating empty strings that are unused. something like this should work instead.
```suggestion
		let is_manga = defaults_get::<String>("type").is_some_and(|s| s == "manga");
```

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+		}));
+
+		send_partial_result(&HomePartialResult::Component(HomeComponent {
+			title: Some(String::from("Truyện Nổi Bật")),
+			subtitle: None,
+			value: aidoku::HomeComponentValue::BigScroller {
+				entries: Request::get(format!(
+					"{}/api/books/top/featured?category={}&take=10",
+					API_URL,
+					if is_manga { "manga" } else { "comics" }
```
kkantan, line 66: one option could be just using the `type` default directly, since it's already either `manga` or `comics`.

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+				))?
+				.send()?
+				.get_json::<FeaturedRoot>()?
+				.data
+				.books
+				.into_iter()
+				.map(|v| v.into())
```
kkantan, line 72: you can use the trait function directly instead of creating a closure to call the function.
```suggestion
				.map(Into::into)
```

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+				page_size: Some(3),
+			},
+		}));
+
+		let html = Request::get(format!(
+			"{BASE_URL}/{}",
+			if is_manga { "manga" } else { "comics" }
+		))?
+		.send()?
+		.get_html()?;
```
kkantan:
```suggestion
		.html()?;
```

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+		let text = html
+			.select("script")
+			.and_then(|mut v| {
+				v.rfind(|node| {
+					node.html()
+						.unwrap_or_default()
```
kkantan: allocating an empty string unnecessarily. you can also use `is_some_and` here.

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+						.contains("self.__next_f.push([1,\"28:[[")
+				})
+			})
+			.and_then(|f| f.html())
+			.map(|t| {
+				let input = t.replace("\\\"", "\"").replace("\\\\\"", "\\\"");
+
+				input[43..(input.len() - 14)].to_string()
```
kkantan, line 115: where do these magic numbers come from?

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+		let mut qs = QueryParameters::new();
+
+		qs.push("take", "24".into());
+		qs.push("page", Some(&page.to_string()));
+
+		let r#type = defaults_get::<String>("type");
+		qs.push("category", r#type.as_deref());
+
+		if let Some(query) = query {
+			qs.push("q", Some(&query))
```
kkantan: could also use `as_deref` here like you do above.

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+		}
+
+		for filter in filters {
+			if let FilterValue::MultiSelect {
+				included, excluded, ..
+			} = filter
+			{
+				qs.push("genres", Some(&included.join(",").to_lowercase()));
+				qs.push("notgenres", Some(&excluded.join(",").to_lowercase()));
```
kkantan: these shouldn't have to be lowercased, since you control the values in `filters.json` already, right?

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+
+			let Some(json) = extract_next_object(&text, Some(3)) else {
+				bail!("extract Next error");
+			};
+
+			manga.copy_from(serde_json::from_str::<WrapBook>(&json)?.book.into());
+		}
+
+		if needs_chapters {
+			let chapters = Request::get(format!(
```
kkantan, line 231: since you're making another request here, it's nice to use `send_partial_result` with the manga after checking if `needs_chapters` is true inside and at the end of the previous if branch for details fetching.

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+				let html = n.html().unwrap_or_default();
+				if let Some(data) = extract_data_chapter_block(&html) {
+					return Some(data);
+				}
+			}
+			None
+		}) else {
+			return Err(AidokuError::Message(
+				"extract data chapter block error".to_owned(),
+			));
```
kkantan:
```suggestion
			bail!("extract data chapter block error");
```

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+		};
+
+		let servers = serde_json::from_str::<Vec<ChapterContent>>(
+			&decrypt_cryptojs_passphrase(&data, SECRET_DATA_CHAPTER)
+				.map_err(|_| AidokuError::Message("decrypt error".to_owned()))?,
```
kkantan:
```suggestion
				.map_err(|_| error!("decrypt error"))?,
```

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+};
+
+pub const BASE_URL: &str = "https://minotruyenv5.xyz";
+pub const API_URL: &str = "https://api.cloudkk.art";
+
+struct MinoTruyen {
+	drm_tool: Arc<Mutex<Option<DrmToolWasm>>>,
+}
+
+impl Home for MinoTruyen {
```
kkantan, line 35: conventionally, I like to put the `Home` trait conformance below the `Source` implementation, but it doesn't really matter.

sources/vi.minotruyen/src/models.rs
```diff
@@ -0,0 +1,319 @@
...
+	pub book: BookItem,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct BookItem {
+	#[serde(rename = "type")]
+	pub r#type: Option<String>,
+	pub slug: String,
+	pub title: String,
+	#[serde(rename = "bookId")]
```
kkantan, line 81: I think you can just rename_all with camelCase, right?

sources/vi.minotruyen/src/models.rs
```diff
@@ -0,0 +1,319 @@
...
+			..Default::default()
+		}
+	}
+}
+impl From<BookItem> for MangaWithChapter {
+	fn from(value: BookItem) -> Self {
+		let chapter = value
+			.chapters
+			.as_ref()
+			.and_then(|v| v.first().cloned())
```
kkantan: this shouldn't need to be cloned since you're consuming `value`.

sources/vi.minotruyen/src/models.rs
```diff
@@ -0,0 +1,319 @@
...
+	Str(String),
+	Arr(Box<FlightChild<FlightRoot<FlightChild<FlightNode>>>>),
+}
+
+#[derive(Debug, Deserialize)]
+pub struct FlightNode {
+	pub book: BookItem,
+}
+
+pub fn extract_next_object(input: &str, skip: Option<usize>) -> Option<String> {
```
kkantan: it seems like a function like this should be in `utils.rs` instead?

sources/vi.minotruyen/src/crypto.rs
```diff
@@ -0,0 +1,107 @@
...
...CKvp6sc0c2TgLrbwVnvUxMeekoecjKQPtcLIFMJZ+PSxo9JzKchbtqSCvrUUIq9yZ8iHJ7aHT0CWI6V+1qj419ZyfRnn57eY1V8tCCRUPhMIlmD/gp4lGBAyiXJBFnsy6ENwbphtMOtIVyDriXcqBuuFTKBTQP3EGX8rDl52gf8vIzBnQIPC0MYXq5QiLaDU34rE1CNxoA1eB81Vu0xbQ0pWWndkIqeFUg5pbG/n02PAp1aaMUfP1Pgx58LSoDDKxzddKWwbrZakiKSmgJo9fvB1/r7XIyj7r1K1hpCEkPLImWJZT16AfPfT4fusqeX4xGkfGTJVS5jyn66ILaDUK0UlFav3Smk/Aswxw+yOvhV6aBDS78RQ8ZIOOnR4VeloJ9/X+oZCfu8IUGPWnz7dSXJ+w9zBk9CxpHRcbBASljSfKWc+My6hqeV8jVmWk6vbdn4MgcWk7BUw7v8myHoGLePWePbkY8E2meZXJuvxKpfeyOHKNiCQBRfMyn+mNKqZkOKt+1ihdu3SrFPnH1PVG4rXmWdUUoWw1Evo/wVER8Orjkli+qI2yC29QOR64FAdKWDWOFTMcNIGFPP+l3qkQWrNVL/seFZllwRbUOuMadxQtLxuwj7PizUGTHJaCMJtaPYbKW9FYx+n+akM3ANKA/Ws3u2VQFTCwoQ2gdMsEYqYIWIAR/Pz05qI+tjUQ3afHd6ij3/PP6lhiel/z2pntyCfNrHfsULXWMZ4UxQ0m/CGi6a+Cwhv+r++Rim9kraOAf4JJVVjiLp6SV3X4KK7RpmPndNFt+/Lgij8+DbLGzeclTBn0SmqYsB5biS17bUxn9z+i/aY5EnJ4KRq5kPMBgFOZMuckBUDgySoRroCEdLU7dYGjW2ixQUboQOY/XLGt37E3c8afQ/rsjIWM4PVIjBQEQtVzKLlI9sjdZELZTHdEXDefeRw5QqicFaUP/yAiog7oSAX+vfQee2/HDmlxcOtV6/TsA9ko78H7la8sxR6tfUxagKgyf85dwV9JQGTXtg0sPg9CTN6vMbuyxVJDUaZMoG9BwT7dTBD4bihK2pDVsQFGgJlBqUz9XHzHoI+myE3nff79z3+W+BTSKVT3V/+sRrE8gg3YEh2Su654IDPvEA2Xf7dhW84ro6gL+0UULfPvzQhXmS9F2jFv9vfeZhG867sbxiJyo+NeSyalhbw==";
```
kkantan: it would be nice to convert this into a test instead.

sources/vi.minotruyen/src/crypto.rs
kkantan, line 1: it looks like you're doing some manual work here that could be avoided... you should see if the `crypto.rs` file in the madara template does what you need. it should be the same thing.

sources/vi.minotruyen/settings.json
kkantan, line 1: I don't think this file should exist?

sources/vi.minotruyen/src/drm_tool.rs
```diff
@@ -0,0 +1,724 @@
...
+#[derive(Debug)]
+struct ImageElement {
+	pub width: usize,
+	pub image: ImageRef,
+}
+
+#[derive(Debug)]
+struct CanvasRenderingContext2D {
+	// Option hack take box
+	pub canvas: Arc<Mutex<Option<Canvas>>>,
```
kkantan, line 35: as mentioned before, an Arc or Mutex shouldn't be used for non-concurrent code.

sources/vi.minotruyen/src/drm_tool.rs
```diff
@@ -0,0 +1,724 @@
...
+// }
+
+macro_rules! log {
+    ($($t:tt)*) => {
+        {
+        	#[cfg(debug_assertions)]
+            println!($($t)*);
+        }
+    };
+}
```
kkantan: this macro is already built-in to aidoku-rs, called `debug!`.

sources/vi.minotruyen/src/drm_tool.rs
```diff
@@ -0,0 +1,724 @@
...
+
+fn set_table<T>(mut caller: Caller<'_, WasmStore>, value: T) -> Result<i32>
+where
+	T: 'static + core::any::Any + Send + Sync,
+{
+	let index = {
+		let store = caller.data();
+		store.table_alloc_fn.unwrap().call(&mut caller, ())? as u64
```
kkantan, line 87: if you're confident that this is not `None`, and you're unwrapping it everywhere that you use it, it's not necessary for it to be an optional in the first place.

sources/vi.minotruyen/src/drm_tool.rs
kkantan, line 1: this is... interesting. you're shipping an entire webassembly interpreter inside of an interpreted webassembly program? i'm surprised it even works. instead of doing all this, isn't it possible to take advantage of the javascript execution apis? I guess that probably isn't supported by rakuyomi though.

kkantan (changes requested):

sources/vi.minotruyen/src/lib.rs
tachibana-shin: You wouldn't want to see Refcell here and cause a bunch of errors, I'm sure of that because it wasted an hour of my time.

sources/vi.minotruyen/settings.json
tachibana-shin: Why?

sources/vi.minotruyen/src/drm_tool.rs
tachibana-shin: Not using refcells will cause a lot of problems here. Yes, a lot of problems, and it's better to use your beloved mutexes.

sources/vi.minotruyen/src/drm_tool.rs
tachibana-shin: No, the wasm store must be initialized first, then it's assigned to the instance, and then wasm will share the table with this function and call the functions that are used, so it's definitely not None when called.

sources/vi.minotruyen/src/drm_tool.rs
tachibana-shin: Rakuyomi, with the support of the boa engine, can do some things, but it won't run such complex code; Wasmi is the best choice.

sources/vi.minotruyen/src/lib.rs
tachibana-shin: I initially intended to do it that way, but Rakuyomi can directly modify the installation file to some string, so I didn't want to have to search for it someday.

sources/vi.minotruyen/src/lib.rs
tachibana-shin: That's a reverse engineer for next-js; you'll need to study them in depth to figure it out.

sources/vi.minotruyen/src/models.rs
tachibana-shin: Actually, the output properties are quite mixed up, containing both Camel and Pascal, so I'll put each property separately.

sources/vi.minotruyen/src/crypto.rs
tachibana-shin: No code there doesn't meet my expectations, do you see it in fn main? Essentially, a hex table is passed in; it can predict changes in structure, so my code is ready for that.

sources/vi.minotruyen/res/icon.png
tachibana-shin: Please help me

sources/vi.minotruyen/src/lib.rs
tachibana-shin: I think you shouldn't worry about it. I want to make this more explicit and nest it in a function to recreate the callback hell. A 16-byte stack is worth it.

sources/vi.minotruyen/settings.json
tachibana-shin: a i see

sources/vi.minotruyen/src/lib.rs
kkantan: replace `Arc<Mutex<Option<DrmToolWasm>>>` with `RefCell<Option<DrmToolWasm>>`  and the `lock()` calls with `borrow_mut()` instead and it will work identically.

sources/vi.minotruyen/src/crypto.rs
kkantan: I tried decrypting the same encoded value you have in the test with the same secret value and the result was identical using madara's crypto functions. I recommend copying the `decrypt_key_iv` and `evp_bytes_to_key` functions, since they are simpler. you can still have the `decrypt_cryptojs_passphrase` function that does the base64 decoding and string parsing from the resulting utf8 data.

sources/vi.minotruyen/src/drm_tool.rs
kkantan: after checking, it looks like the wasmi library functions explicitly require `Send + Sync` conformance, so concurrent structures are necessary.

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	fn get_page_list(&self, manga: Manga, chapter: Chapter) -> Result<Vec<Page>> {
+		let html = Request::get(format!(
+			"{BASE_URL}/{}/{}",
+			manga.key.replace("/", "/books/"),
+			chapter.key
+		))?
+		.send()?
+		.get_html()?;
```
kkantan: you're using `send()` very frequently when it's almost never necessary. you should use `html()` instead of `get_html()` and `json_owned()` instead of `get_json()`, since they achieve the same results. `send()` should only be used if you need to access the actual response object that's returned.

sources/vi.minotruyen/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+
+		let servers = serde_json::from_str::<Vec<ChapterContent>>(
+			&decrypt_cryptojs_passphrase(&data, SECRET_DATA_CHAPTER)
+				.map_err(|_| error!("decrypt error"))?,
+		)?;
+
+		let server = defaults_get::<String>("server").unwrap_or("0".to_owned());
+		let first_server = servers
+			.first()
+			.ok_or(AidokuError::Message("server not found".to_owned()))?;
```
kkantan: all `AidokuError::Message`s should use `error!` instead.

tachibana-shin:
anyway I still like my crypto.rs more than madara's. the first hex string describes an intended custom IV which the code from madara doesn't give me how to edit the iv

tachibana-shin:
The only reason I think is making the time like that is that the api returns the correct date today or aidoku might miscalculate the std . If parsing fails then surely the json parser will give an error. I am currently returning timestamp seconds for the date_uploaded field

kkantan:
check "mayonaka heart tune". in aidoku, all of the chapters are listed as having the date either today or yesterday, but looking at the site, those are incorrect. I would recommend using the built-in date parsing api instead of using chrono.

I also checked "tonikaku kawaii" and the page processor is never called with `drm_data`, and all pages load correctly.

tachibana-shin:
I remember checking with the most recent chapter, the last image had DRM

kkantan:
I don't think I can in good faith let this source through the review without more revision.

the main reason for this is the wasm drm.

firstly, the actually drm doesn't appear to be *too* complex, so rather than resorting to wasm execution, you could instead reverse the rest of the process. it's the same as most other sources that have page processors: there is some encryption key that is decoded into instructions to shuffle parts of the image. you have a plethora of input-output examples, and the compiled source of a program that does what you need.

secondly, shipping the wasm interpreter has adverse effects: building this source takes extra time because of both the large dependency, and also because of the `include_bytes!` call. this is particularly bad for the action runners that build the source list. if wasm execution is truly necessary, in the case of frequent updates by the website, I would rather wait for the aidoku app to provide a more efficient or robust wasm api before including it in a source hosted on this source list.

apart from that, I am a little confused by your insistence on using unconventional rust paradigms.
- unwrap has extremely little use-case outside of quick scripting. this is fine if you're writing small, personal programs, but there are numerous other ways of handling these code paths in production code, most of which I have already suggested to you.
- you allocate strings more frequently than necessary, which is bad for embedded programs. if you're checking `is_empty`, chances are you can probably check `is_none` instead.
- many of my review comments are discarded without sufficient justification.

I review PRs in my free time, but I don't have the time to refactor your code myself. since this aidoku source list is one of largest, I feel that there is an importance in maintaining a higher standard of contributed code. I am happy to review any further changes or provide more specific refactoring advice.

kkantan:
> I agree with you that native APIs would be best; however, we're not paid to do this, and translating wasm is more complicated than everything else.

understandable, yeah.

> Do you think it's time for the repository to have a review agent?

I'm not really sure the status of ai review bots, so I don't know if they're helpful or not. I'm not planning on paying for stuff like that though, which I assume is part of the business model, so I wouldn't want to get locked into something.

I should work on the github actions though, since I believe there are more things that can be automated, including your other pr that I haven't gotten around to looking at yet.

PR #384, opened 2026-02-27
feat: add zh.bakamh source

### LINKED ISSUE #330: bakamh.com
Zh

dyphire:
Close #330


sources/zh.bakamh/src/html/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+			.select(".post-content>div")
+			.map(|e| e.count())
+			.unwrap_or(0);
+		if len > 0 {
+			manga.description = self
+				.select(aidoku::alloc::format!(
+					".post-content>div:nth-child({})>div>p",
+					len
+				))
+				.and_then(|mut e| e.next())
```
kkantan: isn't this the same as using `select_first`?

sources/zh.bakamh/src/html/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+			.select(".author-content>a")
+			.map(|els| els.filter_map(|el| el.text()).collect());
+
+		let len = self
+			.select(".post-content>div")
+			.map(|e| e.count())
+			.unwrap_or(0);
+		if len > 0 {
+			manga.description = self
+				.select(aidoku::alloc::format!(
```
kkantan: since you have `aidoku::prelude::*` imported, you shouldn't need a full namespace for macros.
```suggestion
				.select(format!(
```

sources/zh.bakamh/src/net/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+			match filter {
+				FilterValue::Text { id, value } => {
+					if id == "author" && !value.is_empty() {
+						let slug = value.to_lowercase().replace(' ', "-");
+						return Ok(Self::Author { name: slug, page });
+					}
+				}
+				FilterValue::Select { id, value } => match id.as_str() {
+					"分类" => path = value.clone(),
+					"genre" => genre = value.to_lowercase().replace('/', "-").clone(),
```
kkantan: I don't think these two things need to be cloned, since they're consumed in the iterator.

kkantan (changes requested):

sources/zh.bakamh/src/net/mod.rs
dyphire: Done

kkantan (approved):

PR #387, opened 2026-02-28
feat: add zh.se8 source

dyphire:


sources/zh.se8/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+
+	fn get_manga_update(
+		&self,
+		mut manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		let detail_url = format!("{}/comic/{}", BASE_URL, manga.key);
+
+		let html = Request::get(detail_url.clone())?
```
kkantan: you can pass a reference instead of cloning.
```suggestion
		let html = Request::get(&detail_url)?
```

sources/zh.se8/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+			let key = url
+				.trim_end_matches('/')
+				.split('/')
+				.next_back()
+				.unwrap_or_default()
+				.to_string();
+			if !key.is_empty() {
+				return Ok(Some(DeepLinkResult::Chapter {
+					key,
+					manga_key: String::new(),
```
kkantan: if you don't provide the manga key then this won't work, so you can just exclude chapter deep links.

sources/zh.se8/src/html/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+		})
+	}
+}
+
+impl MangaDetailPage for Document {
+	fn manga_details(&self, url: String, key: String) -> Result<Manga> {
+		let cover = self
+			.select_first(".de-info__cover img")
+			.and_then(|img| img.attr("src"))
+			.unwrap_or_default();
```
kkantan: the cover shouldn't be unwrapped to an empty string, since it's expected to be an optional. the same goes for line 46 as well.

kkantan (changes requested):

sources/zh.se8/src/html/mod.rs
dyphire: Done

kkantan (approved):

PR #388, opened 2026-03-01
fix(zh.baozimanhua): chapters and pages not loading

dyphire:


sources/zh.baozimanhua/src/html/mod.rs
```diff
@@ -257,7 +272,14 @@ impl ChapterPage for Document {
 			});
 		}
 
-		chapters.reverse();
+		if full_list_title.is_some() {
+			chapters.reverse();
+		}
+
+		if chapters.is_empty() {
+			return Err(error!("No chapters found"));
```
kkantan: you can use `bail!` here.

kkantan (changes requested):

sources/zh.baozimanhua/src/html/mod.rs
dyphire: Done

kkantan (approved):

PR #390, opened 2026-03-02
feat: add multi.ehentai

dyphire:


sources/multi.ehentai/res/settings.json
```diff
@@ -0,0 +1,52 @@
+[
+	{
+		"type": "group",
+		"title": "Account",
+		"subtitle": "Enter your E-Hentai cookies to access ExHentai.",
```
kkantan: I think you mean for this to be footer text.
```suggestion
		"footer": "Enter your E-Hentai cookies to access ExHentai.",
```

sources/multi.ehentai/res/icon.png
kkantan, line 1: this looks a little off center. is this how the site has it too?

sources/multi.ehentai/res/source.json
```diff
@@ -0,0 +1,14 @@
+{
+	"info": {
+		"id": "multi.ehentai",
+		"name": "E-Hentai",
+		"version": 1,
+		"url": "https://e-hentai.org",
+		"contentRating": 2,
+		"languages": ["All", "ja", "en", "zh", "nl", "fr", "de", "hu", "it", "ko", "pl", "pt-BR", "ru", "es", "th", "vi"]
```
kkantan: if the language select type is multi and not single, you shouldn't need an "All" option.

sources/multi.ehentai/src/home.rs
```diff
@@ -0,0 +1,236 @@
...
+				.header("Cookie", &cookies)
+				.header("User-Agent", USER_AGENT))
+		};
+
+		// Build all requests
+		let top_yesterday_req = make_req("https://e-hentai.org/toplist.php?tl=15&p=0")?;
+		let top_month_req = make_req("https://e-hentai.org/toplist.php?tl=13&p=0")?;
+		let top_year_req = make_req("https://e-hentai.org/toplist.php?tl=12&p=0")?;
+		let popular_req = make_req(&format!("{base_url}/popular"))?;
+		let latest_req = if lang_param.is_empty() {
```
kkantan: `lang_param` could be an `Option<String>` instead, and use `None` instead of `String::new()` in the fallback case.

sources/multi.ehentai/src/home.rs
```diff
@@ -0,0 +1,236 @@
...
+			Ok(Request::get(url)?
+				.header("Cookie", &cookies)
+				.header("User-Agent", USER_AGENT))
+		};
+
+		// Build all requests
+		let top_yesterday_req = make_req("https://e-hentai.org/toplist.php?tl=15&p=0")?;
+		let top_month_req = make_req("https://e-hentai.org/toplist.php?tl=13&p=0")?;
+		let top_year_req = make_req("https://e-hentai.org/toplist.php?tl=12&p=0")?;
+		let popular_req = make_req(&format!("{base_url}/popular"))?;
```
kkantan, line 199: does the api support filtering by language here too or no?

sources/multi.ehentai/src/models.rs
```diff
@@ -0,0 +1,340 @@
...
+		let description = if desc_parts.is_empty() {
+			None
+		} else {
+			Some(desc_parts.join("  \n"))
+		};
+
+		let viewer = if gallery
+			.tags
+			.iter()
+			.any(|t| t.namespace == "language" && t.name == "japanese")
```
kkantan: it looks like the website also has "manga" and "doujinshi" categories. is it possible to access those to determine the proper viewer?

sources/multi.ehentai/src/parser.rs
```diff
@@ -0,0 +1,552 @@
...
+	let url = normalize_gallery_url(&url);
+
+	let glink = link_el.select_first(".glink");
+	let title = glink
+		.as_ref()
+		.and_then(|e| e.text())
+		.or_else(|| link_el.text())
+		.unwrap_or_default();
+
+	if title.is_empty() {
```
kkantan: should use an Option and `let Some(...) = ... else` instead.

sources/multi.ehentai/src/lib.rs
```diff
@@ -0,0 +1,855 @@
...
+				refresh_igneous_from_set_cookie(&set_cookie);
+			}
+
+			let retry_doc = resp.get_html()?;
+			let still_rejected = retry_doc
+				.select_first("body")
+				.and_then(|b| b.select_first("div"))
+				.is_none();
+			if still_rejected {
+				return Err(AidokuError::Message(
```
kkantan: `bail!` can be used here.

sources/multi.ehentai/src/lib.rs
```diff
@@ -0,0 +1,855 @@
...
+mod settings;
+
+use parser::*;
+use settings::*;
+const USER_AGENT: &str = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 \
+                          (KHTML, like Gecko) Chrome/122.0.0.0 Safari/537.36";
+
+struct EHentai;
+
+fn eh_get_html(url: &str, cookies: &str, user_agent: &str) -> Result<Document> {
```
kkantan: it would probably be nicer to put these miscellaneous functions inside a `helpers.rs` module.

sources/multi.ehentai/src/lib.rs
```diff
@@ -0,0 +1,855 @@
...
+						qs.push("f_sh", Some("on"));
+					}
+				}
+				FilterValue::Text { id, value } if !value.is_empty() => {
+					match id.as_str() {
+						"tags" => tag_filter = Some(value),
+						"author" => {
+							// clicked from author field: search both artist and group (OR via ~)
+							query_str.push_str(&format!(
+								" ~artist:\"{}$\" ~group:\"{}$\"",
```
kkantan, line 132: this doesn't seem to work. clicking on an author shows the same results regardless of the author.

sources/multi.ehentai/src/lib.rs
```diff
@@ -0,0 +1,855 @@
...
+	}
+
+	fn get_page_list(&self, _manga: Manga, chapter: Chapter) -> Result<Vec<Page>> {
+		let gallery_url = rewrite_domain(&chapter.key);
+		let cookies = build_cookie_header();
+
+		let mut viewer_urls: Vec<String> = Vec::new();
+		let mut next_url: Option<String> = Some(gallery_url.clone());
+
+		while let Some(ref url) = next_url.clone() {
```
kkantan: I don't think either of these clones are necessary.

sources/multi.ehentai/src/lib.rs
```diff
@@ -0,0 +1,855 @@
...
+			viewer_urls.extend(
+				parse_gallery_pages(&html)
+					.into_iter()
+					.map(|u| rewrite_domain(&u)),
+			);
+			next_url = parse_next_gallery_page(&html).map(|u| rewrite_domain(&u));
+		}
+
+		if viewer_urls.is_empty() {
+			return Err(AidokuError::Message("No pages found".into()));
```
kkantan:
```suggestion
			bail!("No pages found");
```

sources/multi.ehentai/src/home.rs
dyphire: no

sources/multi.ehentai/src/models.rs
dyphire: No, e-hentai's manga category refers to H mangas published by third parties, and doujinshi refers to any self-published H mangas or pornographic Western comics. They cannot be used to distinguish viewers.

sources/multi.ehentai/src/lib.rs
dyphire: cannot to reproduce

sources/multi.ehentai/res/icon.png
dyphire: fix

sources/multi.ehentai/src/models.rs
dyphire: Try to select the viewer more reliably

sources/multi.ehentai/res/icon.png
kkantan: it looks like it's shifted a few pixels too high now...

sources/multi.ehentai/src/lib.rs
kkantan: it looks like it only fails to work if the user has multiple languages selected.

sources/multi.ehentai/src/lib.rs
```diff
@@ -0,0 +1,692 @@
...
+
+use helpers::*;
+use parser::*;
+use settings::*;
+const USER_AGENT: &str = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 \
+                          (KHTML, like Gecko) Chrome/122.0.0.0 Safari/537.36";
+
+struct EHentai;
+
+// helper functions moved to `helpers.rs`
```
kkantan: you can remove this comment.

sources/multi.ehentai/src/lib.rs
```diff
@@ -0,0 +1,692 @@
...
+
+		if cats_filtered && cat_mask != ALL_CATS {
+			for (flag_id, mask) in cat_flags {
+				qs.push(flag_id, Some(if cat_mask & mask != 0 { "1" } else { "0" }));
+			}
+		}
+
+		if min_rating > 0 {
+			qs.push("f_sr", Some("on"));
+			qs.push("f_srdd", Some(&(min_rating + 1).to_string()));
```
kkantan: I don't think this `+ 1` is necessary, since it appears as though the options in the source correspond with the option with one higher star on the website.

sources/multi.ehentai/res/filters.json
```diff
@@ -0,0 +1,71 @@
...
+		"ids": ["f_doujinshi", "f_manga", "f_artistcg", "f_gamecg", "f_western", "f_non-h", "f_imageset", "f_cosplay", "f_asianporn", "f_misc"],
+		"defaultIncluded": ["f_doujinshi", "f_manga", "f_artistcg", "f_gamecg", "f_western", "f_non-h", "f_imageset", "f_cosplay", "f_asianporn", "f_misc"]
+	},
+	{
+		"type": "select",
+		"id": "min_rating",
+		"title": "Minimum Rating",
+		"options": ["Any", "2 stars", "3 stars", "4 stars", "5 stars"],
+		"ids": ["0", "2", "3", "4", "5"],
+		"default": "Any"
```
kkantan: the default value needs to use one of the ids.
```suggestion
		"default": "0"
```

sources/multi.ehentai/src/lib.rs
dyphire: This is actually a problem with the OR search syntax. Since the search conditions are all OR, only the same language collection will be searched.

This not only affects the author search but also the search results when using pure OR search syntax in the search box. The correct way is to change the language selection to a single selection. implementation.

PR #393, opened 2026-03-03
feat: add multi.hitomi

### LINKED ISSUE #44: hitomi.la
NSFW

https://hitomi.la

dyphire:
Close #44

### DIFF
--- a/sources/multi.hitomi/src/lib.rs
+++ b/sources/multi.hitomi/src/lib.rs
@@ -444,10 +444,10 @@ impl Source for Hitomi {
 		{
 			let now = current_date();
 			let cached = self.gg_cache.borrow();
-			if cached.as_ref().is_none_or(|&(_, ts)| now - ts > 60) {
-				if let Some(gg) = get_new_gg() {
-					*self.gg_cache.borrow_mut() = Some((gg, now));
-				}
+			if cached.as_ref().is_none_or(|&(_, ts)| now - ts > 60)
+				&& let Some(gg) = get_new_gg()
+			{
+				*self.gg_cache.borrow_mut() = Some((gg, now));
 			}
 		}
 		let cache = self.gg_cache.borrow();
--- a/sources/multi.hitomi/src/models.rs
+++ b/sources/multi.hitomi/src/models.rs
@@ -198,7 +198,7 @@ impl From<HitomiGallery> for Manga {
 		if let Some(parodys) = &g.parodys
 			&& !parodys.is_empty()
 		{
-			let s: Vec<&str> = parodys.into_iter().map(|p| p.parody.as_str()).collect();
+			let s: Vec<&str> = parodys.iter().map(|p| p.parody.as_str()).collect();
 			description_parts.push(format!("Series: {}", s.join(", ")));
 		}
 		if let Some(characters) = &g.characters


sources/multi.hitomi/res/icon.png
kkantan, line 1: the white outline should be removed by filling with the gray background color, if possible.

sources/multi.hitomi/src/gg.rs
```diff
@@ -0,0 +1,152 @@
...
+	pub b: String,
+	pub switch_cases: Vec<u32>,
+	pub switch_o: u32,
+	pub default_o: u32,
+}
+
+pub fn fetch_gg_state() -> Option<GgState> {
+	// Simple short-term cache:
+	static mut CACHED: Option<GgState> = None;
+	static TS: AtomicI64 = AtomicI64::new(0);
```
kkantan: firstly: an atomic i64 isn't necessary since this code shouldn't be run concurrently.
secondly: it's best to avoid `static mut`. could we instead put a `RefCell` on the `Hitomi` source struct to pass in? see the contribution guide for more explanation.

sources/multi.hitomi/src/lib.rs
```diff
@@ -0,0 +1,564 @@
...
+							format!("tag:{v}")
+						});
+					}
+					_ => {}
+				},
+				_ => {}
+			}
+		}
+
+		let raw_q: String = query.as_deref().unwrap_or("").trim().into();
```
kkantan:
```suggestion
		let raw_q: String = query.as_deref().unwrap_or_default().trim().into();
```

sources/multi.hitomi/src/lib.rs
```diff
@@ -0,0 +1,564 @@
...
+		let mut positive_results: Vec<Vec<i64>> = Vec::new();
+		// Collect local (plain text) results immediately, and build nozomi requests for network ones
+		let mut local_results: Vec<(usize, Vec<i64>)> = Vec::new();
+		let mut nozomi_requests: Vec<(usize, String)> = Vec::new();
+		for (i, term) in positive_terms.iter().enumerate() {
+			if term.contains(':') {
+				if let Some(url) = nozomi_url_for_ns_tag(term, &lang) {
+					nozomi_requests.push((i, url));
+				} else {
+					return Err(error!("Unsupported namespace for term: {term}"));
```
kkantan: `bail!`

sources/multi.hitomi/src/lib.rs
```diff
@@ -0,0 +1,564 @@
...
+		&self,
+		mut manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		let id: i64 = manga
+			.key
+			.parse()
+			.map_err(|_| error!("Invalid gallery id"))?;
+		if needs_details || needs_chapters {
```
kkantan: this will always be true.

sources/multi.hitomi/src/gg.rs
dyphire: Done

sources/multi.hitomi/res/icon.png
dyphire: Fix

dyphire:
> it seems like regular searches with spaces in them fail.

Fix

PR #394, opened 2026-03-03
fix(zh.zaimanhua): sync filters & style improvements & behavior fixes

Aiko3993:


sources/zh.zaimanhua/src/lib.rs
```diff
@@ -133,13 +136,14 @@ impl Source for Zaimanhua {
...
 			(true, true) => {
-				// Clone for chapters, consume for details
 				let chapters_source = manga_detail.clone();
 				manga.copy_from(manga_detail.into_manga(manga.key.clone()));
+				aidoku::imports::std::send_partial_result(&manga);
 				manga.chapters = Some(chapters_source.into_chapters(&manga.key));
 			}
 			(true, false) => {
 				manga.copy_from(manga_detail.into_manga(manga.key.clone()));
+				aidoku::imports::std::send_partial_result(&manga);
```
kkantan: there's no need to send the partial manga result if the chapters fetching is relatively quick, and definitely not if there is no chapter fetching. you only really need to use this if chapter fetching requires an additional or separate network request.

sources/zh.zaimanhua/src/net.rs
```diff
@@ -170,7 +204,9 @@ pub fn send_authed_request<T: DeserializeOwned>(
 		&& let Ok(Some(new_token)) = try_refresh_token()
 	{
 		// Retry with new token
-		return auth_request(url, Some(&new_token))?.send()?.get_json_owned();
+		return auth_request(url, Some(&new_token))?
+			.send()?
+			.get_json_owned();
```
kkantan: you can probably just use `.json_owned()` without the explicit `send()` call.

kkantan (changes requested):

PR #432, opened 2026-03-22
feat: rewrite ru.desu with new api

MemesPwr:
Since legacy sources may no longer work in the future i decided to rewrite it with a new source API.

### DIFF
--- a/sources/ru.desu/src/lib.rs
+++ b/sources/ru.desu/src/lib.rs
@@ -32,7 +32,10 @@ impl Source for Desu {
 	) -> Result<MangaPageResult> {
 		search(query, page, filters).map(|r| MangaPageResult {
 			has_next_page: r.len() >= helpers::PAGE_SIZE,
-			entries: r.into_iter().map(|m| m.to_slim_item()).collect(),
+			entries: r
+				.into_iter()
+				.map(|m| m.into_manga(None, true, false, false))
+				.collect(),
 		})
 	}
 
@@ -42,7 +45,8 @@ impl Source for Desu {
 		needs_details: bool,
 		needs_chapters: bool,
 	) -> Result<Manga> {
-		fetch_by_id(manga.key.as_str()).map(|x| x.to_manga(needs_details, needs_chapters))
+		fetch_by_id(manga.key.as_str())
+			.map(|x| x.into_manga(Some(manga), false, needs_details, needs_chapters))
 	}
 
 	fn get_page_list(&self, manga: Manga, chapter: Chapter) -> Result<Vec<Page>> {
--- a/sources/ru.desu/src/models.rs
+++ b/sources/ru.desu/src/models.rs
@@ -131,121 +131,88 @@ impl From<DesuChapter> for Chapter {
 }
 
 impl DesuItem {
-	pub fn get_key(&self) -> String {
-		self.id.to_string()
-	}
+	pub fn into_manga(
+		self,
+		manga: Option<Manga>,
+		slim: bool,
+		details: bool,
+		chapters: bool,
+	) -> Manga {
+		let mut item = manga.unwrap_or(Manga {
+			key: self.id.to_string(),
+			..Default::default()
+		});
 
-	pub fn get_title(&self) -> String {
-		if eng_title() {
-			self.name.clone()
+		item.title = if eng_title() {
+			self.name
 		} else {
-			self.russian.clone().unwrap_or_else(|| self.name.clone())
+			self.russian.unwrap_or(self.name)
+		};
+
+		item.cover = self
+			.image
+			.and_then(|v| v.original.or(v.preview).or(v.x225).or(v.x120));
+
+		if slim {
+			return item;
 		}
-	}
 
-	pub fn get_cover(&self) -> Option<String> {
-		self.image.as_ref().and_then(|v| {
-			v.original
-				.clone()
-				.or_else(|| v.preview.clone())
-				.or_else(|| v.x225.clone())
-				.or_else(|| v.x120.clone())
-		})
-	}
-
-	pub fn get_status(&self) -> MangaStatus {
-		self.status
-			.as_ref()
-			.map(|v| match v.as_str() {
-				// looks like they don't have hiatus status and so on
-				"ongoing" => MangaStatus::Ongoing,
-				"released" => MangaStatus::Completed,
-				_ => MangaStatus::Unknown,
-			})
-			.unwrap_or_default()
-	}
-
-	pub fn get_rating(&self) -> ContentRating {
-		self.age_limit
-			.as_ref()
-			.map(|v| match v.as_str() {
-				// "no" if no age limit. I guess safe by default is ok...
-				"18_plus" => ContentRating::NSFW,
-				"16_plus" => ContentRating::Suggestive,
-				_ => ContentRating::Safe,
-			})
-			.unwrap_or_default()
-	}
-
-	pub fn get_viewer(&self) -> Viewer {
-		match self.kind.as_ref() {
-			// since they can set read_mode to RTL even for manhwa/manhua I decided to do this
-			"manhwa" | "manhua" => Viewer::Webtoon,
-			_ => self
-				.reading
-				.as_ref()
+		if details {
+			item.content_rating = self
+				.age_limit
 				.map(|v| match v.as_str() {
-					"right-to-left" => Viewer::RightToLeft,
-					"left-to-right" => Viewer::LeftToRight,
-					"top-to-bottom" => Viewer::Webtoon,
-					_ => Viewer::RightToLeft,
+					// "no" if no age limit. I guess safe by default is ok...
+					"18_plus" => ContentRating::NSFW,
+					"16_plus" => ContentRating::Suggestive,
+					_ => ContentRating::Safe,
 				})
-				.unwrap_or(Viewer::RightToLeft),
+				.unwrap_or_default();
+
+			item.status = self
+				.status
+				.map(|v| match v.as_str() {
+					// looks like they don't have hiatus status and so on
+					"ongoing" => MangaStatus::Ongoing,
+					"released" => MangaStatus::Completed,
+					_ => MangaStatus::Unknown,
+				})
+				.unwrap_or_default();
+
+			item.viewer = match self.kind.as_str() {
+				// since they can set read_mode to RTL even for manhwa/manhua I decided to do this
+				"manhwa" | "manhua" => Viewer::Webtoon,
+				_ => self
+					.reading
+					.as_ref()
+					.map(|v| match v.as_str() {
+						"right-to-left" => Viewer::RightToLeft,
+						"left-to-right" => Viewer::LeftToRight,
+						"top-to-bottom" => Viewer::Webtoon,
+						_ => Viewer::RightToLeft,
+					})
+					.unwrap_or(Viewer::RightToLeft),
+			};
+
+			item.authors = self
+				.authors
+				.map(|l| l.into_iter().map(|v| v.people_name).collect());
+
+			item.description = self.description;
+
+			item.url = self.url;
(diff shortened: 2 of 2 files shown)


sources/ru.desu/res/settings.json
```diff
@@ -0,0 +1,22 @@
+[
+	{
+		"type": "group",
+		"title": "Settings",
+		"items": [
+			{
+				"type": "switch",
+				"key": "eng_title",
+				"title": "Название на англ.",
```
kkantan: I think the general convention for settings is Title Case, not complete sentences. I'm not sure how this works exactly with russian though.
```suggestion
				"title": "Название на англ",
```

sources/ru.desu/res/settings.json
```diff
@@ -0,0 +1,22 @@
...
+				"default": false,
+				"refreshes": ["content"]
+			},
+			{
+				"type": "text",
+				"key": "domain",
+				"title": "Домен",
+				"default": "desu.uno",
```
kkantan, line 17: since this is a url, you can enable some settings to make the keyboard better for urls: https://github.com/Aidoku-Community/sources/blob/29f5c821e9f01b053711967ceba4d6b0a0da9442/sources/ru.mangalib/res/settings.json#L11-L13

sources/ru.desu/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "ru.desu",
+		"name": "Desu",
+		"version": 4,
+		"url": "https://desu.uno/",
```
kkantan:
```suggestion
		"url": "https://desu.uno",
```

sources/ru.desu/src/helpers.rs
```diff
@@ -0,0 +1,119 @@
...
+	request
+		.header("User-Agent", "Aidoku")
+		.header("Referer", get_base_url().as_str())
+}
+
+pub fn fetch_by_id(id: &str) -> Result<DesuItem> {
+	let url = format!("{}/{}", get_base_api_url(), id);
+	let response = apply_headers(Request::get(url)?)
+		.send()?
+		.get_json::<DesuResponse<DesuItem>>()?;
```
kkantan: you can do this with a single call.
```suggestion
		.json_owned::<DesuResponse<DesuItem>>()?;
```

sources/ru.desu/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+		let pages = fetch_chapter(manga.key.as_str(), chapter.key.as_str())
+			.and_then(|s| {
+				s.pages
+					.and_then(|x| x.list)
+					.ok_or(error!("Chapter {} not found", chapter.key))
+			})?
+			.into_iter()
+			.filter_map(|p| {
+				p.img.map(|u| Page {
+					content: PageContent::Url(u, None),
```
kkantan:
```suggestion
					content: PageContent::url(u),
```

sources/ru.desu/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl ListingProvider for Desu {
+	fn get_manga_list(&self, _listing: Listing, page: i32) -> Result<MangaPageResult> {
+		search(None, page, vec![]).map(|r| MangaPageResult {
+			has_next_page: r.len() >= helpers::PAGE_SIZE,
+			entries: r.into_iter().map(Manga::from).collect(),
+		})
+	}
+}
```
kkantan: this is never used, is it? you aren't providing listings or a home page.

sources/ru.desu/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl DeepLinkHandler for Desu {
+	fn handle_deep_link(&self, _url: String) -> Result<Option<DeepLinkResult>> {
+		let Some(path) = _url.strip_prefix(get_base_url().as_str()) else {
```
kkantan:
```suggestion
	fn handle_deep_link(&self, url: String) -> Result<Option<DeepLinkResult>> {
		let Some(path) = url.strip_prefix(get_base_url().as_str()) else {
```

sources/ru.desu/src/models.rs
```diff
@@ -0,0 +1,193 @@
...
+			description: value.description,
+			content_rating: value
+				.age_limit
+				.map(|v| match v.as_str() {
+					// "no" if no age limit. I guess safe by default is ok...
+					"18_plus" => ContentRating::NSFW,
+					"16_plus" => ContentRating::Suggestive,
+					_ => ContentRating::Safe,
+				})
+				.unwrap_or(ContentRating::Unknown),
```
kkantan: you could also just do `unwrap_or_default()`, but it's up to you whether you find that less intuitive or not.

sources/ru.desu/src/models.rs
```diff
@@ -0,0 +1,193 @@
...
+						"top-to-bottom" => Viewer::Webtoon,
+						_ => Viewer::RightToLeft,
+					})
+					.unwrap_or(Viewer::RightToLeft),
+			},
+			authors: value
+				.authors
+				.map(|l| l.into_iter().map(|v| v.people_name).collect()),
+			artists: value
+				.translators
```
kkantan: are translators really artists? is this just scanlators?

MemesPwr: there are some teams who redrawing some parts so I don't know what will be better. I guess... it can be removed.

upd: decided to remove it here 92e8f22e8411287108ab4d073158c6c04c6da6b9

sources/ru.desu/res/settings.json
MemesPwr: Fixed in d870075f6acbaf840c0c6c705d7b3cee8a3e123d

sources/ru.desu/src/lib.rs
MemesPwr: In this case I'm getting `this enum variant takes 2 arguments but 1 argument was supplied [E0061]` error so I was used it with `None` to pass building process. If this really necessary to remove this `None` then I need help here 💀

sources/ru.desu/src/helpers.rs
MemesPwr: Fixed in 0a4d1eefba504c54e1b3af529f6e3c6a322857ec

sources/ru.desu/res/settings.json
MemesPwr: Fixed in d3c7ce13c20617f958c2f626b314ff9021b52726

sources/ru.desu/src/lib.rs
MemesPwr: Oops I thought that this would be necessary and forgot to check is it really still needed like was for the old API. Fixed in 4489a7c48b1d38805decd3fdba324da3973561ae

sources/ru.desu/src/models.rs
MemesPwr: I guess it's ok. Done it here 76fc2e04b274af9eab9214614006fbeb4fe48ab8

sources/ru.desu/src/lib.rs
kkantan: did you use the lowercase `url` instead of `Url`? the `url` is a helper function rather than the direct enum variant.

sources/ru.desu/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+		})
+	}
+
+	fn get_manga_update(
+		&self,
+		manga: Manga,
+		_needs_details: bool,
+		_needs_chapters: bool,
+	) -> Result<Manga> {
+		fetch_by_id(manga.key.as_str()).map(Manga::from)
```
kkantan: the only thing I'm concerned about now is the disregarding of the `needs_details` and `needs_chapters` parameters. this seems to work fine in the app, but I don't know if there are places where it's used where it would be better if you didn't always provide both things.

you should probably make the `From<DesuItem>` for `Manga` not include chapters, and then you can set them on the manga object directly with a non-consuming `chapters()` helper function maybe. I think there are other sources with this. you can use the [copy_from](https://aidoku.github.io/aidoku-rs/aidoku/struct.Manga.html#method.copy_from) method for details, or have your own function that can optionally include or exclude the chapters.

sources/ru.desu/src/lib.rs
MemesPwr: oh really.. I didn't notice that you're suggesting lowercase 💀

sources/ru.desu/src/lib.rs
MemesPwr: Done in 1712f0b7af988fbdc4634fb95e0983422dc6a551 (here I fix arguments naming 0acfbbb96695160d9d0c25e15a271ed07078d4f2)

sources/ru.desu/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	fn get_manga_update(
+		&self,
+		manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		fetch_by_id(manga.key.as_str()).map(|x| x.to_manga(needs_details, needs_chapters))
+	}
```
kkantan, line 50: I think this could cause issues if the app only requests chapters and passes in a manga with existing details, which would then be returned without them (but again, I can't say for certain).

I would suggest handling it like this instead:
```suggestion
	fn get_manga_update(
		&self,
		mut manga: Manga,
		needs_details: bool,
		needs_chapters: bool,
	) -> Result<Manga> {
		let result = fetch_by_id(&manga.key)?;

		if needs_details {
			manga.copy_from(result.to_manga(needs_chapters));
		} else if needs_chapters {
			manga.chapters = result.to_chapters();
		}

		Ok(manga)
	}
```
in order to mirror how other sources function. to avoid clones, I would make `to_manga` and `to_chapters` both consume `self` rather than take a reference, but that means you need to duplicate the chapter "parsing" code.

sources/ru.desu/src/models.rs
```diff
@@ -0,0 +1,251 @@
...
+
+	pub fn get_cover(&self) -> Option<String> {
+		self.image.as_ref().and_then(|v| {
+			v.original
+				.clone()
+				.or_else(|| v.preview.clone())
+				.or_else(|| v.x225.clone())
+				.or_else(|| v.x120.clone())
+		})
+	}
```
kkantan: even though it's tempting to refactor into helper functions like this, it's better if you can avoid `clone()`s as much as possible.

sources/ru.desu/src/lib.rs
MemesPwr: > `manga.chapters = result.to_chapters();`

wow. didn't know that it's can be used like this.

but isn't will be better then to do it like this?:
```suggestion
	fn get_manga_update(
		&self,
		mut manga: Manga,
		needs_details: bool,
		needs_chapters: bool,
	) -> Result<Manga> {
		let result = fetch_by_id(&manga.key)?;
		if needs_details {
			manga.copy_from(result.to_manga(needs_chapters));
		}
		if needs_chapters {
			manga.chapters = result.to_chapters();
		}
		Ok(manga)
	}
```

or the app will always require only one of this so `if .. else ..` is would be enough here?

kkantan: that wouldn't work if the `to_manga` function consumes the `result`, so it's best to include the chapters in `to_manga` according to the `needs_chapters`.

MemesPwr: I was "killed" two hares with this commit d62f59b170131221a88026a26aa097ead820b1f7 I suppose 😄

sources/ru.desu/src/models.rs
MemesPwr: This one is here too d62f59b170131221a88026a26aa097ead820b1f7

PR #437, opened 2026-03-26
fix(vi.cuutruyen): update proxy url

tachibana-shin:


kkantan (changes requested):
the source version should be incremented too so users can update.

PR #438, opened 2026-03-27
fix(vi.truyenqq): search results not showing

tachibana-shin:


tachibana-shin:
@kkantan Please go to sleep; you're playing with your health

PR #441, opened 2026-03-29
feat: add multi.bakamh

z3niith:
Added bakamh.com as a source (v1).


sources/multi.bakamh/res/source.json
```diff
@@ -0,0 +1,14 @@
+{
+	"info": {
+		"id": "multi.bakamh",
+		"name": "bakamh",
+		"version": 1,
+		"url": "https://bakamh.com/",
```
kkantan:
```suggestion
		"url": "https://bakamh.com",
```

sources/multi.bakamh/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+crate-type = ["cdylib"]
+
+[profile.dev]
+panic = "abort"
+
+[profile.release]
+panic = "abort"
+opt-level = "s"
+strip = true
+lto = true
```
kkantan:
```suggestion
lto = true

```

sources/multi.bakamh/src/lib.rs
kkantan, line 1: it looks like [bakamh uses the madara template](https://github.com/keiyoushi/extensions-source/blob/f18f82294086020a80d6526eba816f3404f95d10/src/zh/bakamh/build.gradle#L4), so instead of implementing all the logic from scratch, you should be able to use the existing template for it.

sources/multi.bakamh/res/source.json
```diff
@@ -0,0 +1,14 @@
...
+		"contentRating": 2,
+		"languages": [
+			"en",
+			"zh",
+			"Ko"
+		]
```
kkantan, line 12: does the source allow filtering by language? and is there actually content in these other languages, or is this just manga/manhwa/manhua?

sources/multi.bakamh/res/icon.png
kkantan, line 1: source icons should be 128x128. also, is there a way to scale down the logo so that there's some padding on the edges?

sources/multi.bakamh/res/source.json
z3niith: Bakamh does offer content in other languages, though limited for English content. Here's the [English](https://bakamh.com/en-manga/), [Korean](https://bakamh.com/manhwa/), and [Chinese](https://bakamh.com/on-going/zh/).

sources/multi.bakamh/Cargo.toml
z3niith: Changes have been made. Please verify the new commit if available.

z3niith:
That was on Aidoku version 0.8.2, build 4-which is the latest version if I am correct.

I was referring to loading the whole site's content in general, not just next page content. I am currently testing that source (zh.bakamh) and I've let it load for ~10 minutes now but nothing's populating.

image

PR #443, opened 2026-03-30
fix(multi.nhentai): update to API v2

### LINKED ISSUE #442: nhentai displaying "No Result, Retry" - Source not loading
As image shows. Looks like no matter which tab I go to, the source isn't displaying things properly. Opening the website functions as normal.

Was curious if others were able to reproduce this issue or if it's just me.

Image

dyphire:
API Doc: https://nhentai.net/api/v2/docs
Close #442

### DIFF
--- a/sources/multi.nhentai/src/lib.rs
+++ b/sources/multi.nhentai/src/lib.rs
@@ -15,7 +15,7 @@ mod settings;
 use models::*;
 
 const BASE_URL: &str = "https://nhentai.net";
-const API_URL: &str = "https://nhentai.net/api";
+const API_URL: &str = "https://nhentai.net/api/v2";
 const USER_AGENT: &str = "Mozilla/5.0 (iPhone; CPU iPhone OS 17_2 like Mac OS X) \
 						  AppleWebKit/605.1.15 (KHTML, like Gecko) GSA/300.0.598994205 \
 						  Mobile/15E148 Safari/604";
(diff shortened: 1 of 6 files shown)


sources/multi.nhentai/src/lib.rs
```diff
@@ -190,24 +185,27 @@ impl Source for NHentai {
...
-				content: PageContent::url(format!(
-					"https://i.nhentai.net/galleries/{}/{}.{}",
-					gallery.media_id,
-					i + 1,
-					extension_from_type(&page.t)
-				)),
-				..Default::default()
+			.map(|page| {
+				let path = if page.path.starts_with("http") {
+					page.path.clone()
```
kkantan: you shouldn't need a clone here if you use `into_iter` instead.

sources/multi.nhentai/src/models.rs
```diff
@@ -63,18 +80,73 @@ pub struct NHentaiTitle {
...
+			TitlePreference::English => {
+				if !value.english_title.is_empty() {
+					value.english_title.clone()
+				} else if let Some(jpn) = &value.japanese_title {
+					jpn.clone()
+				} else {
+					format!("#{}", value.id)
+				}
+			}
+		};
```
kkantan, line 130: clones shouldn't be necessary here since the function consumes `value`.

dyphire: Done

PR #444, opened 2026-03-31
feat: add zh.jmcomic

Aiko3993:
not sure why, but config.toml is missing, will push it again later... :sob


sources/zh.jmcomic/res/settings.json
```diff
@@ -0,0 +1,36 @@
...
+				"logoutTitle": "登出",
+				"method": "basic",
+				"notification": "login",
+				"refreshes": ["settings", "content"]
+			}
+		]
+	},
+	{
+		"type": "group",
+		"key": "contentBlocking",
```
kkantan: I don't think "group" settings need keys.

sources/zh.jmcomic/src/home.rs
```diff
@@ -0,0 +1,179 @@
...
+				image,
+				description: None,
+				category: None,
+				category_sub: None,
+			})
+		})
+		.collect();
+
+	if items.is_empty() {
+		return Err(error!("页面解析失败"));
```
kkantan: there's a shorthand macro for this.
```suggestion
		bail!("页面解析失败");
```

sources/zh.jmcomic/src/home.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+fn manga_link(manga: Manga) -> Link {
+	Link {
+		title: manga.title.clone(),
+		subtitle: manga.authors.as_ref().and_then(|a| a.first().cloned()),
+		image_url: manga.cover.clone(),
+		value: Some(LinkValue::Manga(manga)),
+	}
+}
```
kkantan: I believe `Link` already implements `From<Manga>`, so this shouldn't be strictly necessary, but you might be purposely handling the subtitle differently like this.

sources/zh.jmcomic/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "zh_jmcomic"
```
kkantan:
```suggestion
name = "jmcomic"
```

sources/zh.jmcomic/res/settings.json
```diff
@@ -0,0 +1,36 @@
...
+		"footer": "支持填写漫画 ID，或会出现在标题、作者、简介、类别等位置的文本；命中后将会于可见处隐藏，并适时阻止访问。",
+		"items": [
+			{
+				"type": "editable-list",
+				"key": "blockedMetadataKeywords",
+				"title": "内容屏蔽",
+				"lineLimit": 1,
+				"inline": true,
+				"placeholder": "输入需隐藏的关键字",
+				"default": [],
```
kkantan: you don't have to provide defaults for settings and keys when the defaults you're providing don't differ from the defaults if you don't provide them. it doesn't really matter though, so you can leave it if you want.

sources/zh.jmcomic/src/models.rs
```diff
@@ -0,0 +1,541 @@
...
+	}
+
+	pub fn to_chapters(&self, manga_key: &str) -> Vec<Chapter> {
+		let series = self.series.as_deref().unwrap_or(&[]);
+		let album_time: Option<i64> = self.addtime.as_deref().and_then(|s| s.parse().ok());
+
+		if series.is_empty() {
+			return vec![Chapter {
+				key: manga_key.into(),
+				title: Some("第1話".into()),
```
kkantan: this shouldn't need to be set as the title, since if excluded it will already be `第1話`, according to localization.

PR #446, opened 2026-04-01
feat: add en.aquamanga

### LINKED ISSUE #263: aquareader.net
Add aqua manga please

z3niith:
Pull Request Draft; feedback and correction appreciated
Source: [Aqua Reader](https://aquareader.net/)

closes #263


sources/en.aquareader/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+crate-type = ["cdylib"]
+
+[profile.dev]
+panic = "abort"
+
+[profile.release]
+panic = "abort"
+opt-level = "s"
+strip = true
+lto = true
```
kkantan:
```suggestion
lto = true

```

sources/en.aquareader/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "en.aquareader",
+		"name": "Aqua Reader",
```
kkantan: it seems like the actual site name for itself is "Aqua Manga". it might make sense to id the source "en.aquamanga" too, since that's what it looks like the mihon source does.
```suggestion
		"name": "Aqua Manga",
```

sources/en.aquareader/res/filters.json
```diff
@@ -0,0 +1,117 @@
...
+			"Video Games",
+			"Villainess",
+			"Virtual Reality",
+			"Voilence",
+			"Webcomic",
+			"Wuxia",
+			"Zombies"
+		]
+	}
+]
```
kkantan:
```suggestion
]

```

sources/en.aquareader/res/icon.png
kkantan, line 1: if you could get a higher quality icon somehow, that would be good.

kkantan (changes requested):
aqua-reader.net uses the madara template, so instead of writing your own parsing you should see if you can use the template we have in this repo. [here's the mihon source for reference](https://github.com/keiyoushi/extensions-source/blob/0c91b69b6f33eacd2b3b9823330c0bed0d60a5fa/src/en/aquamanga/src/eu/kanade/tachiyomi/extension/en/aquamanga/AquaManga.kt#L8).

sources/en.aquareader/res/icon.png
z3niith: Yes, the icon was another concern I had. I can't find a higher quality version as of yet, but I'll look around.

z3niith: > if you could get a higher quality icon somehow, that would be good.

Aqua Manga's favicon/icon found on the site (current) is extremely small in size, that when it is resized to fit the required 128x128 Aidoku format, it gets heavily pixelated. I have found alternatives from the [Keiyoushi source](https://github.com/keiyoushi/extensions-source) and I was wondering if it'd be best to use one of those as a replacement. 

I am currently looking at [this](https://github.com/keiyoushi/extensions-source/blob/main/src/en/aquamanga/res/mipmap-xxxhdpi/ic_launcher.png) to resize into the 128x128.

image

sources/en.aquamanga/res/filters.json
```diff
@@ -0,0 +1,38 @@
...
+		"type": "select",
+		"id": "status",
+		"title": "Status",
+		"options": [
+			"All",
+			"Completed",
+			"Ongoing"
+		]
+	}
+]
```
kkantan: all the files should have a trailing newline, including this file, `source.json`, and `Cargo.toml`.
```suggestion
]

```

sources/en.aquamanga/res/source.json
kkantan, line 1: json file indentation should use tabs instead of spaces.

sources/en.aquamanga/res/icon.png
kkantan, line 1: this icon is much better, but it might be good if there was a little more padding like the mihon source icon has. you could scale the image down a bit and fill in the background with the same blue color.

kkantan (changes requested):
it looks like you have two folders now with duplicate files, just switch to using the `en.aquamanga` folder.

sources/en.aquamanga/res/icon.png
kkantan: could you fill in the edges with the blue background color so that there's no extra corner radius on the icon? since aidoku is the one that decides the icon corner radius, we don't want to have the image use a separate one. this will also make it pass `aidoku verify`.

sources/en.aquamanga/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+		let mut components: Vec<HomeComponent> = Vec::new();
+
+		let latest_html = Request::get(format!("{}/", BASE_URL))?.html()?;
+		let mut latest_entries: Vec<MangaWithChapter> = Vec::new();
+		if let Some(items) = latest_html.select(".page-item-detail") {
+			for item in items.take(10) {
+				let key = strip_base(
+					item.select_first("a")
+						.and_then(|a| a.attr("href"))
+						.unwrap_or_default(),
```
kkantan: since we don't want a key value to ever be an empty string, it might make more sense to do a `let Some(...) = ... else { continue; };` instead.

sources/en.aquamanga/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+		entries,
+		has_next_page,
+	})
+}
+
+register_source!(
+	Madara<AquaManga>,
+	ListingProvider,
+	Home,
+	DynamicFilters,
```
kkantan: dynamic filters seem to load inconsistently, sometimes failing. is it possible to provide genre filters statically in `filters.json` instead?

sources/en.aquamanga/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+		if let Ok(result) = parse_manga_list(&format!("{}/manga/?m_orderby=trending", BASE_URL)) {
+			components.push(HomeComponent {
+				title: Some(String::from("Trending")),
+				value: HomeComponentValue::Scroller {
+					entries: manga_to_links(result.entries),
+					listing: Some(make_listing("Trending", "Trending")),
+				},
+				..Default::default()
+			});
+		}
```
kkantan, line 284: these three components are all empty for me on the home page. do they work for you?

kkantan (changes requested):

sources/en.aquamanga/src/lib.rs
z3niith: I pushed the wrong lib.rs earlier. New commit works.

sources/en.aquamanga/res/filters.json
```diff
@@ -0,0 +1,183 @@
+[
+	{
+		"id": "type",
+		"type": "sort",
```
kkantan: it appears that the way you're using this filter would be more accurately represented with a select filter, since this is not for sorting.
```suggestion
		"type": "select",
```

sources/en.aquamanga/src/lib.rs
```diff
@@ -0,0 +1,628 @@
...
+					item.select_first(".chapter-item a")
+						.and_then(|a| a.attr("href"))
+						.unwrap_or_default(),
+				);
+				let chapter_title = item.select_first(".chapter-item a").and_then(|a| a.text());
+				let date_uploaded = item
+					.select_first(".post-on .c-new-tag")
+					.and_then(|a| a.attr("title"))
+					.map(|s: String| helpers::parse_chapter_date(params, s.trim()));
+				if !key.is_empty() && !title.is_empty() {
```
kkantan: rather than allocating empty strings, it's better to keep them as optionals. either `continue` the loop, or use `filter_map` on an iterator instead of pushing to a mutable vector.

sources/en.aquamanga/src/lib.rs
```diff
@@ -0,0 +1,628 @@
...
+			value: HomeComponentValue::MangaChapterList {
+				page_size: Some(5),
+				entries: latest_entries,
+				listing: Some(make_listing("Latest Updates", "Latest Updates")),
+			},
+			..Default::default()
+		});
+
+		let mut popular_entries: Vec<Manga> = Vec::new();
+		if let Ok(popular) = parse_manga_list(&format!("{}/manga/?m_orderby=views", BASE_URL)) {
```
kkantan: you're making a lot of serial requests for the home page content, when instead you could be parsing the actual home page data that contains everything. either do that, or use `Request::send_all` to parallelize the requests to speed up the loading.

sources/en.aquamanga/src/lib.rs
```diff
@@ -0,0 +1,628 @@
...
+			}
+		}
+		components.push(HomeComponent {
+			title: Some(String::from("Popular Series")),
+			value: HomeComponentValue::BigScroller {
+				entries: popular_entries,
+				auto_scroll_interval: Some(5.0),
+			},
+			..Default::default()
+		});
```
kkantan, line 212: on the website home page, this component is shown at the top, so I think it would make more sense to match that with the aidoku source too.
1. Popular Series
2. New Comic Releases
3. Latest Updates

sources/en.aquamanga/src/lib.rs
```diff
@@ -0,0 +1,628 @@
...
+					}]),
+				},
+				FilterItem {
+					title: String::from("Action"),
+					values: Some(vec![FilterValue::MultiSelect {
+						id: String::from("genre[]"),
+						included: vec![String::from("action")],
+						excluded: vec![],
+					}]),
+				},
```
kkantan: all of these genre filter items could be generated from a loop over titles and ids rather than written out by hand, which would simplify the code.

kkantan (changes requested):

sources/en.aquamanga/res/filters.json
z3niith: Yes, this makes more sense. I'll add this suggestion.

sources/en.aquamanga/src/lib.rs
```diff
@@ -0,0 +1,531 @@
...
+	) -> ContentRating {
+		if let Some(ref tags) = manga.tags
+			&& tags.iter().any(|t| t.eq_ignore_ascii_case("ecchi"))
+		{
+			return ContentRating::Suggestive;
+		}
+		ContentRating::Safe
+	}
+
+	fn get_search_manga_list(
```
kkantan: sorry I forgot to add my comment to the last review, but:
> it doesn't seem like overriding this function from the template is strictly necessary, as long as you follow the established conventions for the filters. for example, the id for the status filter should be `status[]` instead of `status`, and the sort method order should be the same as other madara sources.

sources/en.aquamanga/src/lib.rs
```diff
@@ -0,0 +1,531 @@
...
+					title: m.title,
+					image_url: m.cover,
+					value: Some(LinkValue::Manga(Manga {
+						key: m.key,
+						..Default::default()
+					})),
+					..Default::default()
+				})
+				.collect()
+		};
```
kkantan: is this necessary? `Link` implements `From<Manga>`, so you should just be able to do `.into()` on the `Manga` objects instead of re-implementing it yourself.

kkantan (changes requested):

kkantan:
it's defined in aidoku-rs here: https://github.com/Aidoku/aidoku-rs/blob/59788187c937bbf960a8f1eb2b4d4e86a5a75a21/crates/lib/src/structs/home.rs#L208
I'm a little unsure about the passing of only the manga key into the link value. since the data is available to be passed in I would assume that it would expect it to be done, though I am not entirely sure.

sources/en.aquamanga/src/lib.rs
kkantan: is overriding this method necessary, rather than just providing the correct selectors in the params? it appears that the mihon source doesn't make any notable changes from the template, which the template here is also based on.

sources/en.aquamanga/src/lib.rs
kkantan: this is still not resolved, I'm not quite sure why you keep marking it as such without any justification...

sources/en.aquamanga/src/lib.rs
z3niith: you're right that most of it can be moved out. The status values can go into an `ids` array in `filters.json`, and changing the sort filter's `id` to `m_orderby` handles the key name; however, there'll still be a small amount of custom logic left though since the type filter has to remap to `genre[]` as the url key, and the `has_next_page` selector can't be set via params. I'll trim it down to just those two things. let me know if you see it differently though

z3niith: I'll make these suggested changes and see if they all operate normally.

sources/en.aquamanga/src/lib.rs
z3niith: that's my mistake for marking it resolved without actually fixing it. I'll use `Request::send_all` to batch the requests. there are a few examples in the repo I can reference. let me know if you have any other thoughts on the approach

sources/en.aquamanga/src/lib.rs
kkantan: given that the type filter options are already included in the genres filter i don't think it's necessary to separate them. you're better off matching the site filter options.

the `has_next_page` selector doesn't need to be changed it doesn't look like, since the default one already works.

z3niith: After looking further into it the type filter options map to genre[] just fine, and the default has_next_page selector works, so I've removed the override entirely and handled everything through filters.json instead. let me know if you see anything else once I commit later today.

kkantan (approved):
I think the only remaining issue this source has is the rather slow home page load time, but I suppose that's a tradeoff you're choosing to make to have more detailed data.

PR #450, opened 2026-04-03
feat: add zh.noyacg

chibijoam:
first manga source i'll be maintaining
main references: multi.mangadex & zh.zaimanhua
there might be some oversights, needs review :D


sources/zh.noyacg/res/settings.json
```diff
@@ -0,0 +1,56 @@
...
+		"footer": "選擇分流伺服器。若遇到連線問題，可嘗試切換至其他伺服器。",
+		"items": [
+			{
+				"type": "select",
+				"key": "server",
+				"title": "分流伺服器",
+				"values": ["noymanga.com", "noyteam.online", "noy.asia"],
+				"titles": ["伺服器 1", "伺服器 2", "伺服器 3"],
+				"default": "noy.asia"
+			}
```
kkantan, line 54: you may want to consider adding `"refreshes": ["content"]` so that it can refetch the home page after the setting changes.

sources/zh.noyacg/src/models.rs
```diff
@@ -0,0 +1,445 @@
...
+				let tags = m.tags();
+				let status = m.manga_status();
+				let content_rating = m.content_rating();
+				let author_name = format_names(&m.author);
+				let authors = author_name.as_deref().map(|a| vec![a.into()]);
+				MangaWithChapter {
+					manga: Manga {
+						title: m.name,
+						authors,
+						tags: Some(tags),
```
kkantan: I think it might be better not to include tag data for chapter lists since it's never displayed, and there can be issues if the home page is given too much data.

sources/zh.noyacg/src/models.rs
```diff
@@ -0,0 +1,445 @@
...
+	pub time: i64,
+	#[serde(rename = "Len")]
+	pub len: i32,
+	#[serde(rename = "Status")]
+	pub status: i32,
+	#[serde(rename = "RatingSUM")]
+	pub rating: f32,
+	#[serde(rename = "Adult", default)]
+	pub adult: i32,
+}
```
kkantan, line 106: I think you could use `#[serde(rename_all = "PascalCase")]` on the struct to not have to provide a rename for every attribute, but it's up to you.

sources/zh.noyacg/src/models.rs
```diff
@@ -0,0 +1,445 @@
...
+		let cover = Some(format!("{img_base}/{key}/m1.webp"));
+		let tags = m.tags();
+		let status = m.manga_status();
+		let content_rating = m.content_rating();
+		let authors = format_names(&m.author).map(|a| vec![a]);
+		let description = if m.description.is_empty() {
+			None
+		} else {
+			Some(m.description)
+		};
```
kkantan, line 169: this is populating a lot of extra data in cases where it isn't necessary, which we've seen lead to performance issues in other sources. in search results, only the key, title, and cover should be provided, and optionally status and content rating since they're cheap. tags, authors, and descriptions should be avoided. in the hero home component this data should be parsed though.

sources/zh.noyacg/src/models.rs
```diff
@@ -0,0 +1,445 @@
...
+}
+
+#[derive(Deserialize)]
+pub struct ChapterEntry {
+	pub id: i64,
+	pub name: String,
+	pub count: i32,
+	pub sort: Option<i32>,
+	#[serde(rename = "created_at")]
+	pub created_at: Option<i64>,
```
kkantan: is this rename necessary?

sources/zh.noyacg/src/models.rs
```diff
@@ -0,0 +1,445 @@
...
+			};
+		};
+		let tags = m.tags();
+		let status = m.manga_status();
+		let content_rating = m.content_rating();
+		let update_strategy = if m.is_finished() {
+			UpdateStrategy::Never
+		} else {
+			UpdateStrategy::Always
+		};
```
kkantan: are we certain that the site never updates entries that are marked as finished? even metadata?

sources/zh.noyacg/res/icon.png
kkantan, line 1: the mihon source seems to use [a scaled up version of the site favicon](https://github.com/keiyoushi/extensions-source/blob/bbffe0d72a7d55ae46d15cecfa1fc77913d8be18/src/zh/noyacg/res/mipmap-xxxhdpi/ic_launcher.png), which looks more like a logo. is it possible to use it as well?

sources/zh.noyacg/res/settings.json
```diff
@@ -0,0 +1,56 @@
...
+		"title": "帳號",
+		"items": [
+			{
+				"type": "login",
+				"key": "login",
+				"title": "登入",
+				"logoutTitle": "登出",
+				"method": "basic",
+				"notification": "login",
+				"refreshes": ["settings", "content"]
```
kkantan: since you don't have dynamic settings, I don't think refreshing settings will do anything, but you do have dynamic filters that depend on being logged in to fetch.
```suggestion
				"refreshes": ["filters", "content"]
```

sources/zh.noyacg/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+}
+
+impl ListingProvider for NoyAcg {
+	fn get_manga_list(&self, listing: Listing, page: i32) -> Result<MangaPageResult> {
+		ensure_session()?;
+		let adult = get_adult_mode();
+		let base_url = get_base_url();
+		let referer = format!("{base_url}/");
+		let page_str = page.to_string();
+		match listing.id.as_str() {
```
kkantan: this is optional, but I think this could be simplified a bit since you're calling `post_form_listing` in almost every branch with only the url and body changing:
```rust
let (url, body) = match listing.id.as_str() {
	"latest" => (
		format!("{base_url}/api/b1/booklist"),
		build_form_body(&[("page", &page_str), ("sort", "new")])
	)
	...
	"random" => {
		let resp: ListingResp = ...;
		return Ok(resp.into_random_result());
	}
	_ => bail!("未知的列表類型"),
};
post_form_listing(
	&url,
	&body,
	&referer,
	&adult,
	page,
)
```

sources/zh.noyacg/res/settings.json
ghost: i did notice this, but from what i can tell, dynamic filters don’t seem to update immediately — they only show up after re-entering the source

the app itself seems to support it in theory, but i’m not seeing it take effect — not sure if this is something wrong on my side, or a potential issue, see:
https://github.com/Aidoku/Aidoku/blob/main/iOS/New/Views/Source/SourceSettingsView.swift#L93-L97
https://github.com/Aidoku/Aidoku/blob/main/iOS/New/Views/Source/SourceSettingsView.swift#L136-L138
https://github.com/Aidoku/Aidoku/blob/main/iOS/New/Views/Common/Settings/SettingView.swift#L216-L219
https://github.com/Aidoku/Aidoku/blob/main/iOS/New/Views/Source/SearchFilterHeaderView.swift#L61-L65
https://github.com/Aidoku/aidoku-rs/blob/main/crates/lib/src/structs/setting.rs#L277-L281

also, i previously tried replicating some official app behaviors using dynamic settings — ended up removing that approach later, but forgot to clean up this part, thanks for pointing it out

sources/zh.noyacg/res/icon.png
ghost: in fact, i used the icon from the official app, but i do agree this one looks better — i’ll switch to it

sources/zh.noyacg/res/settings.json
ghost: > i did notice this, but from what i can tell, dynamic filters don’t seem to update immediately — they only show up after re-entering the source
> 
> the app itself seems to support it in theory, but i’m not seeing it take effect — not sure if this is something wrong on my side, or a potential issue, see: https://github.com/Aidoku/Aidoku/blob/main/iOS/New/Views/Source/SourceSettingsView.swift#L93-L97 https://github.com/Aidoku/Aidoku/blob/main/iOS/New/Views/Source/SourceSettingsView.swift#L136-L138 https://github.com/Aidoku/Aidoku/blob/main/iOS/New/Views/Common/Settings/SettingView.swift#L216-L219 https://github.com/Aidoku/Aidoku/blob/main/iOS/New/Views/Source/SearchFilterHeaderView.swift#L61-L65 https://github.com/Aidoku/aidoku-rs/blob/main/crates/lib/src/structs/setting.rs#L277-L281
> 
> also, i previously tried replicating some official app behaviors using dynamic settings — ended up removing that approach later, but forgot to clean up this part, thanks for pointing it out

oops, to clarify, it looks like i commented on the wrong part — what i originally meant was that I had previously added `refreshes: ["filters"]` to the content rating setting, but it doesn’t seem to trigger an immediate filter update when the content rating changes, any filter changes still require re-entering the source

from my observation, filters seem to be constructed when first entering the source or after logging in, and in the current behavior, whether filters is included in refreshes doesn’t make a difference — at least on my side, the effect appears identical

i haven’t verified whether filters are considered part of content, but my assumption is that, at the current stage, including filters here may not be necessary — let me know if i’m mistaken

kkantan: according to this code: https://github.com/Aidoku/Aidoku/blob/main/iOS/New/Views/Source/SearchFilterHeaderView.swift#L61-L65
it appears as though having a filters refresh should refetch the dynamic filters. if it's not working properly for you, I suppose it is likely a bug with the app.

it looks like it's misdocumented in aidoku-rs since `filters` does appear to be a valid refreshes value: https://github.com/Aidoku/aidoku-rs/blob/main/crates/lib/src/structs/setting.rs#L277-L281

PR #463, opened 2026-04-05
fix(multi.nhentai): pages not loading

### LINKED ISSUE #456: nhentai.net - Failed to Load Chapter
Image

As screenshot shows. Happens for all the works I tried to open.

Note: This was after moving to Version 0.8.2.

dyphire:
nhentai changed the api again.

Fix #456

### DIFF
--- a/sources/multi.nhentai/src/lib.rs
+++ b/sources/multi.nhentai/src/lib.rs
@@ -12,6 +12,8 @@ mod home;
 mod models;
 mod settings;
 
+use core::cell::RefCell;
+
 use models::*;
 
 const BASE_URL: &str = "https://nhentai.net";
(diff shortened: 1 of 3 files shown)


sources/multi.nhentai/src/lib.rs
```diff
@@ -179,28 +185,42 @@ impl Source for NHentai {
...
+					.json_owned()?;
+				gallery
+			}
+		} else {
+			let api_url = format!("{API_URL}/galleries/{}", chapter.key);
+			let gallery: NHentaiGallery = Request::get(&api_url)?
+				.header("User-Agent", USER_AGENT)
+				.json_owned()?;
+			gallery
+		};
```
kkantan, line 211: you should try and avoid the code duplication here, maybe with something like this:
```rust
let gallery: NHentaiGallery = match &*maybe_cached {
    Some((cached_key, cached_gallery)) if cached_key == &chapter.key => {
        cached_gallery.clone()
    }
    _ => {
        let api_url = format!("{API_URL}/galleries/{}", chapter.key);
        Request::get(api_url)?
            .header("User-Agent", USER_AGENT)
            .json_owned()?
    }
};
```

kkantan (changes requested):

sources/multi.nhentai/src/lib.rs
dyphire: Done

kkantan (approved):

PR #473, opened 2026-04-16
feat: add fr.fmteam

### LINKED ISSUE #405: fmteam.fr
A new FR source for Aidoku:

https://fmteam.fr/

Zareix:
Add source for [FMTeam.fr](https://fmteam.fr)


sources/fr.fmteamfr/res/icon.png
kkantan, line 1: the icon should be 128x128 with a solid background. you can add a white or black background to this and scale it down a little.

sources/fr.fmteamfr/res/source.json
```diff
@@ -0,0 +1,10 @@
+{
+	"info": {
+		"id": "fr.fmteamfr",
+		"name": "FMTeam.fr",
+		"version": 1,
+		"url": "https://fmteam.fr/",
```
kkantan: the source url should have no trailing slash.
```suggestion
		"url": "https://fmteam.fr",
```

sources/fr.fmteamfr/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    prelude::*,
+};
+use chrono::{DateTime, Utc};
+
+const API_BASE: &str = "https://fmteam.fr/api";
+const SITE_BASE: &str = "https://fmteam.fr";
+
+struct Fmteamfr;
+
+#[derive(serde::Deserialize)]
```
kkantan: it's nice to separate serde structs into a `models.rs` file and miscellaneous functions into a `helpers.rs` file.

sources/fr.fmteamfr/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    alloc::{String, Vec, format, vec},
+    imports::net::Request,
+    prelude::*,
+};
+use chrono::{DateTime, Utc};
+
+const API_BASE: &str = "https://fmteam.fr/api";
+const SITE_BASE: &str = "https://fmteam.fr";
+
+struct Fmteamfr;
```
kkantan: I would just name this `FMTeam`.

kkantan (changes requested):
since the corresponding mihon source for this website uses a "pizzareader" template, it would be nice to also refactor this into a template for this repo as well. you can look at other templates like `madara` for reference.

[here is the code for reference](https://github.com/keiyoushi/extensions-source/blob/main/lib-multisrc/pizzareader/src/eu/kanade/tachiyomi/multisrc/pizzareader/PizzaReader.kt).

Zareix:
I've made the requested changes.
I'm not a Rust expert but I'll try taking a look to implement this pizzareader template

Zareix:
I tried an implementation of PizzaReader tell me what you think

sources/fr.fmteam/res/icon.png
kkantan, line 1: could you scale the logo down a bit so that there's more padding around it?

sources/fr.fmteam/res/source.json
```diff
@@ -0,0 +1,10 @@
+{
+  "info": {
+    "id": "fr.fmteam",
+    "name": "FMTeam",
```
kkantan: I think it's still worth calling it "FMTeam.fr", since that is the title on the website.
```suggestion
    "name": "FMTeam.fr",
```

sources/fr.fmteam/res/source.json
kkantan, line 1: json files should use tabs for indentation instead of spaces.

sources/fr.fmteam/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+	}
+
+	fn params(&self) -> Params {
+		Params {
+			base_url: BASE_URL.into(),
+		}
+	}
+}
+
+register_source!(PizzaReader<Fmteam>, DeepLinkHandler, Home);
```
kkantan:
```suggestion
struct FMTeam;

impl Impl for FMTeam {
	fn new() -> Self {
		Self
	}

	fn params(&self) -> Params {
		Params {
			base_url: BASE_URL.into(),
		}
	}
}

register_source!(PizzaReader<FMTeam>, DeepLinkHandler, Home);
```

templates/pizzareader/src/imp.rs
```diff
@@ -0,0 +1,233 @@
...
+			artists: comic
+				.artist
+				.filter(|a| !a.is_empty())
+				.map(|a| vec![a.clone()]),
+			viewer: aidoku::Viewer::RightToLeft,
+			content_rating: self.to_manga_content_rating(comic.adult),
+			status: self.to_manga_status(comic.status.unwrap_or("".into())),
+			..Default::default()
+		}
+	}
```
kkantan, line 52: you can implement `From<PizzaComicDto> for Manga` to do this in a more standard way. you also shouldn't need to clone anything, including the artist.

templates/pizzareader/src/imp.rs
```diff
@@ -0,0 +1,233 @@
...
+	prelude::*,
+};
+use chrono::{DateTime, Utc};
+
+pub trait Impl {
+	fn new() -> Self;
+
+	fn params(&self) -> Params;
+
+	fn to_manga_status(&self, status: String) -> aidoku::MangaStatus {
```
kkantan: usually functions should take string slices, unless you specifically need an allocated string.
```suggestion
	fn to_manga_status(&self, status: &str) -> aidoku::MangaStatus {
```

templates/pizzareader/src/imp.rs
```diff
@@ -0,0 +1,233 @@
...
+	fn get_search_manga_list(
+		&self,
+		params: &Params,
+		query: Option<String>,
+		_page: i32,
+		_filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		let url = if let Some(q) = query {
+			if q.len() > 2 {
+				format!("{}/api/search/{}", params.base_url, q)
```
kkantan: `q` probably needs to be url encoded, with the `encode_uri_component` helper function.

templates/pizzareader/src/imp.rs
```diff
@@ -0,0 +1,233 @@
...
+	fn get_manga_update(
+		&self,
+		params: &Params,
+		mut manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		let slug = manga.key.trim();
+		if slug.is_empty() {
+			return Err(AidokuError::message("Manga key is empty"));
```
kkantan: there's a helper macro for this.
```suggestion
			bail!("Manga key is empty");
```

templates/pizzareader/src/imp.rs
```diff
@@ -0,0 +1,233 @@
...
+		}
+
+		let response: PizzaResultDto =
+			Request::get(format!("{}/api/comics/{slug}", params.base_url))?.json_owned()?;
+		let comic = response.comic.ok_or_else(|| {
+			AidokuError::message(format!(
+				"Comic not found with {}/api/comics/{slug}",
+				params.base_url
+			))
```
kkantan:
```suggestion
			error!(
				"Comic not found with {}/api/comics/{slug}",
				params.base_url
			)
```

templates/pizzareader/src/imp.rs
```diff
@@ -0,0 +1,233 @@
...
+		})?;
+
+		if needs_details {
+			manga = self.to_manga(comic.clone(), &params.base_url);
+		}
```
kkantan: if you move this below the `needs_chapters` check, you shouldn't need to clone the `comic` object.

templates/pizzareader/src/imp.rs
```diff
@@ -0,0 +1,233 @@
...
+			Request::get(format!("{}/api{}", params.base_url, chapter_path))?.json_owned()?;
+
+		let pages = response
+			.chapter
+			.map(|chapter| {
+				chapter
+					.pages
+					.into_iter()
+					.map(|url| Page {
+						content: aidoku::PageContent::Url(url, None),
```
kkantan:
```suggestion
						content: aidoku::PageContent::url(url),
```

templates/pizzareader/src/models.rs
```diff
@@ -0,0 +1,75 @@
...
+pub struct PizzaChapterDto {
+	#[serde(default)]
+	pub chapter: Option<i32>,
+	#[serde(rename = "full_title", default)]
+	pub full_title: String,
+	#[serde(default)]
+	pub title: Option<String>,
+	#[serde(default)]
+	pub pages: Vec<String>,
+	#[serde(rename = "published_on", default)]
```
kkantan: it doesn't look like these two serde renames are necessary?

templates/pizzareader/src/imp.rs
```diff
@@ -0,0 +1,233 @@
...
+		if needs_chapters {
+			let mut chapters: Vec<Chapter> = Vec::new();
+
+			for chapter in comic.chapters {
+				let chapter_title = chapter
+					.title
+					.filter(|t| !t.is_empty())
+					.or(Some(chapter.full_title));
+
+				let date_uploaded = DateTime::parse_from_rfc3339(&chapter.published_on)
```
kkantan: I'm not sure if it's really worth having chrono as an additional dependency when this can be done with the built-in date parsing: `parse_date(&chapter.published_on, "yyyy-MM-dd HH:mm:ss'Z'");`

kkantan (changes requested):
the template looks like it will work, thanks.

Zareix:
Btw, how can I test the deeplink ? Not sure when this is triggered

templates/pizzareader/src/imp.rs
```diff
@@ -0,0 +1,286 @@
...
+use crate::models::*;
+use aidoku::{
+	Chapter, DeepLinkResult, FilterValue, HomeLayout, Manga, MangaPageResult, Page, Result,
+	alloc::{String, Vec, vec},
+	helpers::uri::encode_uri_component,
+	imports::{net::Request, std::parse_date},
+	prelude::*,
+};
+
+impl From<PizzaComicDto> for Manga {
```
kkantan: I think it's best to put this in the `models.rs` file along with the `PizzaComicDto` definition.

templates/pizzareader/src/imp.rs
```diff
@@ -0,0 +1,286 @@
...
+					author,
+					description,
+					title,
+					thumbnail,
+					url,
+					status,
+					adult,
+					last_chapter,
+					genres,
+				},
```
kkantan: is it necessary to extract these variables and then reconstruct a new `PizzaComicDto`? I think something simpler like this would work:

```rust
let comic = ...;

if needs_chapters {
	manga.chapters = Some(comic.chapters.into_iter().map(...).collect());
}

if needs_details {
	manga.copy_from(self.to_manga(comic, &params.base_url));
}
```

kkantan (changes requested):
I tested the source on device and just had two comments:

firstly, looking at the `models.rs` file and the website, it seems there is api support for genres, but there are no genres populated for manga. can this be added to the `From<PizzaComicDto>` implementation, or are the genres served by the api not complete?

secondly, the home page for fmteam differs from the actual website page. the website home page has a title in french and has chapter number/title/upload date as well. I think it's worth either trying to copy this more closely, *or* just ignoring the home page, since the actual number of series the website has is rather small. for the latter option, you could also add in your own sort filter that could achieve the same results as sorting by the last updated chapter if you wanted.

templates/pizzareader/src/imp.rs
Zareix: I get :
```use of partially moved value: comic
partial move occurs because comic.chapters has type Vec<PizzaChapterDto>, which does not implement the Copy trait (rustc E0382)
```

Zareix: Can be fix like that maybe ?
```rust
let comic = ...

if needs_chapters {
	manga.chapters = Some(
		comic
			.chapters
			.iter()
			.map(|chapter| Chapter {
				key: chapter.url.clone(),
				title: chapter
					.title
					.clone()
					.filter(|t| !t.is_empty())
					.or(Some(chapter.full_title.clone())),
				chapter_number: chapter.chapter.map(|n| n as f32),
				volume_number: chapter.volume.map(|n| n as f32),
				date_uploaded: parse_date(
					&chapter.published_on,
					"yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'",
				),
				..Default::default()
			})
			.collect(),
	);
}

if needs_details {
	manga.copy_from(self.to_manga(comic, &params.base_url));
}
	```

**Comment by Zareix** [comment]
Indeed I forgot about the genres. I implemented them along with a function to parse them as some are strangely aggregated.
Also removed home page and added sorting

File: templates/pizzareader/src/models.rs
Comment (kkantan) [inline templates/pizzareader/src/models.rs:1]: I don't think you need clone implemented for any of these structs, and it's not great to have since you typically want to avoid clones whenever possible, but it doesn't matter too much.

File: templates/pizzareader/src/imp.rs [hunk 17]
```diff
@@ -0,0 +1,271 @@
...
+		let comic = self
+			.get_manga_details(&params.base_url, slug)?
+			.comic
+			.ok_or_else(|| error!("Comic not found with {slug}"))?;
+
+		if needs_chapters {
+			manga.chapters = Some(
+				comic
+					.chapters
+					.iter()
```
kkantan: can you use `into_iter` here to avoid cloning the chapter urls and titles? you may need to make `comic` mutable and the `chapters` field optional, so that you can `take()` it and not consume the object. I think it's worth it to avoid the potential three clones for each chapter.

kkantan (changes requested):
there are a couple of clippy lints also that should be fixed, but other than that this should be good.

kkantan (approved):

Zareix:
Thank you !

PR #480, opened 2026-04-24
fix(zh.bilimanga): manga details parsing failed

dyphire:


sources/zh.bilimanga/src/html/mod.rs
```diff
@@ -79,6 +79,16 @@ pub trait MangaPage {
 
 impl MangaPage for Document {
 	fn update_details(&self, manga: &mut Manga) -> Result<()> {
+		// If the page is an error/notice page (e.g. removed content),
+		// Bail early and put the message into manga.description so caller
+		// can surface the info instead of failing on missing selectors.
+		if let Some(err_el) = self.select_first(".aui-ver-form") {
+			let msg = err_el.text().unwrap_or_default().trim().to_string();
+			if !msg.is_empty() {
+				manga.description = Some(msg);
```
kkantan, line 109: is this better that just returning an error message result? that seems to be more like how it would be intended to be used, rather than subverting the usage of the description field.

dyphire: Although the website prompt content has been removed, chapter parsing can still work normally, so we should not directly report an error to interrupt parsing. Not sure if these removed manga actually requires login permission to read, because the website has closed registration and no account is available for testing.

dyphire:
Fix search interface changes

kkantan (approved):

PR #483, opened 2026-04-24
fix(zh.happymh): pages not loading

dyphire:


kkantan (approved):
I'm not sure if the clones are necessary, but it's okay.

PR #487, opened 2026-04-27
feat: add zh.komiic

### LINKED ISSUE #458: komiic.com
Source request
Lang: ZH
URL:https://komiic.com/

Kanscape:
Adds a Komiic source with search, filters, home layout, login support, personalized recommendations, chapter handling, image requests, and deep links.

https://komiic.com/

Validation:
- cargo fmt --check
- cargo check
- cargo clippy --target wasm32-unknown-unknown
- cargo test
- aidoku package
- aidoku verify sources/zh.komiic/package.aix

Closes #458


sources/zh.komiic/res/filters.json
```diff
@@ -0,0 +1,105 @@
+[
+	{
+		"type": "select",
+		"id": "sort",
+		"title": "排序",
+		"options": ["更新", "观看数", "喜爱数"],
+		"ids": ["DATE_UPDATED", "VIEWS", "FAVORITE_COUNT"],
+		"default": "DATE_UPDATED"
+	},
```
kkantan, line 8: I think it's better to use the sort filter type so that the app ui matches accordingly. you can set `canAscend` to `false` and it will behave the same, besides needing to define the ids inside the source and using the index value it stores instead.

sources/zh.komiic/src/lib.rs
```diff
@@ -0,0 +1,1008 @@
...
+favoriteCount
+lastBookUpdate
+lastChapterUpdate
+description
+__typename
+"#;
+
+struct KomiicSource;
+
+impl KomiicSource {
```
kkantan, line 75: this is a pretty large singular file that could benefit from splitting some related blocks of code into different files. you don't need to do all of this, but here are some common files we usually use for sources for ideas:
- `settings.rs`: settings/defaults handling
- `helpers.rs`: miscellaneous helper functions
- `home.rs`: home page parsing/handling

sources/zh.komiic/src/lib.rs
```diff
@@ -0,0 +1,1008 @@
...
+								} else {
+									keyword = Some(String::from(value));
+								}
+							}
+							"tag" => keyword = Some(String::from(value)),
+							_ => {}
+						}
+						break;
+					}
+				}
```
kkantan: does this branch ever get matched? it doesn't look like you have any multi-select filters.

sources/zh.komiic/src/lib.rs
```diff
@@ -0,0 +1,1008 @@
...
+							}
+						}
+						"genre" => {
+							if let Some(id) = Self::category_filter_value(value) {
+								category = id;
+							} else {
+								keyword = Some(String::from(value));
+							}
+						}
+						"tag" | "author" | "artist" | "title" => {
```
kkantan: is author and artist search supported? you should enable it in the source.json config options in that case.

I also don't see where the `tag` filter would come from.

sources/zh.komiic/src/lib.rs
```diff
@@ -0,0 +1,1008 @@
...
+			_ => MangaStatus::Unknown,
+		}
+	}
+
+	fn parse_manga(value: &Value) -> Manga {
+		let key = Self::string_field(value, "id").unwrap_or_default();
+		let title = Self::string_field(value, "title").unwrap_or_else(|| key.clone());
+		let author = Self::first_name(value, "authors");
+		Manga {
+			key: key.clone(),
```
kkantan: if you put the `url` value in a variable before defining the struct, the `key` string shouldn't need to be cloned.

sources/zh.komiic/src/lib.rs
```diff
@@ -0,0 +1,1008 @@
...
+			..Default::default()
+		}
+	}
+
+	fn parse_manga_list(value: &Value, path: &[&str]) -> Result<Vec<Manga>> {
+		let mut current = value;
+		for key in path {
+			current = current
+				.get(*key)
+				.ok_or_else(|| aidoku::AidokuError::message(format!("Komiic missing {key}")))?;
```
kkantan: you can use the shorthand macro for these.
```suggestion
				.ok_or_else(|| error!("Komiic missing {key}"))?;
```

sources/zh.komiic/src/lib.rs
```diff
@@ -0,0 +1,1008 @@
...
+				"pagination": {
+					"limit": CATEGORY_PAGE_SIZE,
+					"offset": Self::offset(page, CATEGORY_PAGE_SIZE),
+					"orderBy": order_by,
+					"asc": false,
+					"status": status
+				}
+			},
+			"query": format!(
+				"query comicByCategories($categoryId: [ID!]!, $pagination: Pagination!) {{ comicByCategories(categoryId: $categoryId, pagination: $pagination) {{ {COMIC_FIELDS} }} }}"
```
kkantan: you could split these long strings across multiple lines by using backslashes if you wanted, but this is optional. sometimes long strings can be annoying to work with since they tend to break rustfmt.
```suggestion
				"query comicByCategories($categoryId: [ID!]!, $pagination: Pagination!) {{ \
				    comicByCategories(categoryId: $categoryId, pagination: $pagination) {{ \
				        {COMIC_FIELDS} \
				    }} \
				}}"
```

kkantan (changes requested):
thank you for the pr, your code looks pretty good.

Kanscape:
Thanks for the review! 

This is one of my first times submitting a PR, so I apologize in advance for any non-standard practices. I've addressed the requested changes in `d729a84`:

- Changed the sort filter to use the sort type with index mapping.
- Removed unused `MultiSelect` and unsupported author/artist/tag handling.
- Switched missing response errors to `error!`.
- Avoided the extra `key` clone.
- Split the source into `helpers.rs`, `home.rs`, and `settings.rs`.
- Fixed pagination to use the raw response count before deduplication.
- Added fallback chapter selection when the preferred chapter version is empty.

I also reran `fmt`, `check`, `clippy`, `tests`, `package`, and `verify` locally.

Kanscape:
Update in `80c64d8`.

Additional changes since my previous comment:

- Implemented and declared author search via Komiic's `searchComicsAndAuthors` and `getComicsByAuthor` APIs.
- Preserved all authors returned by Komiic instead of only using the first author.
- Added tests for author search and multiple authors.

I also reran `fmt`, `check`, `clippy`, `test`, `package`, and `verify` locally.

sources/zh.komiic/res/filters.json
```diff
@@ -0,0 +1,108 @@
...
+	{
+		"type": "sort",
+		"id": "sort",
+		"title": "排序",
+		"canAscend": false,
+		"options": ["更新", "观看数", "喜爱数"],
+		"default": {
+			"index": 0,
+			"ascending": false
+		}
```
kkantan: as a heads up, you don't need to provide these default values for all the filters because they're the same as the original default values, but it's fine to have them if you'd rather.

sources/zh.komiic/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+	fn parse_chapter(value: &Value, manga_key: &str) -> Option<Chapter> {
+		let key = Self::string_field(value, "id")?;
+		let serial = value
+			.get("serial")
+			.and_then(Value::as_str)
+			.unwrap_or_default();
+		let serial_number = serial.parse::<f32>().ok();
+		let is_book = value.get("type").and_then(Value::as_str) == Some("book");
+		Some(Chapter {
+			key: key.clone(),
```
kkantan: same thing here as the manga construction, if you define the url in a variable above then this doesn't need to be cloned.

kkantan (changes requested):
looking at the source running on-device, my only comment is that it appears tags are parsed for all results in searches and listings, which could be the reason I'm seeing little performance issues. in your `parse_manga` function, I would suggest having a `basic` boolean, or something along those lines, that indicates if only the minimal fields should be populated: key, title, and cover. you could maybe include the url and enum values as well, as some sources do this too. it seems like the app has difficulty handling more data than this, particularly if you scroll through many pages.

Kanscape:
Thanks for the review. The requested updates have been pushed.

Searches and regular listings now use a minimal parsing path limited to basic fields (`key`, `title`, `cover`, `url`, `status`, `content_rating`, `viewer`). GraphQL requests for paged results have also been optimized to exclude `authors` and `categories`.

`authors` and `tags` are now kept only for the home sections that need them ("本月最热" and "总排行"). Any paged listings opened from these sections will continue to use the minimal path.

Additionally, I cleaned up the redundant filter defaults and simplified the chapter URL generation, so cloning the chapter key is no longer necessary.

kkantan (approved):

PR #488, opened 2026-04-30
fix(libgroup): chapters not loading

### LINKED ISSUE #460: hentailib.me
No title from hentailib does not open, it gives a network error, although there is authorization and access to the site is also available, because if you click on the button to open the chapter in the browser, everything works immediately. The problem has been going on for 2-3 weeks

Image

### LINKED ISSUE #476: mangalib.me Doesn't load the list of chapters
image
image

remytomme:
- Add `ResponseJsonExt::parse_json` that includes a 300-char context window around the error column.
- Fix `summary` field: API changed from plain string to ProseMirror document (`{"type":"doc","content":[...]}`). Now deserializes as `LibGroupContentModel` and converts through the existing `convert_model_to_markdown` converter.
- Bump versions for all four LibGroup sources.

Closes #460, #476

### DIFF
--- a/templates/libgroup/src/json.rs
+++ b/templates/libgroup/src/json.rs
@@ -1,0 +1,31 @@
+use aidoku::{
+	AidokuError, Result,
+	alloc::String,
+	imports::net::Response,
+	prelude::*,
+};
+use serde::de::DeserializeOwned;
+
+pub trait ResponseJsonExt {
+	fn parse_json<T: DeserializeOwned>(self) -> Result<T>;
+}
+
+impl ResponseJsonExt for Response {
+	fn parse_json<T: DeserializeOwned>(self) -> Result<T> {
+		let bytes = self.get_data()?;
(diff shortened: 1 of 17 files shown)


templates/libgroup/src/json.rs
```diff
@@ -0,0 +1,31 @@
...
+			let context: String = core::str::from_utf8(&bytes)
+				.unwrap_or("<non-utf8>")
+				.chars()
+				.skip(start)
+				.take(300)
+				.collect();
+			AidokuError::message(format!(
+				"JSON error at col {col} (body {} bytes): {e} | context: ...{context}...",
+				bytes.len()
+			))
```
kkantan, line 28: you could just use the shorthand macro for this.
```suggestion
			error!(
				"JSON error at col {col} (body {} bytes): {e} | context: ...{context}...",
				bytes.len()
			)
```

kkantan (approved):

PR #494, opened 2026-05-06
feat: add en.firescans

itss0n1c:
note: version bumped to avoid conflicts with legacy


sources/en.firescans/res/source.json
```diff
@@ -0,0 +1,11 @@
+{
+	"info": {
+		"id": "en.firescans",
+		"name": "FireScans",
```
kkantan: since the site title has a space in the name, this may be a better option.
```suggestion
		"name": "Fire Scans",
```

sources/en.firescans/res/icon.png
kkantan, line 1: I think this may be a more updated icon, since the url is no longer `firecomics.org`: https://firescans.xyz/wp-content/uploads/2023/09/cropped-WhatsApp-Image-2024-07-05-at-5.27.44-PM-192x192.webp

kkantan (changes requested):

kkantan (approved):

PR #497, opened 2026-05-09
fix(vi.truyenqq): update url

tachibana-shin:


kkantan (changes requested):
if you could set `"minAppVersion": "0.7.1"` in the source json so that the verification passes that would be good.

kkantan (approved):

PR #506, opened 2026-05-11
fix(en.comix): fix fetch token

jianmingyong:
This quick fix is based on what was being used currently. I tested this and it does load the chapter list now (though the speed is really slow).

### DIFF
--- a/sources/en.comix/res/source.json
+++ b/sources/en.comix/res/source.json
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "en.comix",
 		"name": "Comix",
-		"version": 10,
+		"version": 11,
 		"url": "https://comix.to",
 		"contentRating": 1,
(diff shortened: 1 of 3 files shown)


jianmingyong:
Now that I think about it, this is probably not going to stay working for long. Then again, I am not so sure as they use the probe to find the hashing function than using “Qi / qi”.

Reference: https://github.com/keiyoushi/extensions-source/pull/15748

jianmingyong:
I have kind of implement the probing function so that it can hopefully survive changes based on the above reference.

The JavaScript is about the same, just that I made the output into a json string so I can pass it down and point to the correct function.

kkantan:
I implemented a similar fix (#507) a little bit more concisely, so I'm going to merge that instead, but thank you.

PR #519, opened 2026-05-15
fix(en.novelbuddy): rewrite to use the json api

neldra:
NovelBuddy is a text-novel site. The existing en.novelbuddy uses the madtheme template, which worked while the site served scrapable HTML; novelbuddy.com has since migrated to a Next.js site backed by a JSON API and no longer serves that HTML, so the source is broken (search returns nothing, chapters don't load). Rewrote it as a standalone source talking to api.novelbuddy.com directly, returning chapter prose via `PageContent::Text`. `minAppVersion` is 0.7 (the text reader was added in Aidoku 0.7).


sources/en.novelbuddy/res/source.json
```diff
@@ -2,9 +2,24 @@
 	"info": {
 		"id": "en.novelbuddy",
 		"name": "NovelBuddy",
-		"version": 1,
+		"version": 2,
 		"url": "https://novelbuddy.com",
-		"contentRating": 1,
-		"languages": ["en"]
-	}
+		"contentRating": 0,
```
kkantan: I think this source still contains nsfw content, so it's better to leave the content rating as `1`.

sources/en.novelbuddy/res/source.json
```diff
@@ -2,9 +2,24 @@
...
+		"languages": ["en"],
+		"minAppVersion": "0.7"
```
kkantan: by using the newer source api, this source can already only be installed on 0.7+, so this is usually not necessary to have. however, you're using some newer apis (`parse_date`) that require `0.7.1`, so if you change that to this the workflow will pass.

sources/en.novelbuddy/src/helpers.rs
```diff
@@ -0,0 +1,158 @@
+use aidoku::{
+	ContentRating, MangaStatus,
+	alloc::{String, Vec, string::ToString},
+	imports::{html::Html, std::parse_date},
+	prelude::*,
+};
+
+pub const BASE_URL: &str = "https://novelbuddy.com";
+pub const API_URL: &str = "https://api.novelbuddy.com";
```
kkantan: usually it's more standard to define these in `lib.rs` along with the other constants, then import them here. it doesn't really matter though.

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,433 @@
...
+	"rating",
+	"views",
+	"bookmarks",
+];
+const STATUS_IDS: &[&str] = &["all", "ongoing", "completed"];
 
 struct NovelBuddy;
 
-impl Impl for NovelBuddy {
+impl NovelBuddy {
```
kkantan: how come all of the functions here, particularly `request` and `resolve_slug`, aren't considers helpers and are implemented on the `NovelBuddy` struct?

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,433 @@
...
+		if !response.success {
+			let msg = response
+				.message
+				.unwrap_or_else(|| "API request failed".into());
+			bail!("{msg}");
+		}
+		response.data.ok_or_else(|| error!("API returned no data"))
+	}
+
+	fn manga_from_list_item(&self, item: TitleListItem) -> Manga {
```
kkantan: you could implement `From<TitleListItem> for Manga` instead to use the `Into` trait.

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,433 @@
...
+		Manga {
+			key: item.id,
+			title: item.name,
+			cover: item.cover,
+			url,
+			..Default::default()
+		}
+	}
+
+	fn populate_from_detail(&self, manga: &mut Manga, detail: TitleDetail) {
```
kkantan: likewise, you could instead implement `From<TitleDetail> for Manga`, and then use `manga.copy_from(detail.into())`, but it's fine to just have this.

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,433 @@
...
+		let number = parse_chapter_number(&item.name);
+		let date = item.updated_at.as_deref().and_then(parse_iso_date);
+		let url = item.url.as_deref().map(absolute_url);
+		Chapter {
+			key: item.id,
+			title: Some(item.name),
+			chapter_number: number,
+			date_uploaded: date,
+			url,
+			language: Some("en".into()),
```
kkantan: you don't need to provide this when the source uses a single language.

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,433 @@
...
+	fn get_search_manga_list(
+		&self,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		let mut sort = "popular";
+		let mut status: Option<&str> = None;
+		let mut included_genres = Vec::<String>::new();
+		let mut excluded_genres = Vec::<String>::new();
```
kkantan: I would use `Option<Vec<String>>`s instead to avoid allocation when there is no genre filter.

however, I think you also don't need any of these variables. just create the `qs` variable above the loop and populate it within the match branches rather than assigning to a variable and then adding it.

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,433 @@
...
+
+		for filter in filters {
+			match filter {
+				FilterValue::Sort { id, index, .. } if id == "sort" => {
+					if let Some(s) = SORT_IDS.get(index as usize) {
+						sort = *s;
+					}
+				}
+				FilterValue::Select { id, value }
+					if id == "status" && value != "all" && STATUS_IDS.contains(&value.as_str()) =>
```
kkantan: is `STATUS_IDS` necessary? I don't see how the value could be something not in the filters definition.

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,433 @@
...
+				}
+				_ => {}
+			}
+		}
+
+		let mut qs = QueryParameters::new();
+		qs.push("sort", Some(sort));
+		qs.push("page", Some(&page.to_string()));
+		if let Some(q) = query.as_deref()
+			&& !q.is_empty()
```
kkantan: the query won't be empty if it's `Some`.

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,433 @@
...
+
+	fn get_manga_update(
+		&self,
+		mut manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		if manga.key.is_empty() {
+			bail!("Manga key is empty");
+		}
```
kkantan: does this happen?

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,433 @@
...
+		let text = if body.is_empty() {
+			"(empty chapter)".to_string()
+		} else {
+			body
+		};
+		Ok(vec![Page {
+			content: PageContent::text(text),
+			thumbnail: None,
+			has_description: false,
+			description: None,
```
kkantan:
```suggestion
			..Default::default()
```

kkantan (changes requested):

sources/en.novelbuddy/src/lib.rs
neldra: agreed, the caller always provides the key here. removed this and the `get_page_list` key guard for the same reason, nothing else in the repo guards keys anyway.

sources/en.novelbuddy/src/helpers.rs
```diff
@@ -0,0 +1,262 @@
...
+			key: item.id,
+			title: item.name,
+			cover: item.cover,
+			url,
+			..Default::default()
+		}
+	}
+}
+
+impl From<TitleDetail> for Manga {
```
kkantan, line 61: usually I like to include trait implementations alongside the struct definitions in `models.rs`, but it doesn't really matter.

sources/en.novelbuddy/src/helpers.rs
```diff
@@ -0,0 +1,262 @@
...
+		return ContentRating::NSFW;
+	}
+	for tag in tags {
+		match tag.as_str() {
+			"Adult" | "Smut" | "Mature" | "Ecchi" | "Lolicon" | "Yaoi" | "Yuri" => {
+				return ContentRating::Suggestive;
+			}
+			_ => {}
+		}
+	}
```
kkantan, line 136: there's a simpler way of writing this, for future reference: https://github.com/Aidoku-Community/sources/blob/f1c1af8f540d844ee56e34ab502724a8a73dcfad/templates/mangabox/src/imp.rs#L159-L161
changing it isn't necessary though

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,285 @@
...
+						qs.push("exclude", Some(&excluded.join(",")));
+					}
+				}
+				_ => {}
+			}
+		}
+
+		if !has_sort {
+			qs.push("sort", Some("popular"));
+		}
```
kkantan, line 76: alternatively, you could use `qs.push` before the filters loop and then `qs.set` inside the sort match branch.

sources/en.novelbuddy/src/lib.rs
```diff
@@ -1,22 +1,285 @@
...
+		qs.push("page", Some(&page.to_string()));
+		if let Some(q) = query.as_deref() {
+			qs.push("q", Some(q));
+		}
```
kkantan, line 80: I think this might be slightly better to avoid unwrapping and rewrapping the optional.
```suggestion
		if query.is_some() {
			qs.push("q", q.as_deref());
		}
```

kkantan (changes requested):
I left a few style comments that may be useful for future reference, but aren't necessary to change. the source appears to work well.

there's one thing that might need to be addressed before merging: do you know if the keys changed due to the site rewrite? I would assume so, since the previous keys were url slugs, but now they're keys from the json api. if you're sure the keys are the same, we can ignore this and merge, but if not, it'd be best to set the `breakingChangeVersion` to `2` in the `source.json` config to prompt users to migrate existing library titles.

neldra:
> I left a few style comments that may be useful for future reference, but aren't necessary to change. the source appears to work well.
> 
> there's one thing that might need to be addressed before merging: do you know if the keys changed due to the site rewrite? I would assume so, since the previous keys were url slugs, but now they're keys from the json api. if you're sure the keys are the same, we can ignore this and merge, but if not, it'd be best to set the `breakingChangeVersion` to `2` in the `source.json` config to prompt users to migrate existing library titles.

The keys did change, the old source keyed titles by the URL slug path, the rewrite keys them by the API id, so existing library entries won't match. Updated the PR with `breakingChangeVersion: 2` as you suggested. Going to leave the other comments for now but will keep them in mind for future PRs thanks!

sources/en.novelbuddy/res/source.json
```diff
@@ -2,9 +2,25 @@
 	"info": {
 		"id": "en.novelbuddy",
 		"name": "NovelBuddy",
-		"version": 1,
+		"version": 2,
+		"breakingChangeVersion": 2,
```
kkantan: this should be inside the `config` object:
```json
{
	"info": { ... },
	"config": {
		"breakingChangeVersion": 2
	},
	"listings": [ ... ]
}

kkantan (changes requested):

kkantan (approved):

PR #527, opened 2026-05-19
feat: add zh.rumanhua

### LINKED ISSUE #186: rumanhua1 {Chinese source request}
http://www.rumanhua1.com/

mirror: http://m.rumanhua1.com/

686udjie:
used https://m.rumanhua.org/ instead of https://www.rumanhua.org/ because search + filtering actually worked on the mobile site and didn’t on the main one

closes #186


ghost:
Forgot to put, closes #186 
Later on in the future I’ll find out a way to get it to work in templates/

sources/zh.rumanhua/res/source.json
```diff
@@ -0,0 +1,16 @@
+{
+	"info": {
+		"id": "zh.rumanhua",
+		"name": "入漫画",
+		"version": 1,
+		"url": "https://m.rumanhua.org/",
```
kkantan: I think the non-mobile website should also be provided so deep links all work, especially for devices like ipads and macs.
```suggestion
		"urls": [
			"https://m.rumanhua.org",
			"https://www.rumanhua.org"
		],
```

sources/zh.rumanhua/res/source.json
```diff
@@ -0,0 +1,16 @@
...
+		"contentRating": 0,
+		"languages": [
+			"zh"
+		]
+	},
+	"config": {
+		"hidesFiltersWhileSearching": true,
+		"supportsAuthorSearch": false
+	}
+}
```
kkantan:
```suggestion
}

```

sources/zh.rumanhua/src/html/details.rs
```diff
@@ -0,0 +1,90 @@
...
+	fn update_details(&self, manga: &mut Manga) -> Result<()>;
+	fn get_chapters(&self) -> Result<Vec<Chapter>>;
+}
+
+impl RumanhuaDetailsHtml for Document {
+	fn update_details(&self, manga: &mut Manga) -> Result<()> {
+		if let Some(img) = self.select_first("div.detailTop div.content img.cover") {
+			let cover = img.attr("src").unwrap_or_default();
+			let title = img.attr("alt").unwrap_or_default();
+			if !cover.is_empty() {
```
kkantan: you're allocating empty strings in the case of `None`s and then checking if they're empty, instead of just checking the optional. I don't believe there is a case where the string is `Some` and empty, so `unwrap_or_default` is unnecessary.

sources/zh.rumanhua/src/html/list.rs
```diff
@@ -0,0 +1,75 @@
...
+			let mut title = String::new();
+			let mut cover_src = String::new();
+
+			if let Some(a) = node.select_first("a") {
+				href = a.attr("href").unwrap_or_else(String::new);
+			}
+
+			if href.is_empty() {
+				continue;
+			}
```
kkantan: again, try to avoid allocating empty strings for no reason.
```rust
let href = node.select_first("a").and_then(|a| a.attr("href");
let Some(href) = href else {
	continue;
};
```

sources/zh.rumanhua/src/json/decrypt.rs
```diff
@@ -0,0 +1,23 @@
...
+use cbc::cipher::{BlockDecryptMut, KeyIvInit};
+
+pub fn decrypt_params(encrypted_str: &str) -> Option<String> {
+	let decoded = STANDARD.decode(encrypted_str.trim()).ok()?;
+	if decoded.len() < 16 {
+		return None;
+	}
+	let iv = &decoded[0..16];
+	let ciphertext = &decoded[16..];
+	let key = b"9S8$vJnU2ANeSRoF";
```
kkantan: this might make more sense defined as a `const KEY` outside of the function.

sources/zh.rumanhua/src/net/url.rs
```diff
@@ -0,0 +1,107 @@
...
+		} else if found_digit {
+			break;
+		}
+	}
+	num_str.parse::<f32>().ok()
+}
+
+pub fn get_search_url(query: Option<String>, page: i32, filters: Vec<FilterValue>) -> String {
+	if let Some(q) = query
+		&& !q.is_empty()
```
kkantan: `query` won't be empty if it's `Some`.

sources/zh.rumanhua/src/net/url.rs
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	let mut leaderboard_id = String::new();
+	let mut status_id = String::new();
+	let mut audience_id = String::new();
+
+	for filter in filters {
+		if let FilterValue::Select { id, value } = filter {
+			if id == "leaderboard" {
+				leaderboard_id = value;
```
kkantan: there's no reason to continue the loop if you only use one id to determine the url.

being able to configure multiple filters when it only supports one filter at a time is a little unnatural, but I'm not sure if there's a better solution.

sources/zh.rumanhua/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+		let html_str = get_request(&url)?.string()?;
+
+		let images = parse_images_from_html(&html_str)?;
+		let mut pages = Vec::new();
+		for img_url in images {
+			pages.push(Page {
+				content: PageContent::url(img_url),
+				..Default::default()
+			});
+		}
```
kkantan: you could just collect into a new vec rather than pushing to a mutable one.
```suggestion
		let pages = images
			.into_iter()
			.map(|img_url| {
				Page {
					content: PageContent::url(img_url),
					..Default::default()
				}
			})
			.collect();
```

sources/zh.rumanhua/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+crate-type = ["cdylib"]
+
+[profile.dev]
+panic = "abort"
+
+[profile.release]
+panic = "abort"
+opt-level = "s"
+strip = true
+lto = true
```
kkantan:
```suggestion
lto = true

```

sources/zh.rumanhua/src/net/url.rs
```diff
@@ -0,0 +1,107 @@
...
+
+pub fn extract_chapter_key(url: &str) -> Option<String> {
+	let clean = url.trim_end_matches('/');
+	let segment = clean.split('/').next_back()?;
+	if segment.ends_with(".html") {
+		Some(segment.to_string())
+	} else {
+		None
+	}
+}
```
kkantan, line 36: it seems like your keys are including the base url, which using isn't a good practice if it ends up changing in the future. could you use the url path instead?

kkantan (changes requested):
it would be better if this was a template instead, like the keiyoushi source, but it's okay if that's done in the future I guess.

ghost:
I addressed all the issues, but could you take a closer look at `src/html/lists.rs`? I think the refactor to using `and_then()` is a bit wonky.

Thanks for all the code reviews :happy_teto:

sources/zh.rumanhua/src/html/details.rs
```diff
@@ -0,0 +1,87 @@
...
+				manga.cover = Some(get_absolute_url(&cover));
+			}
+			if let Some(title) = img.attr("alt") {
+				manga.title = title;
+			}
+		}
+
+		if let Some(p) = self.select_first("div.detailContent p")
+			&& let Some(desc) = p.text()
+		{
```
kkantan: I think avoiding let chains is a little simpler:
```suggestion
		if let Some(desc) = self.select_first("div.detailContent p").and_then(|p| p.text()) {
```

sources/zh.rumanhua/src/html/details.rs
```diff
@@ -0,0 +1,87 @@
...
+		manga.status = status;
+		manga.url = Some(get_absolute_url(&format!("/news/{}", manga.key)));
+		Ok(())
+	}
+
+	fn get_chapters(&self) -> Result<Vec<Chapter>> {
+		let mut chapters = Vec::new();
+
+		if let Some(elements) = self.select("ul.chapterList li a") {
+			for a in elements {
```
kkantan: I think you can just reverse the iterator, which should be cheaper than reversing an already populated vector.
```suggestion
			for a in elements.rev() {
```

sources/zh.rumanhua/src/html/list.rs
```diff
@@ -0,0 +1,56 @@
...
+};
+
+pub fn parse_manga_list(html: &Document) -> Result<MangaPageResult> {
+	let mut manga = Vec::new();
+	if let Some(elements) = html.select("div.item, ul.rankList li") {
+		for node in elements {
+			let href = node.select_first("a").and_then(|a| a.attr("href"));
+			let Some(href) = href else {
+				continue;
+			};
```
kkantan: instead of using a simple for loop and `continue` statements, you could alternatively collect from an iterator. this lets you use the `?` operator inside a `filter_map`.

for example:
https://github.com/Aidoku-Community/sources/blob/29f1762c0df1936ff59e20a0b3744c03bfb93945/sources/en.asurascans/src/lib.rs#L81-L97

sources/zh.rumanhua/src/html/list.rs
```diff
@@ -0,0 +1,56 @@
...
+
+	let has_more = html.select("a")
+		.map(|mut elements| {
+			elements.any(|a| {
+				a.text()
+					.map(|text| text.contains("下一页") || text.contains("下页"))
+					.unwrap_or(false)
+			})
+		})
+		.unwrap_or(false);
```
kkantan: I'm not entirely certain, but you may be able to use the `:contains` operator to simplify this. iterating over and fetching the text of every `a` element seems like it may be a bit expensive.

for example:
https://github.com/Aidoku-Community/sources/blob/29f1762c0df1936ff59e20a0b3744c03bfb93945/sources/en.asurascans/src/lib.rs#L99-L101

sources/zh.rumanhua/src/net/url.rs
```diff
@@ -0,0 +1,107 @@
...
+		return if page <= 1 {
+			format!("{}/search/{}", BASE_URL, encoded)
+		} else {
+			format!("{}/search/{}/{}", BASE_URL, encoded, page)
+		};
+	}
+
+	let mut leaderboard_id = String::new();
+	let mut status_id = String::new();
+	let mut audience_id = String::new();
```
kkantan: I still don't think there's any reason to allocate these empty strings when they aren't used in every case. you should just return the url from inside the filter loop, but if not, at the very least you should use `Option<String>` instead.

sources/zh.rumanhua/src/html/details.rs
```diff
@@ -0,0 +1,87 @@
...
+				let Some(url) = a.attr("href") else { continue; };
+				let Some(title) = a.text() else { continue; };
+				let Some(key) = extract_chapter_key(&url) else {
+					continue;
+				};
+				let chapter_num = extract_chapter_number(&title).unwrap_or(0.0);
```
kkantan: does setting a chapter number of 0 by default make sense? it seems there are a fair few chapters with no numbers on the website.

kkantan (changes requested):
just left a few suggestions for improvements, but source appears to work well otherwise.

ghost:
should pass all qol checks i think

sources/zh.rumanhua/src/html/details.rs
```diff
@@ -0,0 +1,81 @@
...
+				manga.cover = Some(get_absolute_url(&cover));
+			}
+			if let Some(title) = img.attr("alt") {
+				manga.title = title;
+			}
+		}
+
+		if let Some(desc) = self.select_first("div.detailContent p").and_then(|p| p.text()) {
+			manga.description = Some(desc);
+		}
```
kkantan: my mistake with the previous comment, but there's no reason to unwrap and rewrap the optional here:
```suggestion
		manga.description = self.select_first("div.detailContent p").and_then(|p| p.text());
```

sources/zh.rumanhua/src/html/details.rs
```diff
@@ -0,0 +1,81 @@
...
+
+			chapters.push(Chapter {
+				key,
+				title: Some(title),
+				chapter_number: Some(chapter_num),
+				url: Some(get_absolute_url(&url)),
+				..Default::default()
+			});
+		}
+	}
```
kkantan: the indentation here looks a little off. make sure to run `cargo fmt`.

kkantan (changes requested):
sorry to block merge one last time. just running `cargo fmt` should be good.

ghost:
Should be good now. Thanks for the reviews and for all that you do. I’ll make sure future contributions are less of a hassle. Thanks for maintaining all these sources. Goodnight :3

kkantan (approved):

PR #532, opened 2026-05-23
fix(en.comix): handle scrambled page images

### LINKED ISSUE #533: Comix.to: some page images are tile-scrambled in reader
Some individual pages in Comix chapters display as scrambled rectangular/vertical tile chunks in Aidoku.

This is not chapter page order being wrong. The affected page image itself appears broken/tile-scrambled, with pieces of the same page arranged in the wrong positions and large blank gaps.

The same pages display correctly in the Comix website reader.

Other Aidoku sources I tested do not show this issue.

It looks like Comix may be serving raw obfuscated/tile-scrambled images and the website reader may be descrambling/reassembling them client-side. If that is true, the source may need image descrambling support, or Aidoku may need an image-processing hook if sources cannot currently rewrite image bytes.

### Screenshots

Image

Image

jianmingyong:
# Summary
This PR attempt to solve existing comix issues by adding fallback path if installer fails.

Installer will now do the following:
- Check for `vm` objects and probe each key for existing functions
- If fails, load the ES module into `vm` object and try searching again
- If all else fails, it should return an error on the user end

This new installer ensures that it will survive the current changes as they keep rotating between them every day.

After a few changes, comix now decided to scramble images. This PR will also handle scrambled images via web_view by loading a custom canvas and dump the image data out. The raw image is PNG in base64 format.

# Scrambled Images
Some Comix pages are marked with `s: 1` and use scrambled `/si/` or `/sii/` image routes.

- Normal page remain unchanged.
- If they are marked as scrambled, it would use the descrambler to get the original image.

The reason why this PR isn't changing the URL is because of the potential breaking image routes. As we figure out, pages marked with `s: 1` will return 404 errors if you do not have the required request header. This is enforced properly if you use the descrambler function. Conversely, if you were to add custom request header into non scrambled routes, you will get 404 as well. This is really annoying, and you might end up getting no image at all if you try changing URL and testing them.

So the best solution is, keep the image route as is, let the descrambler handle them. They can't break this unless they change the requirement.

For every scrambled image, the output result would convert the image into PNG file in base64 format. This is a limitation set by IOS devices since web_view internally uses the device browser and all browser uses the same web kit engine anyway.

Then again PNG is a lossless format so it will not make the image any less blurry than before. Since it is just a slightly bigger size image, that shouldn't be much issue.

# PR Checklist
- [x] cargo fmt has been run before submission.
- [x] cargo clippy outputs no lint warnings.

### Notes
This is just a proof of concept that I made that I am testing which may help if they ever decided to move the code to ES module making it impossible to find via window object. It happens once and who knows if they ever rotate back to that again.

Making this as a draft as the code is probably not the cleanest thing and need testing. (Sorry, my rust knowledge is not that great)
There is no way for me to handle async function properly without ugly hacks like the while loop.

comix had yet to change anything with the axios interceptors for a while now. They are always exposed in the module. We are now using the interceptors to handle both signing the request and decoding the response.

Feel free to comment if you have any other better idea while we wait and see what comix would do.
(Maybe we won't even need this as comix loves changing stuff that it make this obsolete)

### DIFF
--- a/sources/en.comix/Cargo.toml
+++ b/sources/en.comix/Cargo.toml
@@ -10,6 +10,7 @@ crate-type = ["cdylib"]
 aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", features = ["json"] }
 serde = { version = "1.0", features = ["derive", "alloc"], default-features = false }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+regex = { version = "1.12.3", default-features = false }
 
 [dev-dependencies]
 aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", features = ["test"] }
--- a/sources/en.comix/Cargo.toml
+++ b/sources/en.comix/Cargo.toml
@@ -10,8 +10,8 @@ crate-type = ["cdylib"]
 aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", features = ["json"] }
 serde = { version = "1.0", features = ["derive", "alloc"], default-features = false }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
-regex = { version = "1.12.3", default-features = false }
-base64 = { version = "0.22.1", default-features = false, features = ["alloc"] }
+regex = { version = "1.12", default-features = false }
+base64 = { version = "0.22", default-features = false, features = ["alloc"] }
 
 [dev-dependencies]
 aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", features = ["test"] }
--- a/sources/en.comix/Cargo.toml
+++ b/sources/en.comix/Cargo.toml
@@ -11,6 +11,7 @@ aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", features = ["json"] 
 serde = { version = "1.0", features = ["derive", "alloc"], default-features = false }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
 regex = { version = "1.12.3", default-features = false }
+base64 = { version = "0.22.1", default-features = false, features = ["alloc"] }
 
 [dev-dependencies]
 aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", features = ["test"] }
--- a/sources/en.comix/src/lib.rs
+++ b/sources/en.comix/src/lib.rs
@@ -285,7 +285,13 @@ impl Source for Comix {
 					format!("{base_url}/{}", page.url.trim_start_matches('/'))
 				};
 				Page {
-					content: PageContent::url(url),
+					content: if let Some(s) = page.s {
+						let mut context = PageContext::new();
+						context.insert("s".into(), s.to_string());
+						PageContent::url_context(url, context)
+					} else {
+						PageContent::url(url)
+					},
 					..Default::default()
 				}
 			})
@@ -501,57 +507,31 @@ impl PageImageProcessor for Comix {
 	fn process_page_image(
 		&self,
 		response: ImageResponse,
-		_context: Option<PageContext>,
+		context: Option<PageContext>,
 	) -> Result<ImageRef> {
-		let Some(url) = response.request.url else {
-			bail!("Unable to get the image url")
-		};
+		if context.is_some_and(|ctx| ctx.get("s").is_some_and(|s| s == "1")) {
+			let Some(url) = response.request.url else {
+				bail!("Unable to get the image url")
+			};
 
-		let mut is_scrambled = false;
+			let web_view = web::create_web_view()?;
+			let data_url = web::descramble_image(
+				&web_view,
+				response.image.width(),
+				response.image.height(),
+				url.as_ref(),
+			)?;
+			let Some((_, base64_data)) = data_url.split_once(',') else {
+				bail!("Unable to get the raw image data")
+			};
+			let bytes: Vec<u8> = general_purpose::STANDARD
+				.decode(base64_data)
+				.or_else(|_| bail!("Invalid base64 data given"))?;
 
-		if !response.headers.contains_key("x-scramble-grid")
-			|| !response.headers.contains_key("x-scramble-seed")
-		{
-			// There is a bug with the current response header that can be empty when there should be something.
-			// This might not be needed in the future but for now, we need it to ensure correct response header.
-			let image_response = Request::head(&url)?.send()?;
-			if image_response.get_header("x-scramble-grid").is_some()
-				&& image_response
-					.get_header("x-scramble-seed")
-					.is_some_and(|seed| seed != "0")
-			{
-				is_scrambled = true;
-			}
+			Ok(ImageRef::new(bytes.as_ref()))
 		} else {
-			// If both exists, we can check if the seed is scrambled or not.
-			if response
-				.headers
-				.get("x-scramble-seed")
-				.is_some_and(|seed| seed != "0")
-			{
-				is_scrambled = true;
-			}
+			Ok(response.image)
 		}
-
-		if !is_scrambled {
-			return Ok(response.image);
-		}
-
-		let web_view = web::create_web_view()?;
-		let data_url = web::descramble_image(
-			&web_view,
-			response.image.width(),
-			response.image.height(),
-			url.as_ref(),
-		)?;
-		let Some((_, base64_data)) = data_url.split_once(',') else {
-			bail!("Unable to get the raw image data")
-		};
-		let bytes: Vec<u8> = general_purpose::STANDARD
-			.decode(base64_data)
-			.or_else(|_| bail!("Invalid base64 data given"))?;
-
-		Ok(ImageRef::new(bytes.as_ref()))
 	}
 }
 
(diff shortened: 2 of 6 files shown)


jianmingyong:
Patch a bug where it kinda fails in a try {} catch {} loading module. It is ugly but too bad, we can't use await as this JavaScript eval doesn't support it.

Smexhy:
I tested a similar module-based approach locally and hit the same async setup limitation. What worked reliably for me was to have the WebView-side module setup expose a small Aidoku-owned JSON state on `window` (`pending`, `ready`, or `failed`), then poll that state from Rust with bounded waits. It is still polling, but it avoids a tight indefinite loop and preserves the actual install failure stage when something breaks.

One related finding for the image handling in #533: rewriting scrambled `/si/...` or `/sii/...` URLs to `/i/...` works for some older affected pages, but not consistently for newer chapters because the clean `/i/...` counterpart can be unavailable. In those cases I found the module-based reader/security helper useful for obtaining the tile mapping so Aidoku can reassemble the scrambled image locally.

I have chapter/page loading and affected scrambled pages working in local on-device testing with that approach, so I wanted to share the findings in case they are useful here.

jianmingyong:
I found a way to make it work but yea it is SUPER SLOW. Also probably because of my bad rust code as I have the idea, but I can't write it properly.

jianmingyong:
I end up using two more dependency... regex really is just to get the module string quickly, base64 is to convert the base64 image output into bytes.

regex = { version = "1.12.3", default-features = false }
base64 = { version = "0.22.1", default-features = false, features = ["alloc"] }

There are still some flaws when I tried to check the header for scrambled image or not so I had no choice but to treat every single image as scrambled... Still finding ways to detect properly so we can skip non scrambled images.

Idk if that is Aidoku bug but every single image response header is empty -_- but I am sure there are something when loaded in browser though... welp there is nothing I could do if that is the case.

I don't want to just assuming N*3 chapter being scrambled as they could just change the number. Deal with it, slow or nothing.

sources/en.comix/src/lib.rs
```diff
@@ -495,15 +498,31 @@ impl ListingProvider for Comix {
...
+	) -> Result<ImageRef> {
+		let web_view = web::create_web_view()?;
+		let data_url = web::descramble_image(
+			&web_view,
+			response.image.width(),
+			response.image.height(),
+			response.request.url.unwrap().as_ref(),
+		)?;
+		let base64_data = data_url.split_once(',').map(|(_, data)| data).unwrap();
+		let bytes: Vec<u8> = general_purpose::STANDARD.decode(base64_data).unwrap();
```
kkantan: you should use `if let Some(...) = ... else { ... }` or `.ok_or(error!("..."))?`, instead of unwrapping optionals.

sources/en.comix/src/models.rs
```diff
@@ -239,14 +240,15 @@ pub struct ComixChapter {
...
+	pub url: String,
 }
 
 impl ComixChapter {
 	pub fn created_at(&self) -> i64 {
 		helpers::parse_relative_date_string(&self.created_at_formatted)
 	}
 
-	pub fn into_chapter(self, manga_id: &str) -> Chapter {
+	pub fn into_chapter(self) -> Chapter {
```
kkantan: since this doesn't take a parameter anymore it could be a `impl From<ComixChapter> for Chapter` now like the manga struct.

sources/en.comix/src/web.rs
```diff
@@ -14,55 +15,121 @@ if (!vmObj || typeof vmObj !== 'object' || vmObj === window) {\
...
 	Ok(comix_web_view)
 }
 
+fn find_secure_module_src(web_view: &mut ComixWebView) -> Result<()> {
+	let main_module_src = Request::get(BASE_URL)?
+		.html()?
+		.select("head > script[type=\"module\"][src*=\"main\"]")
+		.and_then(|e| e.first())
+		.and_then(|e| e.attr("src"))
+		.ok_or(AidokuError::message("Main module not found"))?;
```
kkantan:
```suggestion
		.ok_or(error!("Main module not found"))?;
```

sources/en.comix/src/web.rs
```diff
@@ -14,55 +15,121 @@ if (!vmObj || typeof vmObj !== 'object' || vmObj === window) {\
...
+	if let Some(js_asset_path_index) = main_module_src.rfind("/") {
+		let js_asset_path = &main_module_src[0..js_asset_path_index + 1];
+		let secure_script_regex = Regex::new("(secure-[A-Za-z0-9-_]+?\\.js)").unwrap();
+		let main_module_contents =
+			Request::get(format!("{BASE_URL}{main_module_src}"))?.string()?;
+		if let Some(captures) = secure_script_regex.captures(main_module_contents.as_str()) {
+			let secure_script_path = captures.get(1).unwrap().as_str();
```
kkantan: you should avoid the `unwrap()` here, probably with `and_then` on the `.captures(...)` so you can reuse the if let.

kkantan (changes requested):
the source no longer loads for me in aidoku. is this a new change or just an issue with my network?

jianmingyong:
> the source no longer loads for me in aidoku. is this a new change or just an issue with my network?

Going to love the cat and mouse game with comix. Every day the script would rotate between normal JS and ESM module and break the old implementation.

sources/en.comix/src/web.rs
jianmingyong: Done

sources/en.comix/src/lib.rs
jianmingyong: Done

sources/en.comix/src/web.rs
jianmingyong: Done

sources/en.comix/src/models.rs
jianmingyong: Done

jianmingyong:
Comix is doing weird things again… but it sounds like a comix side issue rather than the extension itself. Some image is loading 404 errors but then it seems to be loading fine in comix until I check the network tab to see it as 404 as well.

Edit:
Seems like it kind of affects the comix website too in a weird way that makes the website crash after a short while. Lol, so it is not another breaking change. Server side issues.

sources/en.comix/Cargo.toml
```diff
@@ -10,6 +10,8 @@ crate-type = ["cdylib"]
 aidoku = { git = "https://github.com/Aidoku/aidoku-rs.git", features = ["json"] }
 serde = { version = "1.0", features = ["derive", "alloc"], default-features = false }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+regex = { version = "1.12.3", default-features = false }
+base64 = { version = "0.22.1", default-features = false, features = ["alloc"] }
```
kkantan: usually we ignore the patch version so that it's easier to update packages in the future, as typically there aren't any breaking changes.
```suggestion
regex = { version = "1.12", default-features = false }
base64 = { version = "0.22", default-features = false, features = ["alloc"] }
```

kkantan (changes requested):
I think overall these changes are fine, but after testing on device, it's true that page loading time is significantly increased.

there are two things that could be taken from the other pr to make this better: the shared web view, and the processing skip for `/i/` urls. maybe the HEAD requests for page urls to see if they can just be rewritten to `/i/` too.

jianmingyong:
> I think overall these changes are fine, but after testing on device, it's true that page loading time is significantly increased.
> 
> there are two things that could be taken from the other pr to make this better: the shared web view, and the processing skip for `/i/` urls. maybe the HEAD requests for page urls to see if they can just be rewritten to `/i/` too.

The shared webview can be looked into. The process skip initially was implemented in a different way without additional request but then Aidoku had a bug with response header which effectively prevented me from using ProcessPageImage response. (So I didn’t commit that)

Having to add another request to check the header is the only way right now. oh well. I will look into adding that.

jianmingyong:
Now all unscrambled images had a fast path of returning original image without any other modifications. No need for additional request or changing the URL.

In fact for some reason, if you were to use head request to check stuff, some images return 404. (I checked with my browser and yea it is returning 404 too) Likely they want you to use their custom request header, but I have no idea what that is, so I just throw all of these scrambled ones into their JS function to handle it properly.

Update:
From #538, we have learned that if the request header does not contain origin URL, the scrambled image will return 404. Since I fed the URL into their JS function, I have effectively solved the issue. However, if you do the same with non-scrambled image, they return 404 unless you remove the origin URL.

So, if the page contains `s = 1`, you need to add custom request header. Otherwise, you must not have custom request header, or you get no image. Though maybe `/i/` path isn't that strict, but I doubt they are going to work if they do the same. (It doesn't matter if the scrambled path had seed = 0, as long as they are marked with `s=1` you must use their custom request header)

Seriously, in any case, using the canvas and get the image data seems to be easier without any means of breaking unless they decide to add new form of obfuscation.

Update 2:

Some images will still have 404 errors despite running the descrambled function. However from my testing, simply close the chapter and reopen again would fix it. So there is likely some issue loading the images.

kkantan (approved):
I'm going to merge this for now since the changes appear to be good enough to at least be more reliable for users until another site change. the next best improvement would probably be a single shared web view instance for the source, which might also help avoid the async execution workaround somewhat.

PR #538, opened 2026-05-25
fix(en.comix): update reader flow and scrambled image handling

### LINKED ISSUE #533: Comix.to: some page images are tile-scrambled in reader

Smexhy:
## Summary

Updates the Comix source for current site behavior:

- uses the module-based reader/security helper flow for chapter-list and chapter-page requests;
- supports scrambled page images reported in #533;
- handles search results where `poster` is `null`;
- omits result entries without a usable title route key.

This overlaps with #532, but differs in image handling: normal pages are untouched, only pages marked `s:1` can enter descrambling, and image processing uses prepared per-page context rather than creating or using a WebView while rendering each image.

## Reader Flow

The previous source depended on helper functions exposed through a `window.vm*` object. Current Comix reader behavior exposes the required signing, response decoding, and scramble-map functionality through module-based frontend code.

Chapter-list and chapter-page requests now use that helper path directly. Because module initialization is asynchronous inside the WebView, the source stores a small JSON install state and polls it with bounded waits rather than expecting `eval` to await a Promise.

Plain JSON response handling remains supported where returned by the API.

## Scrambled Images

Some Comix pages are marked with `s: 1` and use scrambled `/si/...` or `/sii/...` image routes.

- Normal pages are returned unchanged.
- Each scrambled-marked page checks its own `/i/...` counterpart and uses the clean route as a fast path when it is available.
- The clean-route check remains page-specific because Comix can expose `/i/...` for some marked pages while neighboring marked pages in the same chapter still require `/si/...` processing.
- When `/i/...` is unavailable, the source prepares per-page tile-map context through the reader/security helper and reassembles only affected images through Aidoku's `PageImageProcessor`.
- The processor performs local canvas work only; it does not call the WebView or helper asynchronously while rendering images.
- If descrambling cannot be prepared or processing fails, the raw image is returned instead of failing the entire chapter.
- Unknown future scrambled route names are not guessed as clean `/i/...` routes.

The image-specific implementation is isolated in `images.rs` to keep the main source flow readable.

## Response Parsing

Comix search results can contain `poster: null`. These results now parse successfully and display without a cover rather than failing the result list.

Comix can also return entries without a usable title route key. Those entries are omitted rather than producing an error when opened.

No search ranking changes are intended.

## Validation

Passed:

- `cargo fmt --check`
- `cargo test`
- `cargo check`
- `cargo clippy -- -D warnings`
- `aidoku package`
- `aidoku verify package.aix`
- `git diff --check`

On-device testing passed for normal chapters, recent chapters requiring descrambling, older affected chapters using the clean `/i/` path, chapters with mixed per-page clean-route availability, nullable-poster search results, Preload Pages, and valid titles with no uploaded chapters.

## Retest Checklist

- Open a normal older chapter.
- Open a recently uploaded chapter containing scrambled pages.
- Verify affected scrambled pages from #533 render correctly.
- Verify a chapter with mixed clean-route availability does not show blank pages.
- Verify reading with Preload Pages enabled and disabled.
- Search for a result set containing a title without a poster, such as `friends`, and confirm results display.
- Open a valid series with no uploaded chapters and confirm it loads normally.

## Remaining Risk

This source follows Comix's current module, API, and image-scrambling behavior. Further site changes may require source adjustments.

### DIFF
--- a/sources/en.comix/src/images.rs
+++ b/sources/en.comix/src/images.rs
@@ -1,0 +1,535 @@
+use crate::{BASE_URL, Comix, models::ComixPage, web};
+use aidoku::{
+	ImageResponse, Page, PageContent, PageContext, PageImageProcessor, Result,
+	alloc::{String, Vec, string::ToString, vec},
+	imports::{
+		canvas::{Canvas, ImageRef, Rect},
+		net::{Request, Response},
+	},
+	prelude::*,
+};
+
+const SCRAMBLED_PAGE_GRID_CONTEXT_KEY: &str = "comixScrambleGrid";
+const SCRAMBLED_PAGE_MAP_CONTEXT_KEY: &str = "comixScrambleMap";
+pub const PAGE_S_FLAG_CONTEXT_KEY: &str = "comixPageS";
+const PAGE_WIDTH_CONTEXT_KEY: &str = "comixPageWidth";
+const PAGE_HEIGHT_CONTEXT_KEY: &str = "comixPageHeight";
+
+/// Builds the page list, batching the s:1 clean-route and scramble-metadata
+/// probes through `send_all` so latency scales with round-trips, not page count.
+pub fn build_page_list(source: &Comix, base_url: &str, items: Vec<ComixPage>) -> Result<Vec<Page>> {
+	let resolved: Vec<ResolvedPage> = items
+		.into_iter()
+		.enumerate()
+		.map(|(index, page)| ResolvedPage::new(base_url, page, index))
+		.collect::<Result<Vec<_>>>()?;
+
+	let clean_available = probe_clean_routes(&resolved);
+	let scramble_headers = probe_scramble_headers(&resolved, &clean_available);
+
+	let pages = resolved
+		.into_iter()
+		.enumerate()
+		.map(|(i, page)| Page {
+			content: page.into_content(
+				source,
+				clean_available.get(&i).copied().unwrap_or(false),
+				scramble_headers.get(&i),
+			),
+			..Default::default()
+		})
+		.collect();
+	Ok(pages)
+}
+
+struct ResolvedPage {
+	url: String,
+	index: usize,
+	s_flag: bool,
+	scrambled_bounds: Option<(usize, usize)>,
+	width: f32,
+	height: f32,
+}
+
+impl ResolvedPage {
+	fn new(base_url: &str, page: ComixPage, index: usize) -> Result<Self> {
+		let (width, height) = (page.width, page.height);
+		let url = absolute_page_url(base_url, page.url)?;
+		let s_flag = page.s == Some(1);
+		let scrambled_bounds = s_flag.then(|| scrambled_route_bounds(&url)).flatten();
+		Ok(Self {
+			url,
+			index,
+			s_flag,
+			scrambled_bounds,
+			width,
+			height,
+		})
+	}
+
+	fn clean_probe(&self) -> Option<(usize, String)> {
+		self.scrambled_bounds
+			.map(|(start, end)| (self.index, clean_image_url(&self.url, start, end)))
+	}
+
+	fn into_content(
+		self,
+		source: &Comix,
+		clean_available: bool,
+		scramble: Option<&PageScramble>,
+	) -> PageContent {
+		if !self.s_flag {
+			return PageContent::url(self.url);
+		}
+		let (width, height) = (self.width, self.height);
+		let Some((start, end)) = self.scrambled_bounds else {
+			// s:1 but the URL doesn't match a known scrambled route — still
+			// carry the s:1 context so get_image_request adds Origin.
+			return PageContent::url_context(self.url, with_dims(s_flag_context(), width, height));
+		};
+		if clean_available {
+			return PageContent::url_context(
+				clean_image_url(&self.url, start, end),
+				with_dims(s_flag_context(), width, height),
+			);
+		}
+		// Build the descramble context locally so image processing at render
+		// time doesn't need to touch the WebView.
+		let context = scramble
+			.and_then(|s| source.prepare_scrambled_page_context(&s.seed, &s.grid))
+			.unwrap_or_else(s_flag_context);
+		// When the seed/grid came from a cache-busted probe, fetch the image from
+		// the same busted URL so its bytes match the seed the map was built from.
+		let url = if scramble.is_some_and(|s| s.cache_bust) {
+			let separator = if self.url.contains('?') { '&' } else { '?' };
+			format!("{}{separator}r=1", self.url)
+		} else {
+			self.url
+		};
+		PageContent::url_context(url, with_dims(context, width, height))
+	}
+}
+
+fn probe_clean_routes(resolved: &[ResolvedPage]) -> aidoku::HashMap<usize, bool> {
+	let requests: Vec<(usize, Request)> = resolved
+		.iter()
+		.filter_map(|p| {
+			let (index, clean_url) = p.clean_probe()?;
+			Some((index, apply_s1_headers(Request::head(&clean_url).ok()?)))
+		})
+		.collect();
+	batch_responses(requests)
+		.into_iter()
+		.map(|(index, response)| (index, is_image_response(response.as_ref())))
+		.collect()
+}
+
+enum ScrambleProbe {
+	Valid(String, String),
+	/// No usable seed/grid (e.g. the CDN's `X-Scramble-Seed: 0` response); the
+	/// caller retries with a cache-bust, then falls through to the raw image.
+	Unknown,
+}
+
+struct PageScramble {
+	seed: String,
+	grid: String,
+	/// The plain URL returned a degraded `seed:0` response and the real seed/grid
+	/// only came back from a cache-busted request, so the image must be fetched
+	/// from the busted URL too.
+	cache_bust: bool,
+}
+
+/// Fetches X-Scramble-Seed/Grid for s:1 pages without a clean `/i/` route via a
+/// HEAD batch. The CDN sometimes caches a degraded response (`seed:0`/no grid)
+/// for a page that is actually scrambled; a second HEAD with a cache-busting
+/// param forces a fresh response with the real seed/grid (the website recovers
+/// the same way). Pages still without seed/grid aren't scrambled and fall
+/// through to the raw image.
+fn probe_scramble_headers(
+	resolved: &[ResolvedPage],
+	clean_available: &aidoku::HashMap<usize, bool>,
+) -> aidoku::HashMap<usize, PageScramble> {
+	let targets: aidoku::HashMap<usize, &str> = resolved
+		.iter()
+		.filter(|p| {
+			p.s_flag
+				&& p.scrambled_bounds.is_some()
+				&& !clean_available.get(&p.index).copied().unwrap_or(false)
+		})
+		.map(|p| (p.index, p.url.as_str()))
+		.collect();
+	if targets.is_empty() {
+		return aidoku::HashMap::new();
+	}
+
+	let mut results = aidoku::HashMap::new();
+	let head_requests: Vec<(usize, Request)> = targets
+		.iter()
+		.filter_map(|(&i, &url)| Some((i, apply_s1_headers(Request::head(url).ok()?))))
+		.collect();
+	let mut unresolved = Vec::new();
+	for (index, response) in batch_responses(head_requests) {
+		match extract_scramble_probe(response) {
+			ScrambleProbe::Valid(seed, grid) => {
+				results.insert(
+					index,
+					PageScramble {
+						seed,
+						grid,
+						cache_bust: false,
+					},
+				);
+			}
+			ScrambleProbe::Unknown => unresolved.push(index),
+		}
+	}
+
+	let bust_requests: Vec<(usize, Request)> = unresolved
+		.into_iter()
+		.filter_map(|index| {
+			let url = *targets.get(&index)?;
+			let separator = if url.contains('?') { '&' } else { '?' };
+			let busted = format!("{url}{separator}r=1");
+			Some((index, apply_s1_headers(Request::head(&busted).ok()?)))
+		})
+		.collect();
+	for (index, response) in batch_responses(bust_requests) {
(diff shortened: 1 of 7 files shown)


sources/en.comix/src/lib.rs
```diff
@@ -26,12 +28,444 @@ const API_URL: &str = "https://comix.to/api/v1";
...
 // adult, boys love, ecchi, girls love, hentai, smut
 const NSFW_GENRE_IDS: &[&str] = &["87264", "8", "87265", "13", "87266", "87268"];
+const CLEAN_IMAGE_ROUTE_SEGMENT: &str = "i";
+const SCRAMBLED_PAGE_CONTEXT_KEY: &str = "comixScrambled";
+const SCRAMBLED_PAGE_GRID_CONTEXT_KEY: &str = "comixScrambleGrid";
+const SCRAMBLED_PAGE_MAP_CONTEXT_KEY: &str = "comixScrambleMap";
+const SCRAMBLED_PAGE_INDEX_CONTEXT_KEY: &str = "comixPageIndex";
 
-struct Comix;
+#[derive(Clone, Copy)]
```
kkantan: all these additions would be better refactored into a separate module rather than bloating the main `lib.rs`.

sources/en.comix/src/lib.rs
```diff
@@ -26,12 +28,444 @@ const API_URL: &str = "https://comix.to/api/v1";
...
+}
+
+fn clean_scrambled_image_route_available(url: &str) -> bool {
+	let clean_url = normalize_scrambled_page_url(url.to_string());
+
+	if clean_url == url {
+		return false;
+	}
+
+	match Request::head(clean_url)
```
kkantan: making a request for every image while loading the page list seems like a bad idea. does it cause any slowdowns?

sources/en.comix/src/lib.rs
```diff
@@ -26,12 +28,444 @@ const API_URL: &str = "https://comix.to/api/v1";
...
+	);
+	context
+}
+
+fn is_scrambled_page_context(context: Option<&PageContext>) -> bool {
+	context
+		.and_then(|context| context.get(SCRAMBLED_PAGE_CONTEXT_KEY))
+		.map(|value| value == "1")
+		.unwrap_or(false)
+}
```
kkantan: is there any point having this? it's always set to `"1"`.

kkantan (commented):
this seems a little overcomplicated for what it does, essentially achieving the same result as the other pr with 7x the additions?

sources/en.comix/src/lib.rs
Smexhy: Update: later testing showed the representative clean-route probe was not correct for mixed chapters, so the current branch restores per-page `/i/...` availability checks only for pages marked `s:1`. Normal pages still incur no probe or image processing.

This originally only probed pages marked `s:1`, not normal pages. I chose the per-page check because testing showed older chapters where `/i/...` existed and newer chapters where it returned `404`, so I did not want to assume clean-route availability was uniform within a chapter. It did not cause noticeable slowdown in my Aidoku testing. I temporarily reduced the request count, but testing then reproduced mixed clean-route availability within the same chapter, so restoring the page-specific check is needed for the current behavior.

sources/en.comix/src/lib.rs
Smexhy: Agreed. I moved the scrambled-image route selection, context preparation, canvas processing, fallbacks, and tests into a separate `images.rs` module. `lib.rs` now only delegates page construction to that module.

sources/en.comix/src/lib.rs
Smexhy: Agreed, this was redundant. The prepared grid/tile-map context already identifies pages that should go through image processing, so I removed the extra marker.

Smexhy:
Update: later testing showed the representative clean-route probe was not correct for mixed chapters, so the current branch restores per-page `/i/...` availability checks only for pages marked `s:1`. Normal pages still incur no probe or image processing. See the follow-up below for the reproduced regression.

Fair point, the first version was heavy to review.

There is overlap with #532, but this implementation is not quite the same path: normal images are untouched, only pages marked `s:1` can enter descrambling, and affected pages use the clean `/i/` route when it is available. For newer pages where `/i/` is unavailable, `PageImageProcessor` uses prepared per-page tile context and local canvas work; it does not create or call a WebView while Aidoku is rendering or preloading images. If setup or processing fails, it falls back to the raw image instead of failing the chapter.

The module-based helper code is also used for chapter/page signed requests and for preparing scramble tile maps, so using one helper path avoids maintaining separate `window.vm*` and module-based flows.

I cleaned up the branch in response to the review: the image logic is now in `images.rs` and the redundant context marker is removed. The original per-page clean-route check was there because I had observed older chapters with working `/i/` images and newer chapters where `/i/` returns `404`; it did not cause noticeable slowdown in my Aidoku testing. I temporarily reduced that probing, but testing then found mixed availability within a chapter, so the current branch uses the page-specific check for `s:1` pages to avoid returning unavailable clean images.

I have tested the revised build on-device with older clean-route chapters and newer chapters requiring descrambling, including Preload Pages. If there are any further parts that would be better simplified or aligned with existing source patterns, I am happy to adjust them.

jianmingyong:
The huge difference is probably the performance aspects. My PR (#532) create a new web_view per image scrambling which could be a little heavy depending on the logic. Though this one is super verbose in trying to get the thing done.

I want to optimize that as well, but I am keeping the things simple to get it over with before comix had time to make life harder.

Smexhy:
That is the tradeoff I was trying to avoid. This version prepares tile-map context before rendering and keeps PageImageProcessor local to each image, rather than creating or calling a WebView during image processing. That should avoid multiplying WebView work when Aidoku preloads affected pages, although I have not benchmarked the two approaches directly.

I agree it is a larger implementation. I have already moved the image-specific parts out of lib.rs and reduced the clean-route probing after the review. I would not call it future-proof, since it still follows Comix’s current internal reader flow; the remaining size is mainly for the current behavior, fallback handling, and avoiding WebView work in the image-processing path.

Smexhy:
Follow-up after additional testing: the representative clean-route probe introduced a correctness regression. Comix does not provide clean `/i/...` images uniformly within a chapter or even within the same scrambled image family; an individual `s:1` page may require `/si/...` plus processing while neighboring marked pages have working `/i/...` counterparts.

I therefore restored the clean-route availability check per `s:1` page. Normal pages still incur no probe or processing, and affected pages only use `PageImageProcessor` when their own clean route is unavailable.

This does reintroduce one request per scrambled-marked page, which was the original review concern, but the mixed availability case means that cost is required to avoid blank pages with the current Comix behavior. I tested the updated build locally against the case that exposed the issue, and the previously blank page now renders correctly.

sources/en.comix/src/images.rs
```diff
@@ -0,0 +1,486 @@
...
+	index: usize,
+) -> Result<PageContent> {
+	let is_scrambled = page.s == Some(1);
+	let url = absolute_page_url(base_url, page.url)?;
+	// Clean-route availability can differ per page within one chapter.
+	let use_clean_scrambled_route = is_scrambled && clean_image_route_available(&url);
+	let url = normalize_page_url(url, index, is_scrambled, use_clean_scrambled_route)?;
+
+	if is_scrambled && !has_clean_image_route(&url) {
+		if let Some(context) = source.prepare_scrambled_page_context(&url, index) {
```
kkantan: I'm unsure why this extra work needs to happen before the image processing, considering the url doesn't change. you could just put it in the image processing and it would speed up the initial page list loading, no?

kkantan: it looks like `comix_image_route` is called twice with this, which seems like a waste. overall, there seems to be a lot of overcomplicated processing here in the guise of being more modular. it would be better written by hand than ai I think, as the general logic isn't necessarily that complex.

sources/en.comix/src/models.rs
```diff
@@ -118,18 +122,31 @@ pub struct ComixManga {
...
+		alias = "content_rating",
+		default,
+		deserialize_with = "content_rating_or_unknown"
+	)]
 	pub content_rating: ComixContentRating,
 	pub authors: Option<Vec<Term>>,
 	pub artists: Option<Vec<Term>>,
 	pub genres: Option<Vec<Term>>,
 	pub tags: Option<Vec<Term>>,
+	#[serde(alias = "latest_chapter")]
```
kkantan: is it not always camelCase?

sources/en.comix/src/web.rs
kkantan, line 1: this is my main issue with the pr: I find it hard to believe that this file having 1k lines of code is actually necessary.

I think part of the reason is for extra generalization to maybe handle future site changes, but I don't think that should be a priority since we don't know what to actually expect for changes, and it makes maintaining the source in the future more difficult for others.

there's also lots of logging/error handling that doesn't seem useful enough to justify the added complexity.

sources/en.comix/src/web.rs
```diff
@@ -1,170 +1,1046 @@
...
+			const path = new URL(url).pathname;
+			if (/\/secure-[^/]+\.js$/.test(path)) score += 7;
+			if (EXPORT_RE.test(code)) score += 1;
+			if (code.includes("vmK_") && code.includes("globalThis")) score += 3;
+			if (code.includes("VM decryption key not available")) score += 2;
+			if (code.includes("interceptors")) score += 1;
+			if (code.includes(".request") && code.includes(".response")) score += 1;
+			if (code.includes("x-enc")) score += 1;
+			return score;
+		};
```
kkantan: is this just for future proofing? I don't understand why you need to fetch the content when you already know `secure-xxx.js` is most likely what you're looking for. I mean, it seems like this is only called when it finds a url that matches the `SECURE_CHUNK_RE`, so we already know that it's `secure-xxx.js`. testing the path once again is pointless, and this would only make sense if there are multiple `secure-xxx.js` files.

kkantan (changes requested):

sources/en.comix/src/images.rs
Smexhy: I did try doing more of the descramble work through the image-processing path initially, and in Aidoku those pages ended up black/empty or stuck loading. The version that worked reliably prepares the tile-map data before returning the pages and passes it through PageContext, leaving PageImageProcessor to perform local canvas reassembly only. That is why it is structured this way currently. I cannot say a different deferred implementation is impossible, only that the earlier one was not reliable in actual app testing.

sources/en.comix/src/models.rs
Smexhy: Yes, from the responses I have seen it is camelCase. I added those aliases while checking whether the malformed search result was caused by a response-shape variation, but the actual issue was only poster: null. They are not needed for the fix, so I will remove them.

sources/en.comix/src/web.rs
Smexhy: The entry-module fetch is needed to locate the current hashed secure-*.js import. I did run into a discovery break when the bundle used a more compact import form, so that part is required.

The additional content-scoring fallback was added defensively while the frontend bundle was changing, but I do not have a concrete case where it was needed once the secure-*.js import could be located. So yeah I agree that it adds unnecessary complexity here, since the helper installation step already validates whether the located module provides the required behavior. I can simplify the discovery path accordingly.

sources/en.comix/src/images.rs
Smexhy: Fair point. This grew incrementally while I was working through the image cases: URL normalization, clean-route probing, processing fallback, and safe logging were added at different stages and each ended up doing its own route check. Once the behavior was settled, those checks should have been collapsed into one parsed-route decision; I missed that in cleanup. I will simplify that part.

The per-page /i/... availability check itself is still intentional, since testing found mixed availability within the same chapter: individual s:1 pages can differ in whether the clean route exists.

sources/en.comix/src/web.rs
Smexhy: That’s fair. I removed the speculative parts I could identify, including the content-scoring fallback during module discovery, unnecessary response aliases, repeated route parsing, and redundant context state.

The remaining size is mostly from two pieces that are tied together: loading the current module-based helper flow, and using that same reader/helper path to prepare tile maps for scrambled pages. The async state/polling is also there because the module setup does not complete synchronously through eval.

I agree the patch is still large, and I’m not trying to argue that the size is ideal. I’ve been careful not to keep changing the working path just for the sake of making the diff smaller, since the current version has been stable and fast in local testing for several days. If there are specific parts that still look unnecessary, I’m happy to trim them further, but I think the remaining complexity is mostly from the module setup and selective scrambled-image handling rather than unrelated changes.

jianmingyong:
I don't think the scrambled ones work anymore with this PR as they make them 404 for some reason if they don't match a certain request header or something else.

Update:
For all pages that are marked with `s=1`, you need to include `{ Origin: "https://comix.to/" }` in the header in order to get the scrambled/non-scrambled image. (Seed can be 0 which is non-scrambled)

For all pages that doesn't have `s=1`, you must exclude `{ Origin: "https://comix.to/" }` in the header (or simply do not add anything into the request header) to get the normal image.

They may change this requirement though as long as you use their JS function to handle image, they will have correct request header. (Then again, any images without `s=1` must not run the JS function at all, or you get no image)

Smexhy:
Refreshed against main after [#532](https://github.com/Aidoku-Community/sources/issues/532) merged, also trimmed and simplified things further along the way. Squashed into one commit so the diff against current main is readable. No urgency, leaving open in case the WebView-per-image cost in [#532](https://github.com/Aidoku-Community/sources/issues/532) becomes a concern, otherwise feel free to close.

jianmingyong:
There’s still some good information here that I could use.

PR #540, opened 2026-05-28
feat: add en.freewebnovel

MrGcGamer:
Hey there,
This PR adds en.freewebnovel source.

~~Fair warning:~~
~~A lot of this is vibe-coded, but I reworked some of the more lackluster code and proofread the code.~~
~~However, if you do find anything which can be improved upon, do let me know.~~

~~For now, I will open this as a draft, as I am still working on it, but if you already have stuff to say, I'd love to hear it.~~

## Summary:
- Add a new English freewebnovel.com source
- Implement home sections, listings, search, series details, chapters, pages, and deep links
- Resolve genre tags

--- 

Alright, after vibecoding the start to get it into a somewhat working state I basically rewrote all of the logic, so the code should be mostly clean now.

NOTE: The icon is AI generated, as the website doesn't seem to have a high enough quality version of their logo available anywhere

This ready for review now


sources/en.freewebnovel/src/helpers.rs
```diff
@@ -0,0 +1,314 @@
...
+	imports::{
+		html::{Document, Element},
+		net::Request,
+	},
+	prelude::*,
+};
+
+pub fn request_html(url: &str) -> Result<Document> {
+	Ok(Request::get(url)?.html()?)
+}
```
kkantan, line 15: I guess this is fine to have but it seems like it's not really doing anything to justify having it.

sources/en.freewebnovel/src/helpers.rs
```diff
@@ -0,0 +1,314 @@
...
+
+/// Extract Chapters from Novel page.
+pub fn extract_chapters(html: &Document) -> Vec<Chapter> {
+	let mut chapters = Vec::new();
+	append_chapters_from_doc(html, &mut chapters);
+	chapters.reverse();
+	chapters
+}
+
+fn append_chapters_from_doc(doc: &Document, chapters: &mut Vec<Chapter>) {
```
kkantan: I don't think there's any reason to have two separate functions for this.

sources/en.freewebnovel/src/helpers.rs
```diff
@@ -0,0 +1,314 @@
...
+	let Some(items) = doc.select("div.m-newest2 > ul.ul-list5 > li") else {
+		return;
+	};
+
+	for item in items {
```
kkantan: instead of using `.reverse()` on a `Vec`, you should use `.rev()` on the iterator here to be more efficient.
```suggestion
	for item in items.rev() {
```

sources/en.freewebnovel/src/helpers.rs
```diff
@@ -0,0 +1,314 @@
...
+		let Some(link) = item.select_first("a[href]") else {
+			continue;
+		};
+		let Some(url) = link.attr("abs:href") else {
+			continue;
+		};
+		let Some((_, Some(chapter_key))) = parse_novel_and_chapter(&url) else {
+			continue;
+		};
```
kkantan: instead of using a for loop, you could collect `Chapter`s into a vec from an iterator, and then leverage the `?` inside of a `filter_map` to simplify. most other sources do this and could be checked for reference.

sources/en.freewebnovel/src/helpers.rs
```diff
@@ -0,0 +1,314 @@
...
+	Ok(parts.join("\n\n"))
+}
+
+pub fn parse_search_results(html: &Document) -> Vec<Manga> {
+	let mut entries = Vec::new();
+	for heading in ["Hot Novels", "Hot Latest Novels", "Hot Completed Novels"] {
+		if let Some(hot) = find_section_container(html, heading) {
+			hot.remove();
+		}
+	}
```
kkantan: is this necessary? usually it's best to avoid the html mutation apis when possible. if certain items are being selected unintentionally, you can usually change the css selector to take that into account.

sources/en.freewebnovel/src/helpers.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+pub fn fill_manga_details(html: &Document, mut manga: Manga) -> Result<Manga> {
+	let Some(title) = get_meta_data(html, MetaSelector::Title) else {
+		bail!("Title not found");
+	};
+	manga.title = title;
+	manga.cover = get_meta_data(html, MetaSelector::Cover);
+	manga.url = get_meta_data(html, MetaSelector::Url);
+	const DESCRIPTION_QUERY: &str = "h4.abstract + div.txt p";
+	if let Some(parts) = html.select(DESCRIPTION_QUERY) {
```
kkantan: I don't think the const variable is necessary.
```suggestion
	if let Some(parts) = html.select("h4.abstract + div.txt p") {
```

sources/en.freewebnovel/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+		format!("{BASE_URL}/sort/{kind}")
+	} else {
+		format!("{BASE_URL}/sort/{kind}/{page}")
+	}
+}
+
+fn has_next_page(html: &Document, kind: &str, page: i32) -> bool {
+	html.select_first(format!("a[href*='/sort/{kind}/{}']", page + 1))
+		.is_some()
+}
```
kkantan: these are helper functions that should probably also be in `helpers.rs`.

sources/en.freewebnovel/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+
+fn empty_page_result() -> MangaPageResult {
+	MangaPageResult {
+		entries: Vec::new(),
+		has_next_page: false,
+	}
+}
```
kkantan: I think this is just the same as `MangaPageResult::default()`, so it shouldn't be necessary.

kkantan (changes requested):
to fix the "Test PR" check, `minAppVersion` should be set to `0.8` in `source.json`.

overall looks fine, and appears to work fine on device. I just left a few comments about code styling and organization.

sources/en.freewebnovel/src/helpers.rs
MrGcGamer: Personally I like having it like that..

sources/en.freewebnovel/src/helpers.rs
MrGcGamer: that was actually not needed anymore, as I adjusted the selector after I wrote that part to fix something else

sources/en.freewebnovel/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+crate-type = ["cdylib"]
+
+[profile.dev]
+panic = "abort"
+
+[profile.release]
+panic = "abort"
+opt-level = "s"
+strip = true
+lto = true
```
kkantan:
```suggestion
lto = true

```

kkantan (changes requested):
have you run `cargo fmt`? there are a few things that look like they're formatted slightly inconsistently. other than that it looks good.

MrGcGamer:
> have you run `cargo fmt`? there are a few things that look like they're formatted slightly inconsistently. other than that it looks good.

I did.. I am also not 100% with how it did things, but that's the result of cargo fmt

kkantan:
we can merge this since it looks like the test is only failing due to an aidoku cli bug: https://github.com/Aidoku/aidoku-rs/issues/36.

kkantan (approved):

PR #541, opened 2026-05-28
feat: add multi.myreadingmanga

### LINKED ISSUE #335: myreadingmanga.info
https://github.com/Skittyblock/aidoku-community-sources/blob/main/src/rust/multi.myreadingmanga

The source already exists in the legacy version of the api but it hasn't been updated to work in the newer one.

Erizur:
- **feat: add multi.myreadingmanga**
- **chore: update multi.myreadingmanga/cargo.lock**
- **chore: add .cargo**
- **fix(multi.myreadingmanga): lint warn cleanup**

Mostly tried to fix the old version from the old sources, but ended up doing something a little different.
This should close #335

# Summary
- Implemented Homepage, Latest, Popular, Bara & Manga (sections in the site's navbar), Search, Chapters (sub-pages in the site)
- Added basic dynamic filters (queries change between searches in the site, Aidoku cannot do this yet)
- Tag searching & lookup, Video content ignored, Language support

# Validation
- cargo fmt
- cargo clippy
- aidoku package


sources/multi.myreadingmanga/res/source.json
```diff
@@ -0,0 +1,20 @@
+{
+	"info": {
+		"id": "multi.myreadingmanga",
+		"name": "MyReadingManga",
+		"version": 1,
```
kkantan: since you're reusing the id of the legacy source, this should be incremented by one from the last version.
```suggestion
		"version": 12,
```

sources/multi.myreadingmanga/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+		MyReadingManga
+	}
+
+	fn get_search_manga_list(
+		&self,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> aidoku::imports::error::Result<MangaPageResult> {
+		let mut query_str = alloc::format!("?s={}", urlencode(query.as_deref().unwrap_or("")));
```
kkantan: you can either use the `QueryParameters` helper struct, or `encode_uri_component` from aidoku-rs helpers instead of shipping your own urlencode function.

sources/multi.myreadingmanga/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+					let param = match id.as_str() {
+						"status" => "ep_filter_status",
+						"genre" => "ep_filter_genre",
+						"category" => "ep_filter_category",
+						"tag" => "ep_filter_post_tag",
+						"artist" => "ep_filter_artist",
+						"pairing" => "ep_filter_pairing",
+						_ => continue,
+					};
+					query_str.push_str(&alloc::format!("&{}={}", param, value));
```
kkantan: if you import `aidoku::prelude::*` then you don't need to prefix any macros with a namespace.

sources/multi.myreadingmanga/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+
+	fn get_manga_update(
+		&self,
+		manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> aidoku::imports::error::Result<Manga> {
+		if !needs_details && !needs_chapters {
+			return Ok(manga);
+		}
```
kkantan: this shouldn't ever be the case.

sources/multi.myreadingmanga/src/helpers.rs
```diff
@@ -0,0 +1,95 @@
...
+	raw.find(" [")
+		.map_or_else(|| raw.trim().to_string(), |i| raw[..i].trim().to_string())
+}
+
+pub fn get_user_languages() -> Vec<String> {
+	let mut slugs: Vec<String> = Vec::new();
+
+	let langs = defaults_get::<Vec<String>>("languages")
+		.or_else(|| defaults_get::<String>("languages").map(|s| alloc::vec![s]))
+		.or_else(|| defaults_get::<String>("language").map(|s| alloc::vec![s]))
```
kkantan: is this necessary?

sources/multi.myreadingmanga/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+
+		let url = manga.url.as_deref().unwrap_or(&manga.key).to_string();
+		let doc = get(&url)?;
+		let mut updated = parse_manga(&doc, &manga.key)?;
+
+		if !needs_details {
```
kkantan: why are you updating details if `need_details` is `false`?

sources/multi.myreadingmanga/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+		&self,
+		manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> aidoku::imports::error::Result<Manga> {
+		if !needs_details && !needs_chapters {
+			return Ok(manga);
+		}
+
+		let url = manga.url.as_deref().unwrap_or(&manga.key).to_string();
```
kkantan: full urls shouldn't be used as keys, in case the domain ever changes. usually we use url paths and just append them to the base url. ideally, you should match the format of the legacy source's keys, but if not, you need to mark the update as a breaking change so that the app prompts users to migrate.

sources/multi.myreadingmanga/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+impl DeepLinkHandler for MyReadingManga {
+	fn handle_deep_link(
+		&self,
+		url: String,
+	) -> aidoku::imports::error::Result<Option<DeepLinkResult>> {
+		let key = url
+			.replace("aidoku://", "https://")
+			.replace("http://myreadingmanga.info", "https://myreadingmanga.info");
```
kkantan: I don't think either of these are necessary. it only uses the url(s) you define in `source.json`.

sources/multi.myreadingmanga/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+		Ok(Some(DeepLinkResult::Manga { key }))
+	}
+}
+
+impl ImageRequestProvider for MyReadingManga {
+	fn get_image_request(
+		&self,
+		url: String,
+		_context: Option<PageContext>,
+	) -> aidoku::imports::error::Result<Request> {
```
kkantan: I think you should import this into the namespace since it's a frequently used type, but you could also access it with `aidoku::Result`, which is shorter.

sources/multi.myreadingmanga/src/parser.rs
```diff
@@ -0,0 +1,267 @@
...
+				for link in links {
+					if !tags.contains(&link) {
+						tags.push(link);
+					}
+				}
+			}
+		}
+	}
+
+	let mut chapters: Vec<Chapter> = Vec::new();
```
kkantan, line 234: in my opinion, you should have a separate function for parsing chapters. there's no reason to parse them if `needs_chapters` is false. you could also pass it in as a parameter.

sources/multi.myreadingmanga/src/parser.rs
```diff
@@ -0,0 +1,267 @@
...
+		chapters.push(Chapter {
+			key: key.to_string(),
+			chapter_number: Some(1.0),
+			language: chapter_language,
+			url: Some(key.to_string()),
+			..Default::default()
+		});
+	} else {
+		// I'm not sure about this but this looks cleaner when I tested
+		chapters.reverse();
```
kkantan: you should probably reverse the iterator instead of an already collected vec. something like `for link in &page_links.rev()`.

kkantan (changes requested):

kkantan:
to fix the "Test PR" workflow failing, you should run `cargo update aidoku`.

sources/multi.myreadingmanga/src/helpers.rs
```diff
@@ -0,0 +1,65 @@
...
+	let mut slugs: Vec<String> = Vec::new();
+
+	let langs = defaults_get::<Vec<String>>("languages").unwrap_or_default();
+
+	for lang in langs {
+		if let Some(slug) = map_lang_to_class(&lang) {
+			let slug = slug.to_string();
+			if !slugs.contains(&slug) {
+				slugs.push(slug);
+			}
```
kkantan: you don't need to allocate a string before checking if it's pushed in case it isn't, but I'm guessing that this is unnecessary since the setting value can only contain the unique options you give it.
```suggestion
			if !slugs.contains(slug) {
				slugs.push(slug.into());
			}
```

sources/multi.myreadingmanga/src/helpers.rs
```diff
@@ -0,0 +1,65 @@
...
+		"all" | "none" | "any" | "" => None,
+		"en" | "english" => Some("english"),
+		"ja" | "jp" | "japanese" => Some("jp"),
+		"zh" | "cn" | "chinese" => Some("chinese"),
+		"ko" | "kr" | "korean" => Some("korean"),
+		"es" | "spanish" => Some("spanish"),
+		"fr" | "french" => Some("french"),
+		"de" | "german" => Some("german"),
+		"it" | "italian" => Some("italian"),
+		"pt" | "portuguese" => Some("portuguese"),
```
kkantan: I don't think you need these extra options to match against, since it should only be the values you provide for the source languages.

sources/multi.myreadingmanga/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+		})
+	}
+}
+
+impl DeepLinkHandler for MyReadingManga {
+	fn handle_deep_link(&self, url: String) -> Result<Option<DeepLinkResult>> {
+		let key = url
+			.trim_start_matches(BASE_URL)
+			.trim_start_matches(HTTP_URL)
+			.trim_end_matches('/')
```
kkantan, line 147: I believe only https links will be passed to the deeplink handler, since only the https link is provided for the source url.
```suggestion
		let key = url
			.trim_start_matches(BASE_URL)
			.trim_end_matches('/')
```

sources/multi.myreadingmanga/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+		})
+	}
+
+	fn get_manga_update(
+		&self,
+		manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		let mut manga = manga;
```
kkantan: you should be able to make the function parameter mutable instead of defining a new variable.
```suggestion
	fn get_manga_update(
		&self,
		mut manga: Manga,
		needs_details: bool,
		needs_chapters: bool,
	) -> Result<Manga> {
```

sources/multi.myreadingmanga/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+		MyReadingManga
+	}
+
+	fn get_search_manga_list(
+		&self,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> aidoku::imports::error::Result<MangaPageResult> {
+		let mut query_str = alloc::format!(
```
kkantan: you're importing `Result` and `format` already, so you could skip the full namespace.
```suggestion
	) -> Result<MangaPageResult> {
		let mut query_str = format!(
```

kkantan (changes requested):
the source appears to work fine, I just left a couple extra suggestions.

kkantan (approved):

Issue #502, opened 2026-05-10
mangadot.net


nightspringer:
library is big and constantly expanding

https://mangadot.net

[keiyoushi source reference](https://github.com/keiyoushi/extensions-source/blob/main/src/en/mangadotnet/src/eu/kanade/tachiyomi/extension/en/mangadotnet/Mangadotnet.kt)

[paperback source reference](https://github.com/Catta1997/Sinon-Paperback-Extensions/tree/0.9/stable/src/MangaDot)

InFamouz22:
add this one ASAP pls. thanks in advance.

jianmingyong:
The most atrocious json array output... especially to implement in rust. It might be slightly harder for someone to pick up the mantle to implement the basic feature. I might try but don't get your hopes up.

jianmingyong:
Update: I don't think it is possible currently. The CloudFlare challenge is too aggressive making all images unable to load.
Update 2: It seems to be a broken cache issue. Had to restart device and clear source cache.

You can track from:
https://github.com/jianmingyong/Aidoku-sources/tree/mangadotnet

nightspringer:
Thank you! Also the dev in their discord said they know about the genre issue, and will fix it in the next update or two since the site is new. You can suggest changes on the website on the community forum or discord listed there as well if you see any issues, the dev's really open to fixing it and improving it. There's a list of to do's on the site as well as transparency on the fixes here (u can comment and suggest as well) https://mangadot.net/community

I think there's a big updated coming within the next week or two.

jianmingyong:
It is getting close to a release though now left with some last minute additions and testing. It will still be awful to use due to aidoku CF handler being sux.

jianmingyong:
@nightspringer enjoy. Make sure you update your aidoku to 0.8.3 or it won’t show up.

nightspringer:
> @nightspringer enjoy. Make sure you update your aidoku to 0.8.3 or it won’t show up.


thank you! just added it! appreciate your hard work 🙏🤍. let's see how it holds up with the cloudflare thing haha

PR #545, opened 2026-05-30
feat: add multi.mangadotnet

jianmingyong:
# Summary

This PR add a new source for [https://mangadot.net](https://mangadot.net).

~~The website may support other languages, but the majority of the uploads are only in English hence it is only scoped to English in this source.~~ Change of plans, just allow other languages so it is future proofed instead of a breaking change later.

Just a side note, this website had a VERY AGGRESSIVE Cloudflare challenge. If you get an error, most likely you will need to clear the cache and restart the app to continue. There is nothing I could do. I could use web_view to bypass the session, but it won't solve the images.

Reference:
- #502 

# To do
- [x] Implement Basic Requirements
- [x] Add Proper Listings
- [x] Clean up Garbage Genres (This is not really my fault, but the original website had such a broken genre list)
- [x] Add Content Filters in Settings
- [x] Testing
- [x] Let Clippy be happy


jianmingyong:
You know what, maybe I shall update to multi instead and future proof it rather than scoping to english only.

I checked the upload form and there is a list of languages supported and it is up to the user to upload them. (But yea the library for other languages is as close to none though… maybe 1% if not less)

jianmingyong:
Majority of the things had been refactored just to make it consistent.

This source had its own CF quirks and aidoku cant handle properly. I added a custom error message for that particular case so people know what they can do to resolve it.

It is dumb but you had to remove source cache and restart application for it to start a clean session.

Genres are now currently generated dynamically as this website is still new and will change. It is hard to maintain the static version when nsfw and normal mode had different genres as well -_-

NSFW mode now will render non-nsfw content as well as it seems by default, they would set adult=1 (which hides non-nsfw content) but if you login and configure the settings, it shows adult=both. Hence the api is updated to handle this particular case.

sources/multi.mangadotnet/res/source.json
```diff
@@ -0,0 +1,53 @@
...
+			"ru",
+			"es",
+			"es-419",
+			"sv",
+			"th",
+			"tr",
+			"uk",
+			"vi"
+		],
+		"minAppVersion": "0.8.0"
```
kkantan:
```suggestion
		"minAppVersion": "0.8"
```

sources/multi.mangadotnet/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+crate-type = ["cdylib"]
+
+[profile.dev]
+panic = "abort"
+
+[profile.release]
+panic = "abort"
+opt-level = "s"
+strip = true
+lto = true
```
kkantan:
```suggestion
lto = true

```

sources/multi.mangadotnet/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+		&self,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		let mut query_parameters = QueryParameters::new();
+
+		if let Some(query) = query {
+			query_parameters.push("search", Some(&query));
+		}
```
kkantan:
```suggestion
		if query.is_some() {
			query_parameters.push("search", query.as_deref());
		}
```

sources/multi.mangadotnet/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+			entries: if let Some(results) = search_response.results {
+				results.into_iter().map(Into::into).collect()
+			} else {
+				Vec::new()
+			},
+			has_next_page: if let Some(pagination) = search_response.pagination {
+				pagination.current_page < pagination.total_pages
+			} else {
+				false
+			},
```
kkantan:
```suggestion
			entries: search_response.results
				.map(|results| results.into_iter().map(Into::into).collect())
				.unwrap_or_default(),
			has_next_page: search_response.pagination
				.map(|p| p.current_page < p.total_pages)
				.unwrap_or_default(),
```

sources/multi.mangadotnet/src/settings.rs
```diff
@@ -0,0 +1,33 @@
+use aidoku::imports::defaults::defaults_get;
+
+// const LANGUAGES_KEY: &str = "languages";
+const HIDE_NSFW_KEY: &str = "hideNSFW";
+const DEDUPED_CHAPTER_KEY: &str = "dedupedChapter";
+
+/* Not in use yet, but maybe we need to do some mapping once we get enough data on how the language field works.
```
kkantan, line 18: if this isn't in use by the time the source is ready to merge, it might be better to just set the languages array in `source.json` to `["multi"]`, which would still mark the source as multi-language but disable the language filter.

jianmingyong:
Experimenting with CF Workaround. Clearly the session cache is causing issues and it expires with CF in about 2 hours or so. With this workaround using web_view, the page can load again albeit that any new cover cannot be loaded. This is a sign that internally, the cache is interfering with the retry function. It simply will never work no matter how many times you spam it. Further testing will need to be done to see how I can extend this to chapter images as well.

That way, you can have non interrupted reading. 2 hours is a bit short for avid reader. Thumbnail can be fixed by clearing the source cache. (and if you have already loaded them upfront in your library, it will remain as is)

IMG_2698
IMG_2699

jianmingyong:
CF workaround is now a feature for testing. Well it makes it able to read without annoying hassle of having to clear cache every X hours.

I found out that they had special api for user uploads making them fetch the wrong manga images… that is fixed as well.

The timestamp too probably because of changes in between, I need to handle two different case.

This is still pretty much early test. Things had been further refactored so that I can handle different case and right now there is a logger for fetch request to see if there anything getting stuck.

jianmingyong:
Looks like they just updated the site to fix the genre list though now they also updated the API path to return more stuff. The old API still works right now, but I don't know for how long. I am migrating them to the new API path once I figured out the JSON format.

jianmingyong:
More or less close to ready. They might have disabled Cloudflare turnstile temporarily making the webview unnecessary, but I am keeping it just in case they turn it back on once everything is stabilized.

No harm, that function is optional opt-in when you want to avoid CF errors that can't be resolved without clearing the cache.

sources/multi.mangadotnet/res/settings.json
```diff
@@ -0,0 +1,90 @@
...
+				"key": "useWebViewFetch",
+				"title": "Use CF Workaround",
+				"default": false,
+				"refreshes": [
+					"content",
+					"filters"
+				]
+			}
+		],
+		"footer": "CloudFlare workaround uses an invisible WebView to maintain a proper session. This will allow basic function to work except for cover images. If you want to have less interruption between reading, use this option."
```
kkantan: maybe it would be helpful to explain when a user would turn this setting on, since I assume many would not know that the source suddenly not working has to with cloudflare.
```suggestion
		"footer": "CloudFlare workaround uses an invisible WebView to maintain a proper session. This will allow basic function to work except for cover images. If the source stops working for you periodically, try this option."
```

sources/multi.mangadotnet/res/settings.json
```diff
@@ -0,0 +1,90 @@
...
+				"key": "resetFilters",
+				"title": "Reset Filters",
+				"confirmTitle": "Reset Filters",
+				"confirmText": "Are you sure you want to remove all configured content filters?",
+				"destructive": true,
+				"notification": "resetFilters",
+				"refreshes": ["content"]
+			}
+		],
+		"footer": "Content filters are applied to the home page, and listings only."
```
kkantan:
```suggestion
		"footer": "Content filters are applied to the home page and listings only."
```

sources/multi.mangadotnet/src/settings.rs
```diff
@@ -0,0 +1,60 @@
...
+
+pub fn get_default_content_types() -> Option<String> {
+	defaults_get::<Vec<String>>(DEFAULT_CONTENT_TYPES_KEY).map(|ids| ids.join(","))
+}
+
+pub fn reset_filters() {
+	defaults_set(
+		DEFAULT_CONTENT_TYPES_KEY,
+		DefaultValue::StringArray(vec!["JP".into(), "KR".into(), "CN,TW".into()]),
+	)
```
kkantan: I think setting this to `DefaultValue::Null` might be better, since that should use the default you have set in `filters.json`, in case it ever changes.

sources/multi.mangadotnet/src/lib.rs
```diff
@@ -0,0 +1,603 @@
...
+			query_parameters.push("page", Some(&format!("{}", page)));
+		}
+
+		if let Some(content_types) = get_default_content_types() {
+			for content_type in content_types.split(",") {
+				query_parameters.push("origin", Some(content_type));
+			}
+		}
+
+		match listing.id.as_str() {
```
kkantan: it seems like the code could be deduplicated a bit here since all the branches are very similar.

jianmingyong:
hmmm it seems like having a preload > 2 images can cause a permanent freezing. I was told it will simply just run things in random order without any need to care about thread safety… have to debug further what is going on.

jianmingyong:
Edit: It seems to be just a normal request flow getting stuck with a very long timeout. I can't do anything about this. Webview variant had a 30 seconds timeout which is faster to unstuck itself.

jianmingyong:
oh great, CF workaround doesn’t work with 4G/5G but wifi seems fine. I guess CF doesn’t like my mobile network but wifi seems okay to them. Someone else has to test that.

Other than that, it seems fine. People might complain the source not working in certain conditions. The best alternative is to use other source or have aidoku handle CF better.

I will do the static genre filter once the website is mature enough. (If I can also handle them properly, I would do so)

kkantan (approved):

PR #546, opened 2026-05-31
fix(en.freewebnovel): Chapters without parsable Chapter numbers got lost

MrGcGamer:
Fixes Chapters without a number like `Prologue` getting lost

### DIFF
--- a/sources/en.freewebnovel/res/source.json
+++ b/sources/en.freewebnovel/res/source.json
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "en.freewebnovel",
 		"name": "FreeWebNovel",
-		"version": 1,
+		"version": 2,
 		"url": "https://freewebnovel.com",
 		"contentRating": 1,
 		"languages": [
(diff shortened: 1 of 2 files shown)


sources/en.freewebnovel/res/source.json
```diff
@@ -2,12 +2,12 @@
...
 		"name": "FreeWebNovel",
-		"version": 1,
+		"version": 2,
 		"url": "https://freewebnovel.com",
 		"contentRating": 1,
 		"languages": [
 			"en"
 		],
-		"minAppVersion": "0.8"
+		"minAppVersion": "0.8.0"
```
kkantan: this shouldn't be necessary after https://github.com/Aidoku/aidoku-rs/commit/a0624d4b899fbec54a6ad99400a18ec9202ea9d3.

MrGcGamer: I guess, but also won't do any harm..

kkantan: maybe not currently but I do personally believe it's incorrect and it could possibly change in the future. the actual version of aidoku that the api support is included in is 0.8, not 0.8.0.

MrGcGamer:
Superseded by #547

PR #547, opened 2026-05-31
feat(en.freewebnovel): add markdown

MrGcGamer:
I mostly took the code from `templates/libgroup/converters/html_to_markdown.rs`.

~~One thing I am unsure about is an issue the AI brought up regarding the logic:~~

> ~~Inline text order can be corrupted in markdown conversion
In [helpers.rs:204], convert_children_to_markdown appends element.own_text before recursing into children. For mixed-content nodes like p with text + nested strong + trailing text, this can reorder output (child text may end up after trailing plain text). \[...\] this outputs "prefix suffix **bold**" instead of "prefix **bold** suffix".~~

~~I am not a 100% the AI and me are understanding `element.own_text()` correctly, as tooltip for that one is rather lackluster, but if we are understanding that correctly, that should at least be given a thought.~~

RE: L233: adjacent bold boundaries would mess up each other / not get rendered right

Supersedes #546 (Fixes Chapters without a number like `Prologue` getting lost)

### DIFF
--- a/sources/en.freewebnovel/res/source.json
+++ b/sources/en.freewebnovel/res/source.json
@@ -8,6 +8,6 @@
 		"languages": [
 			"en"
 		],
-		"minAppVersion": "0.8.0"
+		"minAppVersion": "0.8.3"
 	}
 }
--- a/sources/en.freewebnovel/res/source.json
+++ b/sources/en.freewebnovel/res/source.json
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "en.freewebnovel",
 		"name": "FreeWebNovel",
-		"version": 1,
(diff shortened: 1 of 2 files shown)


kkantan (changes requested):
this should be combined with the other pr so that there's only one update to the source, even though it's two different changes.

I think what your "ai" said is correct. I'm guessing that instead of using `own_text` and `children`, you should use the `child_nodes` function, which would let you iterate over both text and child elements.

kkantan (changes requested):
looks good, it just needs the min app version to be set to 0.8.3 now according to `aidoku verify`.

kkantan (approved):

PR #550, opened 2026-06-01
fix(en.comix): chapters and search sometimes not loading

### LINKED ISSUE #549: Comix.to some series fail to load
Loading chapter list of any entires gives the error "**Failed to find descrambler function**", aside from that if you already have cached the chapters list then if you open any chapter it will fail to load the chapter.

Another bug i noticed is when using the search feature there is a chance that it can cause a JsonParse Error, for example after I searched for "Fallen to paradise" I got this error in the logs:
**JsonParseError(Error("invalid type: null, expected struct Poster", line: 1, column: 47956))**

I am well aware that comix has been a mess for the past few weeks but it is still the source with the largest library, so thank you to everyone still trying to fix the extension every time comix makes a change.

### Screenshots
Image
Image
Image

DedPiht0-MGN:
## Fix async function detection in Comix scraper

Resolves #549

### Summary

This pull request improves the reliability of asynchronous function detection in the Comix source scraping logic. Some functions used by the website may not be declared as async but still return a Promise, causing them to be missed by the previous detection mechanism.

### Changes

- Added an isPromise utility function to identify promise-returning functions.
- Added a reusable testCanvas element for safely testing function execution.
- Enhanced async function detection to:
  - Detect functions whose constructor is AsyncFunction.
  - Detect functions that return a Promise when invoked with the expected arguments.
- Updated the source version from 17 to 18.

### Result

The scraper can now correctly identify and handle a wider range of asynchronous functions, improving compatibility with recent changes on the Comix website and fixing the issue reported in #549.

### DIFF
--- a/sources/en.comix/res/source.json
+++ b/sources/en.comix/res/source.json
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "en.comix",
 		"name": "Comix",
-		"version": 17,
+		"version": 18,
 		"url": "https://comix.to",
 		"contentRating": 1,
 		"languages": ["en"]
--- a/sources/en.comix/src/models.rs
+++ b/sources/en.comix/src/models.rs
@@ -118,7 +118,7 @@ pub struct ComixManga {
 	pub title: String,
 	pub synopsis: Option<String>,
 	pub r#type: String,
-	pub poster: Poster,
+	pub poster: Option<Poster>,
 	pub status: String,
 	pub content_rating: ComixContentRating,
 	pub authors: Option<Vec<Term>>,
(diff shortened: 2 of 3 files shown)


kkantan (changes requested):
this looks good, but could you also make the `poster` field an `Option` so that the other bug that the issue reports is fixed?

DedPiht0-MGN:
Here are the specific changes made for fix #549:
1. Changed pub poster: Poster; to pub poster: Option<Poster>; in the ComixManga struct definition. This leverages serde's built-in capability to deserialize null values into None.
2. Updated the From<ComixManga> for Manga trait implementation to handle the optional poster field when mapping to the cover field in the Manga struct. It now uses a pattern match on value.poster to either extract the medium or large image URL from the poster if it exists, or defaults to None if the poster is missing/null.

kkantan (approved):

PR #572, opened 2026-06-15
Fix for multi.mangadex Cloudflare Problem

Methis-mac:
Hello! 
I worked through this with a bit of Claude help since I wanna read manga asap but I read through the code, tested and verified live on my own device against the real MangaDex API - just putting it here in case this quick fix is good enough (pls don't get mad if its bad).

Summary
- Send User-Agent/Referer headers on all MangaDex API requests when previously only image requests had these so it had problems (identifying headers matches what image requests already do and looks like it avoids being treated as bad traffic).

Problem
- The MangaDex API was sometime giving empty responses, causing "Failed to parse popular/recent manga" errors and search failures on iOS. A few people including me had this problem.

Testing
- Built and deployed via a personal fork's GitHub Pages source list, installed on iOS device on 0.8.2
- Confirmed Home tab (Popular New Titles, Recently Added, Latest Updates) and search now load correctly! :D
- cargo fmt --check and cargo clippy all clean and well

### DIFF
--- a/sources/multi.mangadex/src/home.rs
+++ b/sources/multi.mangadex/src/home.rs
@@ -30,7 +30,7 @@ impl Home for MangaDex {
 			CUSTOM_LISTS
 				.iter()
 				.map(|list| format!("{API_URL}/list/{list}"))
-				.map(|url| Request::get(url).expect("invalid url format")),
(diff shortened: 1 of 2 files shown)


kkantan:
thank you for the fix, but since the source version needed to be incremented as well, I just created a new pr to update it quickly.

PR #575, opened 2026-06-16
fix(vi.nettruyen): update url

tachibana-shin:


kkantan (changes requested):
can you set `"minAppVersion": "0.7.1"` to pass the workflow test?

tachibana-shin:
Wow, it seems like a lot has changed while I was away

kkantan (approved):

PR #577, opened 2026-06-16
fix(vi.zettruyen): update url

tachibana-shin:


kkantan (changes requested):
this is updating the cmanga domain instead of the zettruyen domain in the rust source.

also, can you set `"minAppVersion": "0.7.1"` here too?

kkantan (approved):

PR #579, opened 2026-06-18
fix(en.comix): not loading any manga

### LINKED ISSUE #567: Comix is not loading any manhwas/mangas and even chapters
Comix is not loading even though when a cloud flare verification pops up I verify it but it doesn’t load any manhwas or mangas and also the the manhwas/mangas that have been saved it doesn’t load any chapters

jianmingyong:
This PR contains a few things that is still currently WIP hence it is not in working state just yet. Don't bother testing this just yet until it is marked ready.

1. WebView is now shared instead of making a new one each request.
2. All Request will now use the Axios request interceptors instead so it can determine whether to include the signature token or not.
3. All Response will now use the Axios response interceptors instead so it can determine whether to decode or not.
4. Images now requires guessing between blob or canvas for scrambled pages. This is so annoying and need more time to test.

Fixes #567

### DIFF
--- a/sources/en.comix/src/web.rs
+++ b/sources/en.comix/src/web.rs
@@ -15,76 +15,88 @@ if (!vmObj || typeof vmObj !== 'object' || vmObj === window) {\
 	return '';\
 }";
 
+const INSTALLER_FN_TOKEN: &str = "__AIDOKU_INSTALLER_FN_TOKEN__";
+const INSTALLER_REQUEST_TOKEN: &str = "__AIDOKU_INSTALLER_REQUEST_TOKEN__";
+const INSTALLER_RESPONSE_TOKEN: &str = "__AIDOKU_INSTALLER_RESPONSE_TOKEN__";
+const DESCRAMBLER_FN_TOKEN: &str = "__AIDOKU_DESCRAMBLER_FN_TOKEN__";
+
 const JS_PATCHER: &str = "<head><script>window['originalGetImageData'] = HTMLCanvasElement.prototype.toDataURL;</script>";
 
 pub struct ComixWebView {
 	web_view: WebView,
+	is_initialized: bool,
 	installer_fn: Option<String>,
 	descrambler_fn: Option<String>,
 }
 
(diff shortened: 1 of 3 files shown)


jianmingyong:
I have clean up most of the stuff. Might not be that clean still... pretty much the same idea from mangadotnet is being used here to make the web flow better with a timeout to avoid getting stuck somewhere.

This should work for any URL changes that require signature token or not. (And hopefully they don't change the scrambler function again to add a third layer)

sources/en.comix/src/web.rs
```diff
@@ -100,209 +140,320 @@ fn find_functions(web_view: &mut ComixWebView) -> Result<()> {
...
-						transformResponse: []
+	pub fn decode_json_owned<T>(&mut self, response: &Response) -> Result<T>
+	where
+		T: DeserializeOwned,
+	{
+		if !self.is_initialized {
+			self.load_webview()?;
+		}
+
+		if response.status_code() == 403
```
kkantan, line 329: I believe this function makes an ffi call, so it may be a little better to store it in a variable after being called once instead of calling it multiple times.

sources/en.comix/src/web.rs
```diff
@@ -100,209 +140,320 @@ fn find_functions(web_view: &mut ComixWebView) -> Result<()> {
...
+			if result.starts_with("error:") {
+				bail!("{result}");
+			} else if result.is_empty() {
+				bail!("Failed to fetch result")
+			}
+
+			serde_json::from_str(&result).map_err(|e| error!("Invalid json: {}", e))
+		} else {
+			let json_str = response.get_string()?;
+			serde_json::from_str(&json_str).map_err(|e| error!("Invalid json: {}", e))
```
kkantan, line 372: I think there's a function `get_json` on the response struct that does this same thing already.

kkantan (approved):

jianmingyong:
I will probably improve those on the next PR. (and hopefully comix doesn’t try something complex)

PR #581, opened 2026-06-20
fix(en.comix): Descrambler update to fix the image and api

### LINKED ISSUE #580: Comix is down again
Everything became fine when the recent update came which fix the bug but the next few hours it goes down I don’t if know if I am the only one experiencing this because I heard that it actually working on ios26  which I don’t know if it is true but I am using ios18.7.9. This is the error it says : 

Failed to find descrambler canvas function

jianmingyong:
I made the error less strict this time and keep the old method just in case they decide to rollback. The new method is also added into it.

Previous code review stuff is also added into this fix.

Fixes #580

### DIFF
--- a/sources/en.comix/src/web.rs
+++ b/sources/en.comix/src/web.rs
@@ -194,7 +194,7 @@ impl ComixWebView {
 		if expr[0].is_empty() {
 			bail!("Failed to find installer function");
 		}
-		if expr[1].is_empty() && expr[2].is_empty() {
+		if expr.len() < 3 || expr[1].is_empty() && expr[2].is_empty() {
 			bail!("Failed to find descrambler canvas/blob function");
 		}
 		Ok(())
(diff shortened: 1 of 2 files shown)


sources/en.comix/src/web.rs
```diff
@@ -194,11 +194,8 @@ impl ComixWebView {
 		if expr[0].is_empty() {
 			bail!("Failed to find installer function");
 		}
-		if expr.len() < 2 || expr[1].is_empty() {
-			bail!("Failed to find descrambler canvas function");
-		}
-		if expr.len() < 3 || expr[2].is_empty() {
-			bail!("Failed to find descrambler blob function");
+		if expr[1].is_empty() && expr[2].is_empty() {
+			bail!("Failed to find descrambler canvas/blob function");
```
kkantan: the length of `expr` should still be checked, since I believe this could panic and break the source.

kkantan (changes requested):

kkantan (approved):

PR #582, opened 2026-06-20
feat: add en.ezmanga

### LINKED ISSUE #536: ezmanga.org
source request
url: https://ezmanga.org/
language: EN 

[keiyoushi reference](https://github.com/keiyoushi/extensions-source/blob/main/src/en/ezmanga/src/eu/kanade/tachiyomi/extension/en/ezmanga/EZmanga.kt), same template as https://qimanhwa.com (ezmanhwa)

CtrlAltElite97:
* feat: add ezmanga.org


sources/en.ezmanga/res/icon.png
kkantan, line 1: this should be a 128x128 image with an opaque background.

sources/ezmanga.org/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "en.ezmanga",
+		"name": "ezmanga",
+		"version": 1,
+		"url": "https://ezmanga.org/",
```
kkantan:
```suggestion
		"name": "EzManga",
		"version": 1,
		"url": "https://ezmanga.org",
```

sources/ezmanga.org/src/lib.rs
```diff
@@ -0,0 +1,482 @@
+#![no_std]
+
+#[no_mangle]
+pub unsafe extern "C" fn memcmp(s1: *const u8, s2: *const u8, n: usize) -> i32 {
```
kkantan: what is this for?

sources/ezmanga.org/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+    },
+    helpers::uri::encode_uri_component,
+    imports::net::Request,
+    prelude::*,
+};
+use serde::Deserialize;
+
+const API_BASE: &str = "https://vapi.ezmanga.org/api/v1";
+const BASE_URL: &str = "https://ezmanga.org";
+const UA: &str = "Mozilla/5.0 (iPhone; CPU iPhone OS 17_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.0 Mobile/15E148 Safari/604.1";
```
kkantan: is this necessary?

sources/ezmanga.org/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+
+fn api_get(url: &str) -> Result<Request> {
+    Ok(Request::get(url)?
+        .header("User-Agent", UA)
+        .header("Origin", BASE_URL)
+        .header("Referer", "https://ezmanga.org/"))
+}
+
+// --- API response types ---
```
kkantan: you should put the structs in a `models.rs` file and the helpers in a `helpers.rs` file, to be consistent with other sources.

sources/ezmanga.org/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+    let secs = if b.len() >= 19 {
+        let h = parse_num_bytes(&b[11..13])?;
+        let mi = parse_num_bytes(&b[14..16])?;
+        let se = parse_num_bytes(&b[17..19])?;
+        h * 3600 + mi * 60 + se
+    } else {
+        0
+    };
+    Some(days * 86400 + secs)
+}
```
kkantan: can you not use parse_date function included with aidoku-rs? with the format `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'`.

sources/ezmanga.org/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+
+fn item_to_manga(s: ApiSeriesItem) -> Manga {
```
kkantan: implementing `From<ApiSeriesItem> for Manga` is better so that you can use the `Into` trait, or otherwise implementing a function on the `ApiSeriesItem`.

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+                kind: ListingKind::Default,
+            },
+            Listing {
+                id: String::from("latest"),
+                name: String::from("Latest"),
+                kind: ListingKind::Default,
+            },
+        ])
+    }
+}
```
kkantan, line 220: these should be static instead of dynamic.

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+
+impl DynamicFilters for EzManga {
```
kkantan, line 222: likewise, there is no reason for dynamic filters here. you can make a `filters.json` instead.

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+                HomeComponent {
+                    title: Some(String::from("Latest")),
+                    subtitle: None,
+                    value: HomeComponentValue::Scroller {
+                        entries: latest_links,
+                        listing: Some(latest_listing),
+                    },
+                },
+            ],
+        })
```
kkantan, line 335: it would be nice to match the website home page better, with "Popular Today", "Pinned Series", and "Latest Updates" sections.

sources/ezmanga.org/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "en-ezmanga"
```
kkantan:
```suggestion
name = "ezmanga"
```

sources/ezmanga.org/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "0.1.0"
+edition = "2021"
```
kkantan:
```suggestion
edition = "2024"
```

sources/ezmanga.org/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+                    _ => {}
+                },
+                "type" => match value.as_str() {
+                    "Manhwa" => url.push_str("&type=MANHWA"),
+                    "Manga" => url.push_str("&type=MANGA"),
+                    "Manhua" => url.push_str("&type=MANHUA"),
+                    "Novel" => url.push_str("&type=NOVEL"),
+                    _ => {}
+                },
+                _ => {}
```
kkantan: you could simplify this by optionally using `QueryParameters` and formatting the "value" in the filter correctly so you don't need to do any matching.

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+            "{}/series/{}/chapters/{}",
+            API_BASE, manga.key, chapter.key
+        ))?
+        .json_owned()?;
+
+        let pages = det
+            .images
+            .into_iter()
+            .map(|img| Page {
+                content: PageContent::url(img.url),
```
kkantan, line 175: it looks like there's also one series on the website that's a novel, with text chapters. it might be good to filter those out if the source doesn't support them, or add support.

kkantan (changes requested):
the source should also be contained in an "en.ezmanga" folder instead of "ezmanga.org", to match other sources.

kkantan:
also, it would be nice to refactor this source into a template so that it could be used for qiscans too, like [the keiyoushi source](https://github.com/keiyoushi/extensions-source/blob/006a19d04ea1cbd76ad7ea8fa74f6aab6f16180e/lib-multisrc/ezmanhwa/src/eu/kanade/tachiyomi/multisrc/ezmanhwa/EZManhwa.kt), but that's optional.

sources/ezmanga.org/src/lib.rs
CtrlAltElite97: Removed — it was a leftover from an earlier attempt to satisfy the linker in a no_std environment, but it turns out it isn't needed here.

sources/ezmanga.org/src/lib.rs
CtrlAltElite97: Removed — I tested without it and the API responds correctly. Only Origin and Referer headers are kept.

sources/ezmanga.org/src/lib.rs
CtrlAltElite97: Done — structs moved to src/models.rs and helper functions moved to src/helpers.rs.

sources/ezmanga.org/src/lib.rs
CtrlAltElite97: Filter logic simplified with QueryParameters + ids on SelectFilter

sources/en.ezmanga/src/lib.rs
CtrlAltElite97: Added filtering — series_type is now deserialized from the API's type field on ApiSeriesItem, and any entry with type = "NOVEL" is filtered out of browse, search, and home listings since the source only supports image pages.

sources/en.ezmanga/res/icon.png
CtrlAltElite97: Done size set to 128x128

sources/ezmanga.org/Cargo.toml
CtrlAltElite97: Made the changes as per the instructions

sources/en.ezmanga/src/lib.rs
CtrlAltElite97: Added filtering — series_type is now deserialized from the API's type field on ApiSeriesItem, and any entry with type = "NOVEL" is filtered out of browse, search, and home listings since the source only supports image pages.

PR #586, opened 2026-06-23
feat: add en.ezmanga

CtrlAltElite97:
Feat : Added en.ezmanga

Changes made as per #582 (Comment)


sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+        Self
+    }
+
+    fn get_search_manga_list(
+        &self,
+        query: Option<String>,
+        page: i32,
+        filters: Vec<FilterValue>,
+    ) -> Result<MangaPageResult> {
+        let url = match query.as_deref().filter(|q| !q.is_empty()) {
```
kkantan: the query will never be empty, since it would be `None` instead.

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+
+            let mut authors: Vec<String> = Vec::new();
+            if let Some(a) = det.author.filter(|s| !s.is_empty()) {
+                authors.push(a);
+            }
+            if let Some(a) = det.artist.filter(|s| !s.is_empty())
+                && !authors.contains(&a)
+            {
+                authors.push(a);
+            }
```
kkantan: why not put the artist in the `artists` manga field instead of `authors`?

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+                    subtitle: None,
+                    value: HomeComponentValue::Scroller {
+                        entries: latest_links,
+                        listing: Some(Listing {
+                            id: String::from("Latest"),
+                            name: String::from("Latest"),
+                            kind: ListingKind::Default,
+                        }),
+                    },
+                },
```
kkantan, line 272: could these be `MangaChapterList`s instead, to match the website? as long as you have the latest chapter number.

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+            ],
+        })
+    }
+}
+
+impl DeepLinkHandler for EzManga {
+    fn handle_deep_link(&self, url: String) -> Result<Option<DeepLinkResult>> {
+        let prefix = "https://ezmanga.org/series/";
```
kkantan: it would be nice to use `BASE_URL` here so it's easier to change in the future.

kkantan (changes requested):
why are you closing your prs and opening new ones? it would be easier to just update your existing one.

the test pr workflow is also saying you need to set the source `minAppVersion` to `0.7.1`.

sources/en.ezmanga/src/lib.rs
CtrlAltElite97: Query filter (line 47) — removed .filter(|q| !q.is_empty()) since query is None not Some("") when empty.

sources/en.ezmanga/src/lib.rs
CtrlAltElite97: Artist field (lines 104–109) — author goes into manga.authors, artist goes into manga.artists separately.

sources/en.ezmanga/src/lib.rs
CtrlAltElite97: Deep link handler (line 265) — hardcoded "https://ezmanga.org/series/" replaced with format!("{}/series/", BASE_URL).

sources/en.ezmanga/src/lib.rs
CtrlAltElite97: models.rs — added ApiHomeChapter { slug, number } and a #[serde(default)] chapters: Vec<ApiHomeChapter> field on ApiSeriesItem (defaults to empty if the field is absent).
- lib.rs — swapped Link import for MangaWithChapter, replaced all three Scroller components with MangaChapterList, using a shared to_entry closure that extracts the first chapter from each item (and skips items with no chapters via filter_map).

CtrlAltElite97 (commented):
Added en.ezmanga source changes per PR review: minAppVersion, query fix, artist field separation, BASE_URL in deep link, and MangaChapterList for all home sections.

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+                    {
+                        qs.push(id, Some(value));
+                    }
+                }
+                format!("{}/series?{}", API_BASE, qs)
+            }
+        };
+
+        let resp: ApiList<ApiSeriesItem> = api_get(&url)?.json_owned()?;
+        let has_next = resp.next.is_some();
```
kkantan: small nitpick, but you could just name this `has_next_page` and use the struct initialization shorthand like you do for `entries`.

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+            .filter(filter_novels)
+            .filter_map(to_entry)
+            .collect::<Vec<_>>();
+
+        Ok(HomeLayout {
+            components: vec![
+                HomeComponent {
+                    title: Some(String::from("Popular Today")),
+                    subtitle: None,
+                    value: HomeComponentValue::MangaChapterList {
```
kkantan: this one should still be a Scroller like you had before, not a MangaChapterList.

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+
+                for ch in resp.data {
+                    if ch.locked {
+                        continue;
+                    }
+                    chapters.push(Chapter {
+                        key: ch.slug,
+                        chapter_number: Some(ch.number as f32),
+                        title: ch.title.filter(|t| !t.is_empty()),
+                        date_uploaded: ch.created_at.as_deref().and_then(|s| parse_date(s, "yyyy-MM-dd")),
```
kkantan: is this correct? I don't see any dates for chapters when testing.

sources/en.ezmanga/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+                    API_BASE, manga.key, page
+                ))?
+                .json_owned()?;
+
+                let has_next = resp.next.is_some();
+
+                for ch in resp.data {
+                    if ch.locked {
+                        continue;
+                    }
```
kkantan: I'm not sure if this is reliable. for example, I opened "I Became the Stepmother of a Terminally Ill Child", and it displayed both chapter 63 and 64 despite them being premium chapters that can't be loaded.

kkantan (changes requested):
this is mostly good, I just had a few more comments after testing on a device.

sources/en.ezmanga/src/lib.rs
CtrlAltElite97: Renamed to has_next_page in both search and listing, enabling struct field shorthand

sources/en.ezmanga/src/lib.rs
CtrlAltElite97: "Popular Today" changed to HomeComponentValue::Scroller with Vec<Link> entries

sources/en.ezmanga/src/lib.rs
CtrlAltElite97: Strips milliseconds via split_once('.') then parses with "yyyy-MM-dd'T'HH:mm:ss'Z'" — matches the actual API format "2026-06-24T17:52:43.319Z"

sources/en.ezmanga/src/lib.rs
CtrlAltElite97: Replaced locked: bool with is_free: bool mapped from isFree (defaults to true); updated guard to if !ch.is_free

kkantan (changes requested):
the ezmanga source looks good to merge, but you added another source to this pr. separate sources should be in separate prs so they can be reviewed individually, so you should remove en.dankemoe and make a new pr for it. however, as a heads up if you do: the danke website uses a template named "guya", so you should refactor the source code into a template if possible.

kkantan (approved):

PR #587, opened 2026-06-23
feat: add zh.zerobyw

zyp-0:
Adds zh.zerobyw.

Note: VIP chapters are locked – users need to purchase VIP on the website to access them.


sources/zh.zerobyw/res/source.json
```diff
@@ -0,0 +1,14 @@
...
+	"info": {
+		"id": "zh.zerobyw",
+		"name": "zero搬运网",
+		"version": 1,
+		"url": "https://www.zerobyw33.com",
+		"contentRating": 1,
+		"languages": ["zh"]
+	},
+	"config": {
+		"allowsBaseUrlSelect": true,
```
kkantan: does this do anything? it doesn't look like you're providing multiple urls.

sources/zh.zerobyw/src/net.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    pub fn from_query_or_filters(
+        query: Option<&str>,
+        page: i32,
+        filters: &[FilterValue],
+    ) -> Result<Self> {
+        if let Some(keyword) = query {
+            let url = Self::Search(SearchQuery::new(keyword, page));
+            return Ok(url);
+        }
```
kkantan: the website appears to allow searching at the same time as filtering. is there a reason the source can't do the same?

sources/zh.zerobyw/src/html.rs
```diff
@@ -0,0 +1,245 @@
...
+                key: zjid,
+                title,
+                url,
+                // locked: true,
+                ..Default::default()
+            });
+        }
+
+        if chapters.is_empty() {
+            return Err(error!("No chapters found"));
```
kkantan: does this make sense to have? are there no entries on the website with zero chapters?
```suggestion
            bail!("No chapters found");
```

sources/zh.zerobyw/src/html.rs
```diff
@@ -0,0 +1,245 @@
...
+            };
+            pages.push(Page {
+                content: PageContent::Url(url, None),
+                ..Default::default()
+            })
+        }
+
+        if pages.is_empty() {
+            return Err(error!("No chapters found"));
+        }
```
kkantan: this isn't strictly necessary to have, since the app will error in the same way if the page list is empty.
```suggestion
        if pages.is_empty() {
            bail!("No pages found");
        }
```

sources/zh.zerobyw/src/net.rs
```diff
@@ -0,0 +1,241 @@
...
+        .to_string();
+
+    let loginhash = action
+        .split("loginhash=")
+        .nth(1)
+        .and_then(|s| s.split('&').next())
+        .ok_or_else(|| error!("loginhash not found"))?
+        .to_string();
+    if loginhash.is_empty() {
+        return Err(error!("loginhash is empty"));
```
kkantan:
```suggestion
        bail!("loginhash is empty");
```

sources/zh.zerobyw/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+crate-type = ["cdylib"]
+
+[profile.dev]
+panic = "abort"
+
+[profile.release]
+panic = "abort"
+opt-level = "s"
+strip = true
+lto = true
```
kkantan:
```suggestion
lto = true

```

kkantan (changes requested):
is it possible to detect the vip chapters and mark them as locked unless the user is logged in?

sources/zh.zerobyw/res/source.json
zyp-0: Thanks for the review.I've removed unnecessary configs.

zyp-0:
Thanks for the review. Here's a summary of the changes I've made:
- I've added support for combined search + filters. Now SearchQuery also carries filter parameters.
- Replaced explicit errors with bail! where appropriate.
- VIP chapters are now correctly marked as locked with valid keys.
- Ensured all files have a newline at the end.
All changes have been tested locally. Please review again when you have a moment.

sources/zh.zerobyw/src/html.rs
```diff
@@ -0,0 +1,237 @@
...
+                .split("kuid=")
+                .nth(1)
+                .ok_or_else(|| error!("Invalid manga href found"))?
+                .to_string();
+            // cover
+            let cover = card.select_first("img").and_then(|img| img.attr("src"));
+            // title
+            let title = card
+                .select_first("h3.manga-card-title")
+                .map(|title| title.text().unwrap().trim().to_string())
```
kkantan: `unwrap()` should not be used. you can use `and_then` here instead of `map` to handle it.

sources/zh.zerobyw/src/net.rs
```diff
@@ -0,0 +1,274 @@
...
+        let mut shuxing = String::new();
+        let mut order = String::from("addtime");
+        let mut dir = String::from("desc");
+
+        for filter in filters {
+            match filter {
+                FilterValue::Select { id, value } => match id.as_str() {
+                    "分类" => category_id = value.clone(),
+                    "进度" => jindu = value.clone(),
+                    "语言" => shuxing = value.clone(),
```
kkantan: values shouldn't need to be cloned.

sources/zh.zerobyw/src/net.rs
```diff
@@ -0,0 +1,274 @@
...
+                &category_id,
+                &jindu,
+                &shuxing,
+                &dir,
+                &order,
+            ));
+            return Ok(url);
+        }
+
+        let filters_query = FiltersQuery::new(&category_id, &jindu, &shuxing, &dir, &order, page);
```
kkantan: I don't think it's necessary to split this into two different paths, since they appear to be identical apart from the keyword url parameter. the urls are also the same, so I'm not sure why they would need to be separate. instead of assigning to mutable variables, and then passing those variables to a function, I think you could just assign directly to a `QueryParameters` instance.

sources/zh.zerobyw/src/html.rs
```diff
@@ -0,0 +1,237 @@
...
+                ..Default::default()
+            });
+        }
+
+        for (i, ch) in chapters.iter_mut().enumerate() {
+            ch.chapter_number = Some((i + 1) as f32);
+            if ch.title.is_none() || ch.title.as_ref().map(|s| s.is_empty()).unwrap_or(true) {
+                ch.title = Some(format!("{}", i + 1));
+            }
+        }
```
kkantan, line 187: why not assign the chapter number based on the title, if it exists? they appear to always be numbers, or ranges.

kkantan (changes requested):

sources/zh.zerobyw/src/html.rs
zyp-0: Thanks for the suggestion about using the title to determine chapter numbers. I've considered it carefully, but I found that this particular website has some inconsistencies that make it unreliable:

Examples from actual chapter lists:

1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 10不同, 11, 84, 85-83, 94-101, 102-111, 601, 602

they include mixed text (10不同), ranges (85-83, 94-101), and sudden jumps (...11, 84, 85-83, ...). Parsing these reliably would be complex and error-prone. I've opted to keep the chapter_number as a simple monotonic index, which guarantees correct sorting. The original titles are still displayed, so users won't lose any information.

zyp-0:
I've pushed a new commit addressing the remaining feedback: merged search/filter paths, removed clones, replaced unwrap with and_then. Please take another look when you have time.

sources/zh.zerobyw/src/html.rs
kkantan: the reason I suggest using the numbers is primarily because of trackers: if someone reads "94-101" and the chapter is marked as 94, the tracker will be updated to 94, which is at least closer than whatever the index of the chapter is.

> Parsing these reliably would be complex and error-prone

I don't think you need to do anything more than parse the first number out of the string. if the text isn't one of the three examples you showed and it fails, it's fine to have no chapter number, but otherwise it would work.

> which guarantees correct sorting

the sorting is done by the order that you place the items in the vector, so it would remain the same even if the chapter numbers changed.

regardless, it's not a requirement to change this behavior, just a suggestion from me.

sources/zh.zerobyw/src/html.rs
```diff
@@ -0,0 +1,238 @@
...
+        let cards = self
+            .select("a.group.block[href^='/pc/details/']")
+            .ok_or_else(|| error!("No manga cards found"))?;
+
+        for card in cards {
+            // id
+            let href = match card.attr("href") {
+                Some(href) => href,
+                None => {
+                    error!("No manga href found");
```
kkantan: since you don't return this error it does nothing. you can just remove it.

sources/zh.zerobyw/src/html.rs
```diff
@@ -0,0 +1,238 @@
...
+            .ok_or_else(|| error!("Text not found"))?;
+        manga.cover = self
+            .try_select_first("img[src*='tupa.zerobyw33.com']")?
+            .attr("src");
+
+        let mut authors = Vec::new();
+        let mut tags = Vec::new();
+        let mut status = MangaStatus::Unknown;
+        for span in self.try_select("span.px-3.py-1.bg-gray-100")? {
+            if let Some(text) = span.text() {
```
kkantan: I think using a let else here is a little better than indenting the entire block another level.
```suggestion
            let Some(text) = span.text() else {
                continue;
            };
```

sources/zh.zerobyw/src/html.rs
```diff
@@ -0,0 +1,238 @@
...
+            last_zjid = new_zjid;
+            let zjid = new_zjid.to_string();
+            let url = Url::chapter(&zjid).to_string().ok();
+            chapters.push(Chapter {
+                key: zjid,
+                title,
+                url,
+                locked: true,
+                ..Default::default()
+            });
```
kkantan: instead of using `continue` I think having an else block makes more sense.

sources/zh.zerobyw/src/html.rs
```diff
@@ -0,0 +1,238 @@
...
+
+        for img in imgs {
+            let src = img.attr("src").unwrap_or_default();
+            let url = if src.starts_with("//") {
+                format!("https:{}", src)
+            } else {
+                src.to_string()
+            };
+            pages.push(Page {
+                content: PageContent::Url(url, None),
```
kkantan:
```suggestion
                content: PageContent::url(url),
```

sources/zh.zerobyw/src/net.rs
```diff
@@ -0,0 +1,210 @@
...
+        category_id: &str,
+        jindu: &str,
+        shuxing: &str,
+        order: &str,
+        dir: &str,
+        page: i32,
+    ) -> Self {
+        let mut q = QueryParameters::new();
+        if let Some(keyword) = keyword {
+            q.push_encoded("keyword", Some(keyword));
```
kkantan: this skips encoding, which may cause issues on certain ios versions if the the query contains spaces for example.

kkantan (changes requested):

sources/zh.zerobyw/src/html.rs
zyp-0: Thanks for the detailed explanation. I understand the tracker use case.

But for the actual existing chapters list,such as:

`1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 10不同, 11, 84, 85-83, 94-101, 102-111, 601, 602`

- Titles like 85-83 would be misparsed(extracted as 85 instead of 83) , and could conflict with another 85 chapter.
- Titles like 601 and 602 are often not actual chapter numbers — they might be extras or author notes.This is quite common on this website.
- The numbering across the site is inconsistent, mixing volumes and chapters with sudden jumps.

I'm not entirely sure how trackers work.
So if you think parsing the first number out of the string would be better, 
I'm happy to follow your suggestion and make the change. Just let me know your thoughts.

kkantan: I'm actually confused about what "85-83" actually contains, is it chapters 83, 84, and 85 combined? I think conflicts could occur if it's marked 85 and there's another chapter also marked 85, though I haven't seen duplicates like this in any of the series so I'm not sure if they exist. it would cause issues by being skipped while reading or marked read automatically, depending on some settings.

trackers will set the "last read chapter" after reading a chapter number, and can automatically mark items as read to sync with the current chapter count.

I would recommend using the numbers from the titles like I suggested, since even if it's not perfect, I believe the data is more accurate. however, if you think the site's numbering is too inconsistent and it's better to leave it, you can.

sources/zh.zerobyw/src/html.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+impl ChapterPage for Document {
+    fn pages(&self) -> Result<Vec<Page>> {
+        let mut pages = Vec::new();
+
+        let imgs = self.try_select("img.manga-image")?;
+
+        for img in imgs {
+            let src = img.attr("src").unwrap_or_default();
```
kkantan: I don't think we want an empty page url.
```suggestion
            let Some(src) = img.attr("src") else {
                continue;
            };
```

sources/zh.zerobyw/src/html.rs
```diff
@@ -0,0 +1,235 @@
...
+            let url = if src.starts_with("//") {
+                format!("https:{}", src)
+            } else {
+                src.to_string()
```
kkantan: isn't it already a `String`?
```suggestion
                src
```

kkantan (changes requested):

sources/zh.zerobyw/src/html.rs
zyp-0: I understand your point about using the title numbers to get closer to the actual chapter counts.

However, my main concern is about the inconsistency in the website's numbering — especially cases like 601 and 602. These are often not actual chapter numbers; they seem to be used by the site authors to distinguish extras, specials, or author notes. This kind of labeling is quite common on this site.

My worry is about tracker behavior — for example, on Bangumi (the tracker I use), unfinished manga often don't have an upper limit set for chapter numbers. If we extract 601 from a title like that, could it cause confusion or incorrect syncing? Would the tracker interpret it as "the user has read up to chapter 601", which would be wildly inaccurate?

I understand your preference for getting closer to the actual values, and I do agree that it's generally better to use the title numbers when they're reliable. If you think this concern is not a major issue, I'm happy to update the extraction logic in the next push and go with your suggestion.

kkantan: since you're against it we can just merge it for now, but if there are users that request it to be changed or express confusion about the numbering in the future we can adjust it.

kkantan (approved):

PR #589, opened 2026-06-23
fix(en.aquamanga): update url & v2

### LINKED ISSUE #571: Aquareader
Aquareader had its domain changed from "Aquareader.net" -> "Aquareader.org," thus causing everything to break.

I'll handle the maintenance of the source.

z3niith:
changed Aquareader's url changed from **aquareader.net** to [**aquareader.org**](https://aquareader.org)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "snyk.advanced.autoSelectOrganization": true
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "snyk.advanced.autoSelectOrganization": true
-}
--- a/sources/en.aquamanga/res/source.json
+++ b/sources/en.aquamanga/res/source.json
@@ -8,6 +8,6 @@
 		"languages": [
 			"en"
 		],
-    "minAppVersion": "0.7.1"
+		"minAppVersion": "0.7.1"
 	}
 }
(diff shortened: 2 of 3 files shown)


z3niith:
I am unsure as to what's causing the check to fail. Is there a minimum "api version"?

`defined minimum version accepts api version... no (api version: 0.7.1)`

.vscode/settings.json
kkantan, line 1: can this be excluded?

kkantan (changes requested):
in the `source.json` config info object, you can define `minAppVersion` as `0.7.1`.

.vscode/settings.json
z3niith: Yes, I forgot to ignore it

z3niith:
> in the `source.json` config info object, you can define `minAppVersion` as `0.7.1`.

Just to verify, I would add `minAppVersion` rather than `minApiVersion`?

z3niith:
> > in the `source.json` config info object, you can define `minAppVersion` as `0.7.1`.
> 
> Just to verify, I would add `minAppVersion` rather than `minApiVersion`?

Disregard, all tests have passed

sources/en.aquamanga/res/source.json
```diff
@@ -2,11 +2,12 @@
...
-		"version": 1,
-		"url": "https://aquareader.net",
+		"version": 2,
+		"url": "https://aquareader.org",
 		"contentRating": 1,
 		"languages": [
 			"en"
-		]
+		],
+    "minAppVersion": "0.7.1"
```
kkantan: this should use tabs instead of spaces.
```suggestion
		"minAppVersion": "0.7.1"
```

kkantan (changes requested):

kkantan (approved):

PR #592, opened 2026-06-26
fix(zh.happymh): update page decoding

### LINKED ISSUE #522: Happymh
漫画图片读取顺序错乱，所有页码的顺序都错了imageimage

### LINKED ISSUE #557: Happymh
unable to view all manhua from the source.
error: failed to load chapter. (other sources are working)

Image

avalanceonerd:
## Summary
- update the Happymh reading API request version and required analytics cookie
- decode the new encoded `data.scans` string format before parsing pages
- use raw deflate decompression for decoded scan payloads and strip image quality query parameters

## Verification
- Verified in Aidoku with `shaosong/6596192` and `shaosong/6632830`; both decoded scans successfully and returned 17 pages.
- Checked that `source.json` remains valid JSON with tab indentation and bumps Happymh from v5 to v6.
- `git diff --check` passes.

I could not run `cargo fmt`, `cargo clippy`, or `cargo check` locally after removing the temporary Rust toolchain from my machine; the PR workflow should run the source build/verification.

### DIFF
--- a/sources/zh.happymh/src/json/page_list.rs
+++ b/sources/zh.happymh/src/json/page_list.rs
@@ -179,7 +179,7 @@ fn sha256(data: &[u8]) -> [u8; 32] {
 }
 
 fn decode_hex(input: &str) -> Result<Vec<u8>> {
-	if input.len() % 2 != 0 {
+	if !input.len().is_multiple_of(2) {
 		bail!("Invalid hex string");
 	}
 	let mut out = Vec::with_capacity(input.len() / 2);
--- a/sources/zh.happymh/src/json/page_list.rs
+++ b/sources/zh.happymh/src/json/page_list.rs
@@ -1,19 +1,22 @@
 use crate::BASE_URL;
 use aidoku::{
 	Page, Result,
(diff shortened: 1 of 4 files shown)


sources/zh.happymh/src/json/page_list.rs
```diff
@@ -89,3 +102,100 @@ impl PageList {
...
+}
+
+fn sha256(data: &[u8]) -> [u8; 32] {
+	let mut hasher = Sha256::new();
+	hasher.update(data);
+	hasher.finalize().into()
+}
+
+fn decode_hex(input: &str) -> Result<Vec<u8>> {
+	if input.len() % 2 != 0 {
```
kkantan:
```suggestion
	if !input.len().is_multiple_of(2) {
```

kkantan (changes requested):

sources/zh.happymh/src/json/page_list.rs
avalanceonerd: Applied in 8bdff4e, thanks!

kkantan (approved):

PR #594, opened 2026-06-27
feat : added dankemoe and guya template

### LINKED ISSUE #411: danke.moe
This site was in the legacy source list and I used to use it but it seems like this source list doesn't have it:

https://danke.moe/

CtrlAltElite97:


sources/en.dankemoe/res/source.json
```diff
@@ -0,0 +1,19 @@
+{
+    "info": {
+        "id": "en.dankemoe",
+        "name": "Danke fürs Lesen",
+        "version": 1,
```
kkantan: I think we should replace the existing legacy source, since I believe the ids are the same.
```suggestion
        "id": "en.dankefurslesen",
        "name": "Danke fürs Lesen",
        "version": 2,
```

sources/en.dankemoe/res/source.json
```diff
@@ -0,0 +1,19 @@
...
+        "url": "https://danke.moe",
+        "minAppVersion": "0.7.1",
+        "contentRating": 1,
```
kkantan: given that there is an NSFW listing, I think it's better to classify this source as 18+.
```suggestion
        "contentRating": 2,
```

sources/en.dankemoe/res/source.json
```diff
@@ -0,0 +1,19 @@
...
+        "languages": ["en"]
+    },
+    "listings": [
+        { "id": "Latest" },
```
kkantan: I'm not sure this is necessary since the website doesn't have it, and there's already the "Latest Chapter" listing.
```suggestion
```

sources/en.dankemoe/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+impl Impl for DankeMoe {
+    fn new() -> Self {
+        set_rate_limit(2, 2, TimeUnit::Seconds);
+        Self
+    }
+
+    fn params(&self) -> Params {
+        Params {
+            base_url: "https://danke.moe",
+            content_rating: ContentRating::Suggestive,
```
kkantan: I'm not sure that it makes sense to apply this content rating to every series. I think it would be better to use Safe by default, and then detect the NSFW series and mark those, since the website appears to have a way to tell.

templates/guya/src/lib.rs
kkantan, line 1: the `Impl` struct is what should contain most of these functions, and the `Source` implementation for the template should only call the `Impl` functions, so that they can be overridden by any source using it. all the other templates should do this for reference.

templates/guya/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+
+mod helpers;
+mod models;
+
+pub use helpers::strip_html;
+pub use models::*;
+
+const PAGE_SIZE: usize = 20;
+
+const USER_AGENT: &str = "aidoku/1.0 CFNetwork/1490.0.4 Darwin/23.4.0";
```
kkantan: Is it necessary to have this full string? the legacy source simply uses "Aidoku".
```suggestion
const USER_AGENT: &str = "Aidoku/1.0 CFNetwork/1490.0.4 Darwin/23.4.0";
```

templates/guya/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+        }
+    }
+
+    let all_entries: Vec<Manga> = slugs
+        .into_iter()
+        .map(|slug| {
+            let url = format!("{base_url}/read/manga/{slug}/");
+            let (title, cover) = series_map
+                .get(&slug)
+                .map(|(t, c)| (t.clone(), c.clone()))
```
kkantan: you should be able to just consume the item from the map rather than cloning data out of it.
```suggestion
                .remove(&slug)
```

sources/en.dankefurslesen/res/filters.json
```diff
@@ -0,0 +1,9 @@
+[
+    {
+        "type": "select",
```
kkantan: there's a "sort" filter type you should use instead.

templates/guya/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+        }
+
+        if sort_latest {
+            all.sort_by_key(|(_, item)| core::cmp::Reverse(item.last_updated));
+        } else {
+            all.sort_by(|a, b| a.0.cmp(&b.0));
+        }
+
+        let start = ((page - 1) as usize) * PAGE_SIZE;
+        let has_next_page = start + PAGE_SIZE < all.len();
```
kkantan: if you're arbitrarily limiting the page size, it might make sense to cache the series list so moving on to the next page doesn't require a complete new fetch.

templates/guya/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+        Ok(MangaPageResult { entries, has_next_page })
+    }
+
+    fn get_manga_update(
+        &self,
+        mut manga: Manga,
+        needs_details: bool,
+        needs_chapters: bool,
+    ) -> Result<Manga> {
+        if needs_details || needs_chapters {
```
kkantan: this is always true, and thus unnecessary.

templates/guya/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+                    .filter(|(_, ch)| ch.is_public)
+                    .map(|(num_str, ch)| {
+                        let scanlators: Vec<String> = ch
+                            .groups
+                            .group_ids()
+                            .filter_map(|gid| det.groups.get(gid))
+                            .map(String::from)
+                            .collect();
+                        Chapter {
+                            key: num_str.clone(),
```
kkantan: you should define the url in a variable before creating the `Chapter`, so this doesn't need to be cloned.

templates/guya/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+                            chapter_number: num_str.parse().ok(),
+                            title: ch.title.clone().filter(|t| !t.is_empty()),
```
kkantan: does this need to be cloned?

sources/en.dankemoe/res/icon.png
kkantan, line 1: this image is better: https://github.com/Skittyblock/aidoku-community-sources/blob/main/src/rust/guya/sources/dankefurslesen/res/Icon.png.

kkantan (changes requested):

CtrlAltElite97:
sources/en.dankemoe/res/source.json:
- ID changed to en.dankefurslesen
- Version bumped to 2 (replaces legacy source)
- contentRating set to 2 (Erotica, reflects NSFW listings)
- Removed redundant "Latest" listing (superseded by "Latest Chapter")

sources/en.dankemoe/src/lib.rs:
- Removed ContentRating from imports (no longer used)
- Removed content_rating field from Params struct literal

templates/guya/src/lib.rs:
- Removed content_rating from Params struct — rating is no longer a static config value
- Added content_rating_for(&self, _det: &SeriesDetail) -> ContentRating to Impl trait with default ContentRating::Safe, making it overridable per-series in derived sources (e.g. a source could check tags/categories to detect NSFW)
- Removed unnecessary outer if needs_details || needs_chapters wrapper in get_manga_update
- Switched chapter iteration to core::mem::take(&mut det.chapters.0).into_iter() — takes owned entries so num_str.clone() and ch.title.clone() are no longer needed; URL is computed before num_str is moved into key

sources/en.dankefurslesen/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "dankemoe"
```
kkantan: it might be a good idea to match the source id here too, though I'm not sure how much it matters.
```suggestion
name = "dankefurslesen"
```

kkantan (commented):
it looks like you might've missed some comments from the previous review. if you go to the "files changed" tab you should be able to see them still attached to some files: `filters.json` and the guya `lib.rs`.

CtrlAltElite97:
sources/en.dankefurslesen/res/filters.json
Changed sort filter "type" from "select" → "sort" 
Replaced "ids": ["", "latest"] with "canAscend": false and "default": { "index": 0, "ascending": false }


templates/guya/Cargo.toml
Added spin = "0.10.0" dependency (needed for the RwLock/Once cache)

templates/guya/src/models.rs
 Added Clone to AllSeriesItem's derive — required so the cache can return a cloned copy of the series list on cache hits
 
 templates/guya/src/lib.rs — complete rewrite into a thin scaffold
 - USER_AGENT simplified from the verbose iOS string to "Aidoku"
- All logic removed — moved to new imp.rs
- Now only contains: Params, Guya<T>, SeriesCacheLock type alias, SERIES_CACHE static, series_cache() helper, and thin impl blocks that delegate every method to self.inner.method(&self.params, ...)
- Renamed _inner → inner (now actually used in the delegates)


templates/guya/src/imp.rs — new file
All logic from the old lib.rs moved here as default methods on the Impl trait, matching the madara/wpcomics template pattern so any source can override individual pieces

sources/en.dankefurslesen/res/filters.json
kkantan, line 1: json files should use tabs for indentation.

templates/guya/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+pub(crate) const PAGE_SIZE: usize = 20;
+pub(crate) const SERIES_TTL: i64 = 300; // 5 minutes
+
+type SeriesCacheLock = RwLock<Option<(Vec<(String, AllSeriesItem)>, i64)>>;
+
+static SERIES_CACHE: Once<SeriesCacheLock> = Once::new();
+
+pub(crate) fn series_cache() -> &'static SeriesCacheLock {
+    SERIES_CACHE.call_once(|| RwLock::new(None))
+}
```
kkantan: I think this would be better implemented with a `RefCell` attached to the `Guya` struct, which can be passed into `Impl` functions that need the cache. the `spin` dependency shouldn't be necessary. I believe the wpcomics template does this.

templates/guya/src/models.rs
```diff
@@ -0,0 +1,187 @@
...
+    #[serde(default, deserialize_with = "null_as_empty")]
+    pub cover: String,
+    #[serde(default)]
+    pub last_updated: i64,
+}
+
+impl AllSeriesItem {
+    pub fn into_manga(self, title: String, base_url: &str) -> Manga {
+        Manga {
+            key: self.slug.clone(),
```
kkantan: if you define the `url` before this, `slug` doesn't need to be cloned.

templates/guya/src/imp.rs
```diff
@@ -0,0 +1,364 @@
...
+            let desc = strip_html(&det.description);
+            if !desc.is_empty() {
+                manga.description = Some(desc);
+            }
+            if !det.author.is_empty() {
+                manga.authors = Some(vec![det.author.clone()]);
+            }
+            if !det.artist.is_empty() && det.artist != det.author {
+                manga.artists = Some(vec![det.artist.clone()]);
+            }
```
kkantan: do these need to be cloned?

kkantan (changes requested):

sources/en.dankefurslesen/res/filters.json
CtrlAltElite97: filters.json — converted spaces to tabs

sources/en.dankefurslesen/res/source.json
kkantan, line 1: this file needs to be indented with tabs too.

templates/guya/src/imp.rs
```diff
@@ -0,0 +1,361 @@
...
+
+    fn get_manga_list(&self, params: &Params, listing: Listing, page: i32, cache: &mut SeriesCache) -> Result<MangaPageResult> {
+        let base = params.base_url;
+        match listing.id.as_str() {
+            "Series"         => return self.fetch_html_series_list(params, "/series/", page, cache),
+            "Oneshot"        => return self.fetch_html_series_list(params, "/oneshots/", page, cache),
+            "NSFW"           => return self.fetch_html_series_list(params, "/nsfw/", page, cache),
+            "Latest Chapter" => return self.fetch_latest_chapters_list(params, page, cache),
+            _ => {}
+        }
```
kkantan, line 237: is this really the result after running `rustfmt`? I feel like I usually don't see this type of alignment with the default configuration.

templates/guya/src/imp.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        let mut all = self.fetch_all_series(params, cache)?;
+        match listing.id.as_str() {
+            "Latest" => all.sort_by_key(|(_, item)| Reverse(item.last_updated)),
```
kkantan: there is no more "Latest" listing.

templates/guya/src/imp.rs
```diff
@@ -0,0 +1,361 @@
...
+            _        => all.sort_by(|a, b| a.0.cmp(&b.0)),
```
kkantan: this appears to sort the "All" listing differently than on the website.

templates/guya/src/imp.rs
```diff
@@ -0,0 +1,361 @@
...
+                    entries,
+                    listing: Some(Listing {
+                        id: String::from("Latest"),
+                        name: String::from("Latest"),
+                        kind: ListingKind::Default,
+                    }),
+                },
+            }],
+        })
+    }
```
kkantan: I don't think this is necessary to have, since it's the same content as the All listing, and it isn't displayed as a scroller like this on the website.

sources/en.dankefurslesen/res/source.json
```diff
@@ -0,0 +1,18 @@
...
+        "url": "https://danke.moe",
+        "minAppVersion": "0.7.1",
+        "contentRating": 2,
+        "languages": ["en"]
+    },
+    "listings": [
+        { "id": "All" },
+        { "id": "Series" },
+        { "id": "Oneshot" },
+        { "id": "Latest Chapter" },
```
kkantan:
```suggestion
        { "id": "Oneshots" },
        { "id": "Latest Chapters" },
```

templates/guya/src/imp.rs
```diff
@@ -0,0 +1,361 @@
...
+        Ok(pages)
+    }
+
+    fn get_manga_list(&self, params: &Params, listing: Listing, page: i32, cache: &mut SeriesCache) -> Result<MangaPageResult> {
+        let base = params.base_url;
+        match listing.id.as_str() {
+            "Series"         => return self.fetch_html_series_list(params, "/series/", page, cache),
+            "Oneshot"        => return self.fetch_html_series_list(params, "/oneshots/", page, cache),
+            "NSFW"           => return self.fetch_html_series_list(params, "/nsfw/", page, cache),
+            "Latest Chapter" => return self.fetch_latest_chapters_list(params, page, cache),
```
kkantan: I'm not sure if this listing is actually necessary, since it appears to be identical to the All listing. unfortunately we can't display the actual chapter data such as title and number in the listing.

templates/guya/src/imp.rs
```diff
@@ -0,0 +1,361 @@
...
+use crate::{AllSeriesItem, SeriesCache, SeriesDetail, strip_html};
+use super::{PAGE_SIZE, Params, SERIES_TTL};
+
+pub trait Impl {
+    fn new() -> Self;
+    fn params(&self) -> Params;
+
+    fn content_rating_for(&self, _det: &SeriesDetail) -> ContentRating {
+        ContentRating::Safe
+    }
```
kkantan, line 24: the `en.dankefurslesen` source doesn't implement this, so nsfw items are marked incorrectly. is it possible to implement this? the website has a warning when you open the series page.

kkantan (changes requested):

templates/guya/src/imp.rs
CtrlAltElite97: Removed get_home from Impl trait, removed Home impl from lib.rs, removed Home from register_source!

templates/guya/src/imp.rs
CtrlAltElite97: No, that was manual alignment on my part. Removed the extra padding — match arms now follow standard rustfmt style

templates/guya/src/imp.rs
CtrlAltElite97: removed that dead arm along with the inner match entirely since it was the only other case.

templates/guya/src/imp.rs
CtrlAltElite97: Fixed by removing the explicit sort. fetch_all_series returns entries in BTreeMap iteration order (alphabetical by title), which is the natural API ordering — no re-sorting needed

templates/guya/src/imp.rs
CtrlAltElite97: Kept it — "Latest Chapters" scrapes /latest_chapters/ which returns series ordered by most-recently-updated chapter, whereas "All" is alphabetical. They surface different orderings. The naming should now make the intent clearer ("Latest Chapters" vs "All"). Agreed we can't show chapter metadata in the listing row — that's a platform limitation.

templates/guya/src/imp.rs
CtrlAltElite97: Added adult: bool (serde default false) to SeriesDetail — the guya API includes this field on the series detail endpoint. DankeMoe now overrides content_rating_for to return Erotica when adult is true, Safe otherwise

CtrlAltElite97:
- Convert source.json indentation to tabs; bump version to 3
- Rename listings: "Oneshot" → "Oneshots", "Latest Chapter" → "Latest Chapters"
- Remove manual match arm alignment; let rustfmt format naturally
- Remove dead "Latest" match arm and redundant alphabetical sort in get_manga_list
- Remove get_home from Impl trait and Home impl from template and register_source!
- Add adult: bool to SeriesDetail; implement content_rating_for in DankeMoe

templates/guya/src/imp.rs
kkantan: alphabetical order isn't what the website displays though. I believe it should be sorting by last updated, like you had for the latest listing. it's the same order are the "Latest Chapters" listing.

kkantan (changes requested):
it appears that the content rating is not displayed for NSFW titles in the app, but I am not sure if this is a bug with the app since there is no publication status provided, or if the api isn't returning proper data. it is probably okay to leave it as-is.

templates/guya/src/imp.rs
CtrlAltElite97: The "Latest Chapters" listing degrades to just a series list ordered by recency — and that recency ordering is already available via the filters.json sort option ("Latest Chapter"). So it's redundant. Let me remove it

kkantan:
why was this closed? it was almost done, apart from the all listing sort being incorrect.

CtrlAltElite97:
> why was this closed? it was almost done, apart from the all listing sort being incorrect.

Sorry for the abrupt close — that was on me, not a sign the work was abandoned. You were right that it was basically there.

I've fixed the last open item: the "All" listing now sorts by last_updated descending instead of alphabetical, so it matches the website's ordering (same order as the "Latest Chapter" sort). The default arm of get_manga_list in the guya template now does:

let mut all = self.fetch_all_series(params, cache)?;
all.sort_by_key(|(_, item)| Reverse(item.last_updated));

en.dankefurslesen inherits it through the template, and it builds clean against wasm32-unknown-unknown.

kkantan:
you say that but it's still closed and combined with your kagane pr?

PR #597, opened 2026-06-27
fix(en.freewebnovel): load full paginated chapter list

### LINKED ISSUE #600: freewebnovel doesn't work
when you open any novel from freewebnovel on aidoku, you can only see up to 40 chapters.

MrGcGamer:
The site now paginates chapter lists, so only the first 40 chapters loaded. Fetch all pages via ?page=N&pageSize=200, with 429 rate-limit retry handling.

### DIFF
--- a/sources/en.freewebnovel/src/helpers.rs
+++ b/sources/en.freewebnovel/src/helpers.rs
@@ -5,16 +5,44 @@ use aidoku::{
 	imports::{
 		html::{Document, Element, Kind},
 		net::Request,
+		std::sleep,
 	},
 	prelude::*,
 };
 
+/// Fetches a URL as an HTML document, transparently retrying on HTTP 429.
+///
+/// The site rate-limits bursts (~15 requests) with a `429 Too Many Requests`
+/// and a `Retry-After` header (typically 10s). Fetching a long novel's chapter
+/// list needs many sequential page requests, so we honor `Retry-After` and
+/// retry rather than silently dropping chapters.
 pub fn request_html(url: &str) -> Result<Document> {
-	Ok(Request::get(url)?.html()?)
+	const MAX_RETRIES: u32 = 3;
+	const DEFAULT_RETRY_AFTER: i32 = 10;
+	let mut attempt = 0;
+	loop {
+		let response = Request::get(url)?.send()?;
+		if response.status_code() == 429 && attempt < MAX_RETRIES {
+			let wait = response
+				.get_header("retry-after")
+				.and_then(|v| v.parse::<i32>().ok())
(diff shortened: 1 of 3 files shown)


sources/en.freewebnovel/src/helpers.rs
```diff
@@ -5,16 +5,44 @@ use aidoku::{
...
 		html::{Document, Element, Kind},
 		net::Request,
+		std::sleep,
 	},
 	prelude::*,
 };
 
+/// Fetches a URL as an HTML document, transparently retrying on HTTP 429.
+///
+/// The site rate-limits bursts (~15 requests) with a `429 Too Many Requests`
```
kkantan: does it work to use the `set_rate_limit` function?

sources/en.freewebnovel/src/helpers.rs
```diff
@@ -93,39 +121,72 @@ pub fn content_rating_from_tags(tags: &[String]) -> ContentRating {
...
+				.trim()
+				.strip_prefix(':')
+				.or_else(|| rest.trim().strip_prefix('-'))
+				.map_or_else(|| rest.trim().to_string(), |t| t.trim().to_string()),
+			None => title,
+		};
+	};
+	Some(Chapter {
+		key: chapter_key,
+		title: { if !title.is_empty() { Some(title) } else { None } },
```
kkantan: is the block surrounding this necessary?
```suggestion
		title: if !title.is_empty() { Some(title) } else { None },
```

sources/en.freewebnovel/src/helpers.rs
```diff
@@ -93,39 +121,72 @@ pub fn content_rating_from_tags(tags: &[String]) -> ContentRating {
...
-			Some(Chapter {
-				key: chapter_key,
-				title: { if !title.is_empty() { Some(title) } else { None } },
-				chapter_number,
-				url: Some(url),
-				..Default::default()
-			})
-		})
-		.collect()
+	items.filter_map(|item| chapter_from_item(&item)).collect()
```
kkantan: rather than collecting this into a vector, I wonder if you could just return an iterator? that, or just use it inline in the `fetch_all_chapters` instead of breaking it into a separate function. it would be better to `chain` on to the iterator instead of using `extend` on a vec, and better to `rev` the iterator before collecting instead of `reverse` the vec.

kkantan (changes requested):

sources/en.freewebnovel/src/helpers.rs
MrGcGamer: it does work, but feels a bit slower

kkantan (approved):

PR #602, opened 2026-06-29
fix(en.freewebnovel): load all chapters via pagination

DedPiht0-MGN:
## Firstly
This is completely vibe-coded. I made this PR for anyone who wants to use this source right away.

## What
FreeWebNovel only loaded the first 40 chapters of any novel. The novel detail
page renders just page 1 (40 chapters) of the chapter list; the remaining
chapters are served on demand through an AJAX endpoint.

`extract_chapters` now paginates `/novel/{slug}?ajax=chapters&page=N&pageSize=200`
(header `X-Requested-With: XMLHttpRequest`), reads the authoritative `totalPage`
from the first JSON response, and parses each returned `<li>` HTML fragment — so
the complete chapter list is returned, newest chapter first.

## Why
Reported bug #600: only the first 40 chapters appear. Root cause is the site's
server-side pagination of the chapter list (≤200 per page).

## Commits
- **fix:** load all chapters via the AJAX pagination endpoint.
- **refactor:** simplify the chapter parsing (no behavior change) — rewrite
  `parse_chapter_number` with `strip_prefix`/`find`/`parse`, extract a testable
  `strip_chapter_label` helper, pre-size the `Vec` from `totalChapters`, and add
  hermetic unit tests for the pure helpers.

## Notes for reviewer
- Added the `aidoku` `json` feature + `serde` to deserialize the AJAX response.
- `get_manga_update` now fetches the detail page only when `needs_details`,
  avoiding a wasted request on chapter-only refreshes.
- Bumped `source.json` version 2 → 3.
- Chapter pages are fetched **sequentially on purpose**: the site (Cloudflare)
  rate-limits bursts of parallel requests, so `send_all` was avoided.
- Tests hit the live site — run `cargo test -- --test-threads=1` (parallel runs
  trip the rate-limiting and fail with `RequestError`). All 8 pass; the test
  novel now returns 809 chapters instead of 40.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sources/en.freewebnovel/Cargo.lock
+++ b/sources/en.freewebnovel/Cargo.lock
@@ -14,6 +14,7 @@ dependencies = [
  "paste",
  "postcard",
  "serde",
+ "serde_json",
  "talc",
  "thiserror",
 ]
@@ -108,6 +109,7 @@ version = "0.1.0"
 dependencies = [
  "aidoku",
  "aidoku-test",
+ "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


kkantan:
this was already fixed by #597.

PR #603, opened 2026-07-02
feat: add multi.kagane

### LINKED ISSUE #207: kagane.org
https://kagane.org/

CtrlAltElite97:


sources/en.kagane/res/filters.json
```diff
@@ -0,0 +1,24 @@
...
+		"title": "Status",
+		"options": ["Ongoing", "Completed", "Hiatus", "Cancelled"],
+		"ids": ["Ongoing", "Completed", "Hiatus", "Abandoned"]
+	},
+	{
+		"type": "multi-select",
+		"id": "format",
+		"title": "Format",
+		"options": ["Manga", "Manhwa", "Manhua", "Comic", "Other"],
+		"ids": ["Manga", "Manhwa", "Manhua", "Comic", "Other"]
```
kkantan: providing `ids` is not necessary when they're the same as the `options`.

sources/en.kagane/res/settings.json
```diff
@@ -0,0 +1,26 @@
...
+	{
+		"type": "group",
+		"title": "Content",
+		"items": [
+			{
+				"type": "multi-select",
+				"key": "contentRating",
+				"title": "Content Rating",
+				"values": ["Safe", "Suggestive", "Erotica", "Pornographic"],
+				"titles": ["Safe", "Suggestive", "Erotica", "Pornographic"],
```
kkantan: titles default to values so this isn't necessary.
```suggestion
```

sources/en.kagane/res/source.json
```diff
@@ -0,0 +1,19 @@
+{
+	"info": {
+		"id": "en.kagane",
+		"name": "Kagane",
+		"version": 3,
+		"url": "https://kagane.to",
+		"minAppVersion": "0.7.1",
+		"contentRating": 2,
+		"languages": ["en"]
```
kkantan: the website appears to support many languages, not just english.

sources/en.kagane/res/source.json
```diff
@@ -0,0 +1,19 @@
+{
+	"info": {
+		"id": "en.kagane",
+		"name": "Kagane",
+		"version": 3,
```
kkantan: why version 3?

sources/en.kagane/res/source.json
```diff
@@ -0,0 +1,19 @@
...
+		"url": "https://kagane.to",
+		"minAppVersion": "0.7.1",
+		"contentRating": 2,
```
kkantan: nsfw content is filtered out by default so it's fine to mark the source as only containing nsfw, not primarily nsfw.
```suggestion
		"contentRating": 1,
```

sources/multi.kagane/src/helpers.rs
```diff
@@ -0,0 +1,113 @@
...
+	let lower = s.map(|s| s.to_lowercase());
+	match lower.as_deref() {
+		Some("safe") => ContentRating::Safe,
+		Some("suggestive") => ContentRating::Suggestive,
+		Some("erotica") | Some("pornographic") => ContentRating::NSFW,
+		_ => ContentRating::Suggestive,
+	}
+}
+
+pub fn build_search_body(query: Option<&str>, statuses: &[String], formats: &[String]) -> String {
```
kkantan: it seems like this code could be simplified with `serde_json`.

sources/multi.kagane/src/lib.rs
```diff
@@ -0,0 +1,357 @@
...
+						entries: latest,
+						listing: Some(Listing {
+							id: String::from("Latest"),
+							name: String::from("Latest"),
+							kind: ListingKind::Default,
+						}),
+					},
+				},
+			],
+		})
```
kkantan, line 345: could we mirror the website with "Popular Today" and "Recently Added" scrollers and a "Recently Updated" manga chapter list?

sources/en.kagane/res/source.json
```diff
@@ -0,0 +1,19 @@
...
+		"languages": ["en"]
+	},
+	"listings": [
+		{ "id": "Popular Today" },
+		{ "id": "Popular This Week" },
+		{ "id": "Popular This Month" },
+		{ "id": "Popular All Time" },
+		{ "id": "Latest" },
+		{ "id": "Recently Added" }
+	]
```
kkantan: I'm not sure if these listings are necessary, since the site does not have them and they're just the search page with a sort filter.

kkantan (changes requested):

sources/en.kagane/res/filters.json
CtrlAltElite97: removed the ids from the format filter. Left them on status since Cancelled maps to Abandoned on the API side.

sources/en.kagane/res/settings.json
CtrlAltElite97: Done, dropped titles from both contentRating and sourceType. Kept them on the new language setting though, since the API wants codes (en, pt-br, …) but I wanted to show proper names in the UI

sources/en.kagane/res/source.json
CtrlAltElite97: No real reason, bumped it by accident during testing. Reset to 1.

sources/en.kagane/res/source.json
CtrlAltElite97: You're right. I checked the catalog and it serves 15+ languages. Added a "Languages" multi-select in the settings that feeds content_lang, defaulting to English, so users can pull whichever languages they want instead of being locked to en.

sources/en.kagane/res/source.json
CtrlAltElite97: Changed contentRating to 1.

sources/multi.kagane/src/helpers.rs
CtrlAltElite97: Reworked build_search_body to construct a serde_json map and serialize it, which let me delete the manual escape_json helper entirely. Much cleaner.

sources/multi.kagane/src/lib.rs
CtrlAltElite97: Done. Home is now two scrollers (Popular Today / Recently Added) plus a Recently Updated MangaChapterList. The search response already includes a latest_chapters field, so the chapter list shows the actual latest chapter/volume and upload date.

sources/en.kagane/res/source.json
CtrlAltElite97: Agreed, they were redundant with the sort filter. Removed the listings block. The home sections still link through to their sorted views via ListingProvider.

CtrlAltElite97:
- source.json: reset version to 1, lower contentRating to Suggestive,
  and drop the redundant listings block (duplicated the sort filter)
- filters.json: remove redundant `ids` on the format filter
- settings.json: drop redundant `titles`; add a Languages multi-select
  feeding content_lang so the source isn't locked to English
- helpers: rebuild the search body with serde_json and delete the
  hand-rolled escape_json helper
- home: mirror the site with Popular Today / Recently Added scrollers

sources/multi.kagane/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+		.header("Content-Type", "application/json")
+		.header("Origin", BASE_URL)
+		.header("Referer", &format!("{BASE_URL}/"))
+		.body(body))
+}
+
+/// Build a home-feed entry pairing a series with its most recent chapter,
+/// as returned in the search endpoint's `latest_chapters` field. Skips
+/// series that have no chapters.
+fn to_manga_with_chapter(mut item: SearchItem) -> Option<MangaWithChapter> {
```
kkantan: this function should probably be put in `helpers.rs`, but I also think it could be an implementation of the `TryFrom` trait put into `models.rs` instead.

sources/multi.kagane/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+					None
+				} else {
+					Some(String::from(t))
+				}
+			}),
+			date_uploaded: book.created_at.as_deref().and_then(|s| {
+				let s = s.split_once('.').map_or(s, |(b, _)| b);
+				parse_date(format!("{s}Z"), "yyyy-MM-dd'T'HH:mm:ss'Z'")
+			}),
+			url: Some(url),
```
kkantan: this field isn't necessary for the home page data.

sources/multi.kagane/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+		&self,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		let mut sort_idx = 0usize;
+		let mut statuses: Vec<String> = Vec::new();
+		let mut formats: Vec<String> = Vec::new();
+
+		for filter in &filters {
```
kkantan: I believe this should make it so you don't have to clone.
```suggestion
		for filter in filters {
```

sources/multi.kagane/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+				manga.authors = Some(authors);
+			}
+			if !artists.is_empty() {
+				manga.artists = Some(artists);
+			}
+
+			let tags: Vec<String> = det.genres.into_iter().map(|g| g.genre_name).collect();
+			if !tags.is_empty() {
+				manga.tags = Some(tags);
+			}
```
kkantan: the website also has "tags" that are separate from genres. is it possible to add those to the tags list as well?

sources/multi.kagane/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+						} else {
+							Some(scanlators)
+						},
+						url: Some(url),
+						..Default::default()
+					}
+				})
+				.collect();
+			// API returns chapters oldest-first; reverse for newest-first display.
+			chapters.reverse();
```
kkantan: you should use `rev` on the iter instead.

sources/multi.kagane/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+use models::*;
+
+const BASE_URL: &str = "https://kagane.to";
+const API_BASE: &str = "https://yuzuki.kagane.to/api/v2";
+
+// Sort param by filter index (from filters.json options order).
+const SORT_PARAMS: &[&str] = &[
+	"total_views", // 0: Popular (Total Views)
+	"updated_at",  // 1: Latest Updated
+	"series_name", // 2: By Name
```
kkantan: this doesn't appear to be a valid sort method. at least, it is not ordering results by name. the website also doesn't have this option. could you match the options from the website?

sources/multi.kagane/res/filters.json
```diff
@@ -0,0 +1,23 @@
...
+		"id": "sort",
+		"title": "Sort By",
+		"canAscend": false,
+		"options": ["Popular (Total Views)", "Latest Updated", "By Name", "Newest"],
+		"default": { "index": 0, "ascending": false }
+	},
+	{
+		"type": "multi-select",
+		"id": "status",
+		"title": "Status",
```
kkantan:
```suggestion
		"title": "Upload Status",
```

sources/multi.kagane/res/settings.json
```diff
@@ -0,0 +1,33 @@
...
+		"items": [
+			{
+				"type": "multi-select",
+				"key": "languages",
+				"title": "Languages",
+				"values": ["en", "es", "it", "fr", "tr", "pt", "pt-br", "pl", "ru", "de", "id", "vi", "th", "ar", "he"],
+				"titles": ["English", "Spanish", "Italian", "French", "Turkish", "Portuguese", "Portuguese (Brazil)", "Polish", "Russian", "German", "Indonesian", "Vietnamese", "Thai", "Arabic", "Hebrew"],
+				"default": ["en"],
+				"refreshes": ["content"]
+			},
```
kkantan: this should use the languages setting in `source.json`. see the mangadex source for an example if necessary.

sources/multi.kagane/res/source.json
```diff
@@ -0,0 +1,11 @@
+{
+	"info": {
+		"id": "en.kagane",
```
kkantan:
```suggestion
		"id": "multi.kagane",
```

kkantan (changes requested):
it would be nice to have `supportsTagSearch` implemented. author search is also supported on the website. another optional thing to have would be the `AlternateCoverProvider` trait, but that is definitely not necessary.

sources/multi.kagane/src/lib.rs
CtrlAltElite97: I went with the TryFrom route. It's now an impl TryFrom<SearchItem> for MangaWithChapter in models.rs, and the home feed creates its entries through it. Cleaner than having the helper sitting in lib.rs.

sources/multi.kagane/src/lib.rs
CtrlAltElite97: You're right, it wasn't doing anything useful there. I dropped the chapter url from the home entries entirely.

sources/multi.kagane/src/lib.rs
CtrlAltElite97: Done — switched the loop to take ownership with for filter in filters, so the included/excluded lists get moved instead of cloned.

sources/multi.kagane/src/lib.rs
CtrlAltElite97: Yep, added. The detail response carries both genres and tags, so I now merge them into manga.tags. I also skip any entry flagged is_spoiler so spoiler tags don't get surfaced.

sources/multi.kagane/src/lib.rs
CtrlAltElite97: the chapter list is now built with series_books.into_iter().rev() rather than collecting first and calling .reverse().

sources/multi.kagane/src/lib.rs
CtrlAltElite97: Fixed, and it turned out to be a bigger issue than just series_name. I tested every sort against the live API and found series_name, title, rating, and avg_views_day all silently fall back to default ordering. So the options are now Popular (Total Views) / Recently Updated / Recently Added — the ones that actually sort. That also fixed a hidden bug: "Popular Today" on the home and listing was using the dead avg_views_day, so I moved it to avg_views_week.

sources/multi.kagane/res/filters.json
CtrlAltElite97: Done

sources/multi.kagane/res/settings.json
CtrlAltElite97: Done — removed the custom languages multi-select from settings. source.json now lists all 15 supported languages, and helpers::languages() reads the app's built-in language selection (mapping pt-BR→pt-br for the API), same as MangaDex

sources/multi.kagane/res/source.json
CtrlAltElite97: Done — the id is multi.kagane, and I also removed the leftover en-only languages value.

CtrlAltElite97:
> it would be nice to have `supportsTagSearch` implemented. author search is also supported on the website. another optional thing to have would be the `AlternateCoverProvider` trait, but that is definitely not necessary.

Implemented. I added a Genre multi-select with all 66 genres and their UUIDs (with exclude support), and wired the search body to the API's real shape — genres: { values, exclude, match_all: true }. config.supportsTagSearch is on, so tapping a genre on a series page runs a genre search too.

I dug into this one but couldn't get it working cleanly. The search endpoint only filters authors via a UUID-keyed staff field, and there's no author-lookup endpoint to convert a typed name into a UUID — so it can't work the way MangaDex's author search does. I've left it out for now; happy to add it if you know of an endpoint I missed.

Skipped AlternateCoverProvider for now since it's not necessary

sources/multi.kagane/res/source.json
```diff
@@ -0,0 +1,30 @@
...
+			"id",
+			"vi",
+			"th",
+			"ar",
+			"he"
+		]
+	},
+	"config": {
+		"supportsTagSearch": true
+	}
```
kkantan: actually I don't think setting this to `true` is necessary since the genre filter is already marked `"isGenre": true` in `filters.json`.
```suggestion
	}
```

sources/multi.kagane/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+		.header("Referer", &format!("{BASE_URL}/")))
+}
+
+fn api_post(url: &str, body: String) -> Result<Request> {
+	Ok(Request::post(url)?
+		.header("Content-Type", "application/json")
+		.header("Origin", BASE_URL)
+		.header("Referer", &format!("{BASE_URL}/"))
+		.body(body))
+}
```
kkantan: I think putting these functions in `helpers.rs` would make more sense, because they are helper functions.

sources/multi.kagane/src/lib.rs
```diff
@@ -0,0 +1,397 @@
...
+		Ok(MangaPageResult {
+			entries,
+			has_next_page,
+		})
+	}
+}
+
+impl Home for Kagane {
+	fn get_home(&self) -> Result<HomeLayout> {
+		let pop_url = format!("{API_BASE}/search/series?page=0&size=20&sort=avg_views_week,desc");
```
kkantan, line 279: this looks like the popular listing for this week, rather than today. you should either change the component title, or the sort option here.

kkantan (changes requested):

sources/multi.kagane/src/lib.rs
CtrlAltElite97: Moved api_get and api_post functions from lib.rs to helpers.rs

sources/multi.kagane/src/lib.rs
CtrlAltElite97: Renamed "Popular Today" → "Popular This Week" to accurately reflect the avg_views_week sort being used

sources/multi.kagane/res/source.json
CtrlAltElite97: Removed "supportsTagSearch"

kkantan (approved):

PR #629, opened 2026-07-19
feat(multi.suwayomi): add ui_login support and authenticated image loading

nathanjace:
Alt Text

v5 implements Suwayomi-server's JWT-based ui_login auth mode and fixes image loading when authentication is enabled on the server.
- Auth: Added ui_login option to settings and bumped version to 5.
- Token Management: Added storage helpers for access/refresh tokens.
- Implemented silent token refresh by base64url decoding the JWT payload to check the exp claim.
- Image Loading: Implemented ImageRequestProvider to append the appropriate auth headers (Basic or Bearer) to cover thumbnails and chapter pages.
- Robustness: If a request fails, the client clears the cached token and attempts a silent re-login/refresh once.
- Tests: Added unit tests to verify the JWT expiration parsing logic.

Tested compilation with aidoku-cli and ran tests via host target compilation.

### DIFF
--- a/sources/multi.suwayomi/src/lib.rs
+++ b/sources/multi.suwayomi/src/lib.rs
@@ -11,30 +11,43 @@ use crate::models::{
 	FetchChapterPagesResponse, GraphQLResponse, MangaOnlyDescriptionResponse, MultipleCategories,
 	MultipleChapters, MultipleMangas,
 };
-use aidoku::imports::std::send_partial_result;
+use aidoku::imports::std::{current_date, send_partial_result};
 use aidoku::{
 	AidokuError, BaseUrlProvider, BasicLoginHandler, Chapter, DynamicListings, FilterValue,
-	Listing, ListingProvider, Manga, MangaPageResult, Page, PageContent, Result, Source,
+	ImageRequestProvider, Listing, ListingProvider, Manga, MangaPageResult, Page, PageContent, PageContext,
+	Result, Source,
 	alloc::{String, Vec},
 	imports::net::Request,
 	prelude::*,
 };
 use alloc::string::ToString;
 use alloc::vec;
-use base64::{Engine, engine::general_purpose::STANDARD};
+use base64::{Engine, engine::general_purpose::{STANDARD, URL_SAFE_NO_PAD}};
 
 struct Suwayomi;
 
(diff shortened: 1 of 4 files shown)


sources/multi.suwayomi/src/lib.rs
```diff
@@ -68,6 +84,95 @@ impl Suwayomi {
...
+
+		let resp = Request::post(format!("{base_url}/api/graphql"))?
+			.header("Content-Type", "application/json")
+			.body(mutation.to_string())
+			.json_owned::<GraphQLResponse<serde_json::Value>>()?;
+
+		let login_payload = resp
+			.data
+			.get("login")
+			.ok_or_else(|| aidoku::error!("Missing login payload"))?;
```
kkantan: namespacing the macro calls shouldn't be required since `aidoku::prelude::*` is imported.
```suggestion
			.ok_or_else(|| error!("Missing login payload"))?;
```

sources/multi.suwayomi/src/lib.rs
```diff
@@ -68,6 +84,95 @@ impl Suwayomi {
...
+			.ok_or_else(|| aidoku::error!("Missing accessToken"))?;
+
+		settings::set_access_token(access_token);
+		Ok(access_token.into())
+	}
+
+	fn get_valid_access_token(&self, base_url: &str) -> Result<String> {
+		if settings::get_credentials().is_none() {
+			settings::clear_tokens();
+			return Err(aidoku::error!("Not authenticated"));
```
kkantan:
```suggestion
			bail!("Not authenticated");
```

kkantan (changes requested):

kkantan (changes requested):
my previous comments were just single instances. the same things should be changed everywhere.

nathanjace:
Working on it… I’m also trying to figure out how to mark chapters as read / unread to sync with the suwayomi instance. There’s an extension for paperback that is able to do so, just trying to figure it out for this

kkantan:
that isn't possible with aidoku's source api. however, it looks like aidoku is including a built-in suwayomi source with a tracker in the next update, so I wouldn't worry about it too much: https://github.com/Aidoku/Aidoku/pull/1019.

PR #634, opened 2026-07-21
feat(multi.mangadotnet): Add recommendations and bookmarks

jianmingyong:
This is an initial draft to add logged in features such as showing recommendations and users bookmarks directly in home page.

The CF workaround is marked as obsolete and will be removed in the future. That is no longer needed once user updates the app.

I will later downgrade the target if I don’t update to async js. Well technically speaking this source doesn’t need it than comix.

Some refactoring to clean up shared portions of the code.

I am also moving some parts to use parallel fetching to speed things up.


sources/multi.mangadotnet/res/source.json
```diff
@@ -2,13 +2,13 @@
...
 		"name": "Mangadotnet",
-		"version": 3,
+		"version": 4,
 		"url": "https://mangadot.net",
 		"contentRating": 1,
 		"languages": [
 			"multi"
 		],
-		"minAppVersion": "0.8.3"
+		"minAppVersion": "0.8.4"
```
kkantan, line 11: if you're setting the min version to the latest here, you could just remove the cloudflare feature altogether since it's not necessary on this version.

jianmingyong:
Static filter is now implemented. The script can also regenerate the list though I doubt there will be any changes as it is synced with mangabaka.

Tags search is not implemented... well idk if aidoku really could handle that many tags though... 2314 of them. Yea it would be overkill to implement that though there is a search feature.

Collections recommendation is being planned, but I think this is a little too much to include.

Overall, this big update will make the source even better ~

sources/multi.mangadotnet/src/lib.rs
```diff
@@ -380,6 +408,18 @@ impl Home for Mangadotnet {
 			],
 		}));
 
+		let is_logged_in = self.is_logged_in();
+
+		if is_logged_in {
+			send_partial_result(&HomePartialResult::Layout(HomeLayout {
```
kkantan: it might make sense to conditionally append this component to the vec that gets sent in the above `send_partial_result` so you reduce the ffi call by one, but I am not sure how much difference it would make.

sources/multi.mangadotnet/src/lib.rs
```diff
@@ -448,6 +488,34 @@ impl Home for Mangadotnet {
...
+		if is_logged_in {
+			let mut query_parameters = QueryParameters::new();
+
+			if !hide_nsfw() {
+				query_parameters.push("adult", Some("both"));
+			} else {
+				query_parameters.push("adult", Some("0"));
+			}
+
+			query_parameters.push("limit", Some("21"));
```
kkantan: I think using `QueryParameters` here is adding complexity, since it's only two parameters that both don't need encoding, but it's up to you.
```rs
format!("adult={}&limit=21", if hide_nsfw() { "0" } else { "both" });
```

jianmingyong:
Added more official groups and also personal group preference list (that user can add in settings)

I think this should be more or less what I am going to add for this PR.

I will reserve some of the other minor fun stuff for next version so it is easier to review. (I just have some silly ideas what I wanted to add to improve user experience)

sources/multi.mangadotnet/src/helpers.rs
```diff
@@ -103,6 +211,30 @@ fn is_official_like(chapter: &MangaChapter) -> bool {
...
+			.position(|p| p == &format!("{id}"))
+	});
+	index.push(group_id);
+
+	if let Some(groups) = chapter.groups.as_ref() {
+		groups.iter().for_each(|g| {
+			index.push(
+				deduped_group_list
+					.iter()
+					.position(|p| p == &format!("{}", g.id)),
```
kkantan: are the `format!`s necessary here?

jianmingyong: Technically I guess it is more or less the same with the other solution due to it being a string comparison rather than i32.

For now, it will be `.position(|p| p.eq(&id.to_string()))` and `.position(|p| p.eq(&g.id.to_string())),`
The custom group list will get better improvement to support names as well which was left out in this PR.

kkantan (approved):
I won't block the merge of this since it was already present before, but I noticed that search results are returning descriptions and tags before `get_manga_update`. this is usually not a good idea because the app starts lagging if the user scrolls for a while due to holding so much data.

jianmingyong:
> I won't block the merge of this since it was already present before, but I noticed that search results are returning descriptions and tags before `get_manga_update`. this is usually not a good idea because the app starts lagging if the user scrolls for a while due to holding so much data.

I don’t know about this gimmick with search results. Since the source provided them out front, I added them in. Well, I could somewhat isolate the api I guess to truncate it.

PR #637, opened 2026-07-21
feat: add pt.hqnow

ramonhpr:
Adding support to source hq-now for language pt-BR
The code was mainly written by AI taking as base the resource in the extensions repository: https://github.com/keiyoushi/extensions-source/blob/main/src/pt/hqnow/src/eu/kanade/tachiyomi/extension/pt/hqnow/HQNow.kt 
And was tested in rakuyomi project in a kindle 12th generation device.


sources/pt.hqnow/res/source.json
kkantan, line 1: indentation for json files should use tabs.

sources/pt.hqnow/res/source.json
```diff
@@ -0,0 +1,10 @@
+{
+  "info": {
+    "id": "pt.hqnow",
+    "name": "HQ-now",
```
kkantan:
```suggestion
    "name": "HQ Now",
```

sources/pt.hqnow/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+	prelude::*,
+};
+use alloc::collections::BTreeMap;
+use alloc::string::ToString;
+use core::cell::RefCell;
+use serde::de::DeserializeOwned;
+
+const GRAPHQL_URL: &str = "https://admin.hq-now.com/graphql";
+
+// Maps publisher display name (as it appears in filters.json) to editoraId.
```
kkantan: you can set the ids in `filters.json` instead.

sources/pt.hqnow/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+	("Valiant Comics", 48),
+	("Vents d'Ouest", 32),
+	("Vertigo", 13),
+	("Virgin Comics", 47),
+	("WildStorm", 2),
+];
+
+// ── GraphQL helper ────────────────────────────────────────────────────────────
+
+fn execute_query<T: DeserializeOwned>(
```
kkantan: the helper functions you have here and in `models.rs` could be moved to a `helpers.rs` file.

sources/pt.hqnow/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+	let mut body = serde_json::json!({
+		"operationName": gql.operation_name,
+		"query": gql.query,
+	});
+	if let Some(vars) = variables {
+		body["variables"] = vars;
+	}
+	let body_str = body.to_string();
+	let resp = Request::post(GRAPHQL_URL)
+		.map_err(|_| AidokuError::message("network error"))?
```
kkantan: there's a shorthand macro for message errors.
```suggestion
		.map_err(|_| error!("network error"))?
```

sources/pt.hqnow/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+				_ => {}
+			}
+		}
+
+		let hqs = if let Some(q) = query {
+			execute_query::<models::ByNameResponse>(
+				&graphql::GraphQLQuery::HQS_BY_NAME,
+				Some(serde_json::json!({ "name": q })),
+			)?
+			.get_hqs_by_name
```
kkantan: since you don't use filters when there's a query, you can set the `hidesFiltersWhileSearching` config option in `source.json`. you should also not iterate through the filters if they're not going to be used.

sources/pt.hqnow/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+		})
+	}
+
+	fn get_manga_update(
+		&self,
+		mut manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		if !needs_details && !needs_chapters {
```
kkantan: this won't happen.

sources/pt.hqnow/src/models.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+#[derive(Deserialize)]
+pub struct HqChapter {
+	pub id: i32,
+	pub name: Option<String>,
+	pub number: String,
+}
+
+impl HqChapter {
```
kkantan: you could implement `From<HqChapter> for Chapter` instead to use the `Into::into` trait function.

sources/pt.hqnow/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+
+		Ok(pictures
+			.into_iter()
+			.map(|p| Page {
+				content: PageContent::url(
+					models::to_https(Some(p.picture_url)).unwrap_or_default(),
+				),
+				thumbnail: None,
+				has_description: false,
+				description: None,
```
kkantan:
```suggestion
				..Default::default()
```

kkantan (changes requested):

ramonhpr:
All comments resolved

sources/pt.hqnow/src/helpers.rs
```diff
@@ -0,0 +1,66 @@
...
+	let mut body = serde_json::json!({
+		"operationName": gql.operation_name,
+		"query": gql.query,
+	});
+	if let Some(vars) = variables {
+		body["variables"] = vars;
+	}
+	let body_str = body.to_string();
+	let resp = Request::post(GRAPHQL_URL)
+		.map_err(|_| error!("network error"))?
```
kkantan, line 32: you shouldn't need to map the error since the original one is already a network error. I believe that this only errors if the url is invalid anyways though.
```suggestion
	let resp = Request::post(GRAPHQL_URL)?
```

sources/pt.hqnow/src/helpers.rs
```diff
@@ -0,0 +1,66 @@
...
+		.header("Content-Type", "application/json")
+		.body(body_str.as_bytes())
+		.string()?;
+	let wrapper: GqlResponse<T> = serde_json::from_str(&resp).map_err(|_| error!("parse error"))?;
```
kkantan, line 36: you should be able to use `.json_owned()` or `.send()?.json()` instead, and then there is no reason to map the error.

sources/pt.hqnow/src/models.rs
```diff
@@ -0,0 +1,140 @@
...
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CarouselItem {
+	pub hq_id: i32,
+	pub name: String,
+	pub hq_cover: Option<String>,
+}
+
+impl CarouselItem {
+	pub fn into_manga(self) -> Manga {
```
kkantan, line 130: this can be `impl From<CarouselItem> for Manga` just like the chapter struct. the same applies to `HqBasic`.

sources/pt.hqnow/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+	prelude::*,
+};
+use alloc::collections::BTreeMap;
+use core::cell::RefCell;
+
+// ── Source ────────────────────────────────────────────────────────────────────
+
+struct HQnow {
+	/// Maps manga id → cover URL, kept consistent across all listing methods.
+	cover_cache: RefCell<BTreeMap<i32, String>>,
```
kkantan, line 22: is this necessary to have? I see that the keiyoushi source doesn't have it.

kkantan (changes requested):

PR #646, opened 2026-07-26
feat: add en.stonescape

jckli:
this adds https://stonescape.xyz/ as a source. pretty much based off the keiyoushi source, but thank god for stonescape having a public backend api :D

ps: i did use ai for this so the code quality _may_ not be that good, i did a quick skim and it seems good to me though. let me know if you see any stupid stuff though.

cheers

image


sources/en.stonescape/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+						url.push_str(&included.join(","));
+					}
+				}
+				_ => {}
+			}
+		}
+
+		let bytes = Request::get(&url)?.data()?;
+		let res: SeriesResponse =
+			serde_json::from_slice(&bytes).map_err(|_| AidokuError::Unimplemented)?;
```
kkantan: you can use `.json_owned::<SeriesResponse>()` on the request instead.

sources/en.stonescape/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+			let res: ChapterListResponse =
+				serde_json::from_slice(&bytes).map_err(|_| AidokuError::Unimplemented)?;
+
+			let mut chapters: Vec<Chapter> = res
+				.chapters
+				.into_iter()
+				.map(|c| c.into_chapter(&slug))
+				.collect();
+
+			chapters.reverse();
```
kkantan: using `rev()` on the iterator is better.

sources/en.stonescape/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+				HomeComponent {
+					title: Some("Popular".into()),
+					subtitle: None,
+					value: HomeComponentValue::empty_scroller(),
+				},
+				HomeComponent {
+					title: Some("Latest Updates".into()),
+					subtitle: None,
+					value: HomeComponentValue::empty_manga_chapter_list(),
+				},
```
kkantan, line 167: the website home page calls these "Popular Series" and "Latest Releases", so it might make sense to use the same names.

also, it would be nice if you used the https://stonescape.xyz/api/banner-config endpoint for a "big" scroller or banner component at the top.

sources/en.stonescape/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+							name: "Popular".into(),
+							kind: ListingKind::Default,
+						}),
+					},
+				}));
+			}
+		}
+
+		let latest_url = format!("{API_URL}/series?page=1&limit=20&contentType=manhwa");
+		let latest_res: Result<SeriesResponse> = Request::get(&latest_url)
```
kkantan: you could parallelize these requests with `Request::send_all` to speed up the home loading.

sources/en.stonescape/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+
+impl ImageRequestProvider for StoneScape {
+	fn get_image_request(
+		&self,
+		url: String,
+		_context: Option<aidoku::PageContext>,
+	) -> Result<Request> {
+		Ok(Request::get(url)?
+			.header("Referer", "https://stonescape.xyz/")
+			.header("Origin", "https://stonescape.xyz"))
```
kkantan:
```suggestion
			.header("Origin", BASE_URL))
```

sources/en.stonescape/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+	}
+}
+
+impl DeepLinkHandler for StoneScape {
+	fn handle_deep_link(&self, url: String) -> Result<Option<DeepLinkResult>> {
+		if !url.starts_with(BASE_URL) {
+			return Ok(None);
+		}
```
kkantan: I don't believe this function will ever be called with a url like this.

sources/en.stonescape/src/models.rs
```diff
@@ -0,0 +1,207 @@
...
+	pub author: Option<String>,
+	pub artist: Option<String>,
+	pub genres: Option<Vec<String>>,
+	pub latest_chapter: Option<ChapterData>,
+}
+
+impl Series {
+	pub fn into_manga(self) -> Manga {
+		let cover = self.cover_url.map(|path| format!("{BASE_URL}{path}"));
+		let key = format!("/series/{}", self.slug);
```
kkantan: it seems `/series/` and `/chapters/` are prepended to keys, but then stripped when actually using them. the only thing this seems to simplify is the deep link handling, but I think it would be better just to use the slugs themselves as keys.

kkantan (changes requested):

sources/en.stonescape/src/lib.rs
jckli: hmm i didnt know they had this api route, i was just focused on the popular and latest releases lol yea

sources/en.stonescape/src/lib.rs
```diff
@@ -0,0 +1,300 @@
...
+
+	fn get_search_manga_list(
+		&self,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		let mut url = format!("{API_URL}/series?page={page}&limit=24&contentType=manhwa");
+
+		if let Some(q) = query.filter(|q| !q.is_empty()) {
```
kkantan: the query won't be empty.

sources/en.stonescape/src/lib.rs
```diff
@@ -0,0 +1,300 @@
...
+			url.push_str("&search=");
+			url.push_str(&q);
```
kkantan: this should be url encoded. you can either use `encode_uri_component`, or `QueryParameters`.

sources/en.stonescape/src/models.rs
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	pub fn into_manga_with_chapter(mut self) -> Option<MangaWithChapter> {
+		let slug = String::from(self.slug.as_str());
+		let chapter = self.latest_chapter.take()?.into_chapter(&slug);
+		let manga = self.into_manga();
+		Some(MangaWithChapter { manga, chapter })
+	}
+
+	pub fn apply_details(self, manga: &mut Manga) {
```
kkantan, line 155: the `title` should also be applied here.

sources/en.stonescape/src/models.rs
```diff
@@ -0,0 +1,229 @@
...
+pub struct ChapterData {
+	pub chapter_id: String,
+	pub chapter_number: String,
+	pub title: Option<String>,
+	pub created_at: Option<String>,
+}
+
+impl ChapterData {
+	pub fn into_chapter(self, slug: &str) -> Chapter {
+		let chapter_num = self.chapter_number.parse::<f32>().ok();
```
kkantan: you can use the struct initialization shorthand if you just name this `chapter_number`.
```suggestion
		let chapter_number = self.chapter_number.parse::<f32>().ok();
```

kkantan (changes requested):

sources/en.stonescape/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+
+		let res: SeriesResponse = Request::get(&url)?.json_owned()?;
+
+		let has_next_page = if let Some(pag) = res.pagination {
+			pag.page.unwrap_or(1) < pag.total_pages.unwrap_or(1)
+		} else {
+			false
+		};
+
+		let entries = res.data.into_iter().map(Series::into_manga).collect();
```
kkantan, line 63: this populates the description and tags, which should be avoided for search and listing results. the app sometimes can struggle to process excess data when scrolling through multiple pages.

kkantan (changes requested):
it looks like you're missing the `.cargo/config.toml` file, which is causing the lint workflow to fail. you can copy this from any of the other sources.

kkantan (approved):
thank you for the pr!

PR #647, opened 2026-07-28
feat: add en.ranobes source

BaconDroid:
Adds a new source for ranobes.top (light novel/web novel aggregator).

- Search with full filter support (genre, original language, status, translation status, year range, sort — all include/exclude where applicable)
- Home page (Latest Updates, Popular, Completed)
- Hidden genres/languages settings
- Deep link support for both series and chapter URLs


sources/en.ranobes/res/filters.json
```diff
@@ -0,0 +1,140 @@
...
+			"Sports",
+			"Supernatural",
+			"Tragedy",
+			"Wuxia",
+			"Xianxia",
+			"Xuanhuan",
+			"Yaoi",
+			"Yuri"
+		],
+		"ids": [
```
kkantan: `ids` don't need to be provided if they're the same as `options`.

sources/en.ranobes/res/settings.json
```diff
@@ -0,0 +1,119 @@
+[
+	{
+		"type": "group",
+		"title": "Content Filters",
+		"items": [
+			{
+				"type": "multi-select",
+				"key": "hiddenGenres",
+				"title": "Hidden Genres",
+				"titles": [
```
kkantan: `titles` don't need to be provided if they're the same as `values`.

sources/en.ranobes/src/helpers.rs
```diff
@@ -0,0 +1,253 @@
...
+	manga.tags = html.select("#mc-fs-genre .links a").map(|els| {
+		els.filter_map(|el| el.text())
+			.map(|t| t.trim().to_string())
+			.collect::<Vec<_>>()
+	});
+	manga.content_rating = manga
+		.tags
+		.as_deref()
+		.map(content_rating_from_tags)
+		.unwrap_or(ContentRating::Unknown);
```
kkantan:
```suggestion
		.unwrap_or_default();
```

sources/en.ranobes/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+	"news_read",
+	"comm_num",
+	"d.chap-num",
+	"d.year",
+	"editdate",
+];
+
+/// The site encodes spaces in path-segment values as `+` (confirmed via
+/// `v.genre=Adult,Martial+Arts`).
+fn encode_value(value: &str) -> String {
```
kkantan: all these extra functions at the top here could be in `helpers.rs`.

kkantan (changes requested):
chapters don't load for me, with the error `window.__DATA__ script not found`, and all search results are empty. does the source work for you?

BaconDroid:
Sorry for the ongoing back-and-forth on this branch — still actively
working through it.

To be transparent about where things stand: it tested fine on my end
initially, but I hadn't specifically tried a novel with a large number
of chapter pages (100+), which is where the site's rate limiting/bot
protection actually kicks in — that's likely closer to what you ran
into originally.

Currently iterating on a few approaches to work around it (batching
requests more conservatively, session cookie handling, referer
consistency). If you have any insight into what might help here, or
if you're able to retest once the current round of changes settles,
that'd be very welcome.

kkantan:
cover images no longer load for me. chapters briefly loaded for one series, but loading pages for them failed. after using the source for a minute, no chapters load for any series. pulling to refresh in a series gives me a "Title not found" error.

BaconDroid:
Thanks for testing and for your patience through all of this.

At this point I think the site's bot protection is a bigger obstacle
than we can reasonably work around from within Aidoku's constraints —
we've tried batch size tuning, cookie persistence scoped to chapter
fetching, referer chaining through the actual fetch order, and none of
it has resolved things reliably. The fact that cover images (which
we've never touched) and search are now also failing for you suggests
this may be a broader site/network-level block rather than something
specific to our request logic that we can fix with more header/cookie
adjustments.

I don't think it's worth continuing to iterate on this without a
clearer path forward, so I'm going to step back from this PR rather
than keep asking you to test incremental changes that may not actually
address the root cause. Sorry this didn't land somewhere more solid —
appreciate you taking the time to review and test throughout.

PR #652, opened 2026-07-29
fix(en.comix): handle waf challenge

### LINKED ISSUE #642: Comix breaks at times
Comix at times will break and stop loading, even though the site has no issues. Closing and reopening the app fixes it, but I do not know if this is considered a source issue or something else

jianmingyong:
This is a workaround to comix WAF Challenge page. It is still not heavily tested on whether it would work properly or not so we really need some tester.

For some reason I couldn't get the async JS to work so that is now not in use... idk why the app keep crashing -_-

For anyone who wish to test it out:
[package.zip](https://github.com/user-attachments/files/30516099/package.zip)

Fixes: #642

### DIFF
--- a/sources/en.comix/res/settings.json
+++ b/sources/en.comix/res/settings.json
@@ -214,5 +214,24 @@
 			}
 		],
 		"footer": "Content filters are applied globally: to the home page, listings, and search results."
+	},
+	{
+		"type": "group",
+		"title": "Verify Captcha",
+		"items": [
+			{
+				"type": "login",
+				"method": "web",
+				"url": "https://comix.to/@waf/challenge?return=/",
+				"key": "verify",
+				"title": "Verify Comix Captcha",
+				"logoutTitle": "Re-Verify Comix Captcha",
+				"notification": "verify",
+				"refreshes": [
+					"content"
+				]
+			}
+		],
+		"footer": "Please use this button to open a webview which would allow aidoku to intercept cookies generated after verification. This cookie is only valid for 30 minutes and you have to do this again afterwards. This button is not for Account Login!"
 	}
 ]
--- a/sources/en.comix/res/settings.json
+++ b/sources/en.comix/res/settings.json
@@ -232,6 +232,6 @@
 				]
 			}
 		],
-		"footer": "Please use this button to open a webview which would allow aidoku to intercept cookies generated after verification. This cookie is only valid for 30 minutes and you have to do this again afterwards. This button is not for Account Login!"
+		"footer": "Please use this button to resolve captcha issues that can cause the source to fail to load. The result is only valid for 30 minutes, so you may have to repeat this process. This button is not for Account Login!"
 	}
 ]
--- a/sources/en.comix/res/source.json
+++ b/sources/en.comix/res/source.json
@@ -2,10 +2,11 @@
 	"info": {
 		"id": "en.comix",
 		"name": "Comix",
-		"version": 22,
+		"version": 23,
 		"url": "https://comix.to",
 		"contentRating": 1,
-		"languages": ["en"]
+		"languages": ["en"],
+		"minAppVersion": "0.8.4"
 	},
 	"listings": [
 		{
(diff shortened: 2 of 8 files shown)


kkantan:
@GohanXF could you test if this works for you?

GohanXF:
@kkantan I can verify the Comix WAF but afterwards I'm getting hit with a nonstop Cloudflare verification. Tested on cellular, wifi, and VPN. 

EDIT: I took a screen recording of what's happening on my screen but wanted to make sure I should upload it here or send it somewhere else.

jianmingyong:
@GohanXF 
I think the CF is really just picky for your IP address. I know that aidoku recently had a CF fix for the looping part to get the captcha to show up properly but if it loops forever, that’s probably something else like invalid device datetime or user-agent or simply can’t run their silly js. That one is really hard to tell though if the normal browser works, the webview should work as well.

GohanXF:
The CF part is what's confusing to me now. I didn't have any CF issues before this, it was only the Comix WAF. Another thing to add is that I wasn't getting the WAF challenge when I was on cellular or Wifi + VPN. Comix would load just fine.  The webview is showing up just fine and I can get past that challenge. Is it possible that Comix made the CF challenge more aggressive?

jianmingyong:
@GohanXF they did increase the security level to mess with the mihon users. Of course this had a side effect of having CF checks more often. The app normally handle this just fine but sometimes it can fail.

sources/en.comix/res/settings.json
```diff
@@ -214,5 +214,24 @@
...
+				"key": "verify",
+				"title": "Verify Comix Captcha",
+				"logoutTitle": "Re-Verify Comix Captcha",
+				"notification": "verify",
+				"refreshes": [
+					"content"
+				]
+			}
+		],
+		"footer": "Please use this button to open a webview which would allow aidoku to intercept cookies generated after verification. This cookie is only valid for 30 minutes and you have to do this again afterwards. This button is not for Account Login!"
```
kkantan: I'm thinking that it should be described a little higher level for users, since all they know is that the source stopped working.
```suggestion
		"footer": "Please use this button to resolve captcha issues that can cause the source to fail to load. The result is only valid for 30 minutes, so you may have to repeat this process. This button is not for Account Login!"
```

sources/en.comix/src/lib.rs
```diff
@@ -599,12 +600,34 @@ impl DeepLinkHandler for Comix {
 	}
 }
 
+const VERIFY_COOKIE_KEY: &str = "waf_pass";
+
+impl WebLoginHandler for Comix {
+	fn handle_web_login(&self, key: String, cookies: HashMap<String, String>) -> Result<bool> {
+		if key == "verify" {
```
kkantan: you could use the `VERIFY_KEY` constant you defined.

sources/en.comix/res/source.json
```diff
@@ -2,10 +2,11 @@
 	"info": {
 		"id": "en.comix",
 		"name": "Comix",
-		"version": 22,
+		"version": 23,
 		"url": "https://comix.to",
 		"contentRating": 1,
-		"languages": ["en"]
+		"languages": ["en"],
+		"minAppVersion": "0.8.4"
```
kkantan, line 9: do these changes still require 0.8.4?

kkantan (changes requested):

kkantan:
I think cloudflare may have updated their verification process, since multiple sources that use it have stopped working (#654). I was working on a source that was working before, but now no longer can bypass cloudflare despite completing the captcha. requests return cloudflare page html.

I'm not sure if this is something we can resolve from the source side, or if it requires an app update.

jianmingyong:
Nothing much you could do with cloudflare. If you can solve the cloudflare captcha in a webview, then it should work just fine with 0.8.4 update. You can always manually grab the cloudflare cookies with the fake login button but that should be handled by aidoku app.

0.8.3 should not be used for any cloudflare sources as that one doesn’t set the cookie correctly.

I forgot to add a better error check in the webview initialization. It would also print module not found for cf related error.

kkantan (approved):

PR #657, opened 2026-08-01
fix(ru.desu): migrate to new /api/manga endpoints

remmody:
The legacy /manga/api endpoint now returns api_deprecated. Switch the source to /api/manga for listing, details, chapters, and pages, and update genre filter ids for the new API.


remmody:
The project's authorization needs further work; I'll come back to it later

PR #659, opened 2026-08-01
fix(ru.desu): migrate to new API and web login

remmody:
## Summary
- Migrate ru.desu to the new `/api/manga` API and add ranobe section support
- Switch auth to WebView XenForo cookies and show account status in settings
- Mark version 5 as a breaking change and migrate old numeric manga ids to `m:{id}`

## Test plan
- [x] Catalog / search / manga / chapters / pages without login
- [x] WebView login, cookies persist, username in settings
- [x] Existing library entries with numeric ids migrate to `m:` keys
- [x] Ranobe listing and chapter text reading


sources/ru.desu/src/helpers.rs
```diff
@@ -77,38 +125,61 @@ pub fn search(
...
 					.into_iter()
-					.chain(excluded.into_iter().map(|x| format!("!{}", x.as_str())))
+					.chain(excluded.into_iter().map(|x| format!("!{x}")))
 					.collect();
 				if id.eq("genres") || id.eq("tags") {
-					genres.extend(v)
-				} else {
-					params.push(&id, Some(&v.join(",")))
+					genres.extend(values);
+				} else if !values.is_empty() {
```
kkantan: I don't think it's possible for a multi-select filter to be sent with no included or excluded values, so this check is probably useless.

sources/ru.desu/src/lib.rs
```diff
@@ -45,44 +77,104 @@ impl Source for Desu {
...
+					content: PageContent::url(url),
 					..Page::default()
 				})
-			})
-			.collect();
-
-		Ok(pages)
+				.collect()),
+			Section::Ranobe => {
+				let url = chapter.url.clone().unwrap_or_else(|| {
```
kkantan: does this need to be cloned?

sources/ru.desu/src/lib.rs
```diff
@@ -45,44 +77,104 @@ impl Source for Desu {
...
+			.and_then(|s| s.rsplit_once('.').map(|(_, id)| id))
+			.or_else(|| {
+				path.split('/')
+					.find(|s| !s.is_empty() && s.chars().all(|c| c.is_ascii_digit()))
+			})
 			.ok_or(error!("Invalid URL"))?;
 
 		Ok(Some(DeepLinkResult::Manga {
-			key: manga_id.to_string(),
+			key: format!("m:{manga_id}"),
```
kkantan, line 182: since you're changing ids, you should mark this source version as a breaking change and implement the id migration trait if possible.

kkantan (changes requested):
the desu source changes look good, but you should split the added remanga source into a separate pr.

remmody:
> the desu source changes look good, but you should split the added remanga source into a separate pr.

ok, as for the other changes, I based them on the fact that the source material also includes ranobe.

sources/ru.desu/src/lib.rs
remmody: It is backward compatible, so it works correctly.

remmody:
Split Remanga into a separate PR: https://github.com/Aidoku-Community/sources/pull/664

Also addressed the review notes on this PR:
- removed the empty MultiSelect values guard
- avoided cloning `chapter.url` in `get_page_list`
- marked version 5 as a breaking change and added `MigrationHandler` for numeric -> `m:{id}` manga keys

kkantan (approved):

PR #660, opened 2026-08-02
feat: add ja.rawdevart

### LINKED ISSUE #34: rawdevart.art
rewrite request

https://github.com/Skittyblock/aidoku-community-sources/blob/main/src/rust/ja.rawdevart

kinboy56:
fix https://github.com/Aidoku-Community/sources/issues/34

## Summary

- Search, browse/filtering, series details, chapter list, and page list
- 3 listings: Latest, Popular, Trending Today
- Filters: sort (3 options), status, and genre (fetched dynamically, 60+ options)
- Deep linking for both series and chapter URLs

## Implementation notes

A few site quirks that aren't obvious from the API responses and are worth a reviewer's
attention:

- **The `/spa` endpoints require no special headers.** The site's own scripts send a `Referer`
  and `X-Requested-With`, but a direct request without either returns a byte-identical response
  (verified by diffing both). Plain `Request::get` is used throughout; only the image requests
  set a `Referer`, since the CDN mirrors do require one.
- **Chapter pages carry no image urls in the api's HTML fragment as-is** — the markup uses
  `` with paths relative to a `server` field the same response
  provides (plus `slaves` as mirrors). `abs:` attribute resolution isn't used to join them,
  since it depends on the parsed document keeping a base url, which a `Html::parse_fragment`
  fragment doesn't reliably carry; the join is done by hand instead.
- **Chapters are keyed by `chapter_number`, not `chapter_id`.** The page endpoint is
  `/spa/manga/{manga_id}/{chapter_number}`; passing a `chapter_id` there returns a null
  `chapter_detail` (confirmed against the live api). Chapter numbers are non-integer for roughly
  85% of entries in the sampled catalogue (e.g. `34.2`) and were confirmed to be unique and
  non-null across every chapter checked, so they're safe to use as the key.
- **Genres are fetched instead of hardcoded.** `DynamicFilters` reads the `<option>` markup the
  `/spa/genre/all` response embeds under `genreOpt`, so new genres the site adds don't need a
  source update. The "all" default plus every id past it are validated as numeric in a test.
- **`manga_status` can be entirely absent from the API response**, not just `true`/`false`.
  Missing status maps to `MangaStatus::Unknown` rather than defaulting to "ongoing".
- **Chapter `language` is left unset on purpose.** Setting it (as the previous, now-removed
  version of this source did) subjects chapters to the app's chapter-language filter, which can
  hide every chapter depending on the user's device settings for a single-language source. Most
  existing `ja.*` sources in this repo leave it unset for the same reason.
- **Series and chapter urls prefix the numeric id with two letters** (e.g. `/g/ne854721`) that
  the site's routing ignores; the api only ever wants the digits. `DeepLinkHandler` strips the
  prefix and rejects anything that isn't left with a purely numeric id, and also trims query
  strings / fragments first, since shared links commonly carry tracking parameters.
- **Content rating isn't provided by the api.** It's inferred from genre tags
  (`adult`/`hentai`/`loli`-family tags → NSFW, `ecchi` → Suggestive, otherwise Safe), since the
  source itself is registered as `contains-nsfw`.

## Test plan

- [x] `cargo test --release` — 9 tests, all hitting the live api (no mocked responses), covering:
  all 3 listings, search, filter combinations, dynamic genre filter parsing, series details
  (title/cover/authors/tags/description/status), a chapter with a non-integer number round-tripped
  into a page request, page list urls, deep link parsing (including tracking-parameter and
  non-numeric-id variants), and a regression test against the specific series/chapter pair
  originally reported as unreadable
- [x] `cargo clippy --release` — no warnings
- [x] `aidoku package` + `aidoku verify` — schema validation passes for `source.json` and
  `filters.json`, icon is 128×128 and fully opaque
- [x ] Manual on-device install via `aidoku serve` — not yet done in this environment

## Checklist

- [x] `status` / `content_rating` are set on series when available (`status` is `Unknown` when
  the api doesn't provide it, rather than guessed)
- [x] `DeepLinkHandler` is implemented for both series and chapter urls
- [x] Genre filtering uses the site's own options via `DynamicFilters`, kept in sync automatically
- [ ] Home page / listing components beyond simple listings — not implemented (optional per
  CONTRIBUTING.md)


sources/ja.rawdevart/res/source.json
```diff
@@ -0,0 +1,25 @@
+{
+	"info": {
+		"id": "ja.rawdevart",
+		"name": "Rawdevart",
+		"version": 1,
```
kkantan: this is a rewrite of a legacy source using the same id, so it should start at version 3.
```suggestion
		"version": 3,
```

sources/ja.rawdevart/src/lib.rs
```diff
@@ -0,0 +1,615 @@
...
+		Self
+	}
+
+	fn get_search_manga_list(
+		&self,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		// the search endpoint takes none of the other filters, so a query overrides them
```
kkantan, line 49: you should set the config option in `source.json` that hides filters while searching.

sources/ja.rawdevart/src/lib.rs
```diff
@@ -0,0 +1,615 @@
...
+		if let Some(query) = query.filter(|query| !query.is_empty()) {
```
kkantan: the query should never be empty.
```suggestion
		if let Some(query) = query {
```

sources/ja.rawdevart/src/lib.rs
```diff
@@ -0,0 +1,615 @@
...
+		match tag.to_lowercase().as_str() {
+			"adult" | "hentai" | "loli" | "lolicon" | "mature" | "shotacon" | "smut" => {
+				return ContentRating::NSFW;
+			}
+			"ecchi" => rating = ContentRating::Suggestive,
+			_ => continue,
+		}
+	}
+	rating
+}
```
kkantan: it would be nice to put these three extra functions in a `helpers.rs` file.

kkantan (changes requested):

sources/ja.rawdevart/res/source.json
kinboy56: Fixed

sources/ja.rawdevart/src/lib.rs
kinboy56: Fixed

sources/ja.rawdevart/src/lib.rs
kinboy56: Fixed — added `config.hidesFiltersWhileSearching: true` to source.json, since the search endpoint doesn't accept any other filters alongside a query.

sources/ja.rawdevart/src/lib.rs
kinboy56: Fixed — moved `manga_key`, `strip_html`, and `content_rating` into a new `src/helpers.rs`.

kkantan (approved):

PR #661, opened 2026-08-02
feat: add ja.mangarawbest

### LINKED ISSUE #16: [request] mangaraw.best
Lang: JA
https://mangaraw.best

kinboy56:
Fix https://github.com/Aidoku-Community/sources/issues/16

## Summary

- Search, browse/filtering, series details, chapter list, and page list
- 3 listings: Popular, Weekly Popular, Newest
- Filters: sort (8 options), search-by name/artist, status, and genre (52 options, include/exclude)
- Selectable image server (site ships 3 CDN mirrors; some regions can't reach the default one)
- Deep linking for both series and chapter URLs

## Implementation notes

- **`/search?q=...` silently ignores the query** — it always returns the unfiltered catalogue. The actual search endpoint is `filter[name]=...` / `filter[artist]=...` on `/manga-list`, reverse-engineered from the site's own `app.js`. `filter[doujinshi]=...` (the third "search type" option in the site's own UI) never returns any results regardless of value, so it was left out of `filters.json` rather than shipping a filter option that silently returns nothing.
- **Genres appear twice on series pages**: once in the actual info block, and again in an SEO keyword cloud at the bottom of the page that links to the same `/genre/...` URLs but prefixes every label with the series title (e.g. `<title> Action`). Tags are scraped from a selector scoped to the info block only, and there's a test asserting no tag contains the series title, so a regression here fails loudly instead of silently reintroducing junk tags.
- **Chapter titles carry no information beyond the number** — every chapter across the sampled catalogue is titled exactly `第N話`, which the app already renders from `chapter_number`. Keeping the scraped title as well would show it twice in the reader (e.g. "41話 - 41話"), so titles that are just `第<number>話` are dropped.
- **Dates are relative-only** (`6日前`, `2週間前`, etc.) — there's no absolute date anywhere in the markup, so they're converted from `current_date()`.
- **Status label text is inconsistent**: the "completed" badge reads "完了" on some pages, and the filter link it sits inside consistently uses `filter[status]=1`. Status is parsed from that query value first, falling back to the label text only if the link is missing.
- **The source has no explicit content rating of its own** — `content_rating` is inferred from genre tags (`成人向け`/`hentai`-family tags → NSFW, `Ecchi`/`Mature`/etc. → Suggestive, otherwise Safe), since the source itself is registered as `contains-nsfw`.

## Test plan

- [x] `cargo test` — 19 tests, all hitting the live site (no mocked HTML), covering: browse pagination (including the single-page/no-results case where the pagination bar is absent), all 3 listings, search hits, filter combinations actually narrowing results (asserted against the unfiltered baseline, not just "doesn't error"), series details, chapter list ordering/format, page list, and deep link parsing (including `www.` / query-string / trailing-slash variants)
- [x] `cargo clippy` — no warnings
- [x] `aidoku package` + `aidoku verify` — schema validation passes for `source.json`, `filters.json`, `settings.json`, and the icon (128×128, fully opaque)
- [x] Manually installed via `aidoku serve` and read a chapter in the iOS app

## Checklist

- [x] `status` / `content_rating` are set on series when available
- [x] `DeepLinkHandler` is implemented for both series and chapter URLs
- [ ] Home page / listing components beyond simple sort-based listings — not implemented (optional per CONTRIBUTING.md)


sources/ja.mangarawbest/res/filters.json
```diff
@@ -0,0 +1,152 @@
+[
+	{
+		"type": "sort",
+		"id": "sort",
+		"title": "並び替え",
+		"canAscend": false,
+		"default": {
+			"index": 0,
+			"ascending": false
+		},
```
kkantan: this is already the default so you don't need to provide it explicitly.

sources/ja.mangarawbest/src/lib.rs
```diff
@@ -0,0 +1,663 @@
...
+				},
+				_ => {}
+			}
+		}
+
+		let mut qs = QueryParameters::new();
+		qs.push("sort", Some(sort));
+		qs.push("page", Some(&page.to_string()));
+
+		if let Some(query) = query.as_deref().map(str::trim).filter(|q| !q.is_empty()) {
```
kkantan:
```suggestion
		if let Some(query) = query.as_deref().map(str::trim) {
```

sources/ja.mangarawbest/src/lib.rs
```diff
@@ -0,0 +1,663 @@
...
+		&self,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		let mut sort = SORT_VALUES[0];
+		let mut search_type = String::from("name");
+		let mut statuses: Vec<String> = Vec::new();
+		let mut included_genres: Vec<String> = Vec::new();
+		let mut excluded_genres: Vec<String> = Vec::new();
```
kkantan: it seems like these three items could just be applied to the query parameters inside the filters loop.

sources/ja.mangarawbest/src/lib.rs
```diff
@@ -0,0 +1,663 @@
...
+			},
+			// ex: https://mangaraw.best/raw/tu-long-nobai
+			None => DeepLinkResult::Manga {
+				key: manga_key.into(),
+			},
+		}))
+	}
+}
+
+/// Parses a manga grid page, shared by search, filtering and listings.
```
kkantan: these extra functions should probably be in `helpers.rs`.

sources/ja.mangarawbest/src/lib.rs
```diff
@@ -0,0 +1,663 @@
...
+						.or_else(|| link.attr("title"))?;
+
+					// Covers are lazy-loaded: src holds a placeholder until the
+					// real url in data-src is swapped in.
+					let cover = element
+						.select_first("img.cover")
+						.and_then(|el| el.attr("abs:data-src").or_else(|| el.attr("abs:src")));
+
+					Some(Manga {
+						key: key.clone(),
```
kkantan: if you define the url before this, the key doesn't need to be cloned.

kkantan (changes requested):

sources/ja.mangarawbest/res/filters.json
kinboy56: Fixed

sources/ja.mangarawbest/src/lib.rs
kinboy56: Fixed

sources/ja.mangarawbest/src/lib.rs
kinboy56: Done — `statuses`, `included_genres` and `excluded_genres` are gone, and the
`status` / `genre` arms now push straight onto the `QueryParameters`.

`sort` and `search_type` are still locals: `search_type` decides the key the
search query is pushed under (`filter[name]` vs `filter[artist]`), so neither can
be written until the whole filter list has been walked.

sources/ja.mangarawbest/src/lib.rs
kinboy56: Moved `parse_manga_page`, `parse_tags`, `parse_status` and `parse_chapters` into
`helpers.rs`. `lib.rs` is now just the trait impls (663 -> 501 lines), and the
now-unused `html::Document` import was dropped from it.

sources/ja.mangarawbest/src/lib.rs
kinboy56: Fixed

kkantan (approved):

PR #664, opened 2026-08-02
feat: add ru.remanga

remmody:
## Summary
- Add `ru.remanga` for remanga.org: catalog, search, chapters, and pages
- WebView Bearer cookie login with manual token fallback
- Paid / free chapter handling and account status in settings

## Test plan
- [x] Default catalog sort by updates
- [x] Long chapter lists load progressively
- [x] Free chapters without login / paid chapters after login
- [x] WebView login persists and shows username / balance


sources/ru.remanga/res/source.json
```diff
@@ -0,0 +1,12 @@
+{
+	"info": {
+		"id": "ru.remanga",
+		"name": "ReManga",
+		"version": 1,
```
kkantan: there is [a legacy source with the same id](https://github.com/Skittyblock/aidoku-community-sources/blob/3b0b855465659e21466b56c3a771c52471833068/src/rust/ru.remanga/res/source.json), so this should start at version 3 and probably mark it as a breaking change if the ids changed. it's not necessary to implement the migration handler though.
```suggestion
		"version": 3,
```

sources/ru.remanga/src/helpers.rs
```diff
@@ -0,0 +1,311 @@
...
+	let branches = match branches {
+		Some(b) if !b.is_empty() => b,
+		_ => fetch_branches(slug)?,
+	};
+	let branch = select_primary_branch(branches)?;
+	let label = branch
+		.publishers
+		.as_ref()
+		.and_then(|p| p.first())
+		.and_then(|p| p.name.clone());
```
kkantan: I feel like this shouldn't need to be cloned if you don't use the `as_ref` on publishers, since it's okay to consume as the publishers aren't referenced later.

sources/ru.remanga/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+
+	fn get_manga_update(
+		&self,
+		manga: Manga,
+		needs_details: bool,
+		needs_chapters: bool,
+	) -> Result<Manga> {
+		let key = manga.key.clone();
+		let (mut item, branches) = if needs_details {
+			fetch_manga_with_branches(&key, manga)
```
kkantan: does this function even need to take the slug as a separate parameter if you can just use the key on the manga object you already pass in?

sources/ru.remanga/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+	fn handle_deep_link(&self, url: String) -> Result<Option<DeepLinkResult>> {
+		let rest = url
+			.strip_prefix("https://")
+			.or_else(|| url.strip_prefix("http://"))
+			.unwrap_or(&url);
+		let path = rest
+			.strip_prefix("remanga.org/")
+			.or_else(|| rest.strip_prefix("www.remanga.org/"))
+			.unwrap_or(rest)
+			.trim_start_matches('/');
```
kkantan: this function only sends urls prefixed by the url defined in `source.json`, so you could just strip `https://remanga.org`.

sources/ru.remanga/src/models.rs
```diff
@@ -0,0 +1,523 @@
...
+		let key = self.dir.filter(|s| !s.is_empty())?;
+		let title = pick_title(self.main_name, self.secondary_name, &key);
+		let tname = type_name_card(self.title_type.as_ref());
+		let cover = self
+			.cover
+			.as_ref()
+			.and_then(Cover::best)
+			.or_else(|| self.img.as_ref().and_then(Cover::best));
+		Some(Manga {
+			key: key.clone(),
```
kkantan: you can define the url before and this doesn't need to be cloned.

sources/ru.remanga/src/models.rs
```diff
@@ -0,0 +1,523 @@
...
+
+impl TitleDetail {
+	pub fn into_manga(self, existing: Option<Manga>) -> Manga {
+		let key = self
+			.dir
+			.clone()
+			.or_else(|| existing.as_ref().map(|m| m.key.clone()))
+			.unwrap_or_default();
+		let mut manga = existing.unwrap_or(Manga {
+			key: key.clone(),
```
kkantan: no need to set this here if you're setting it again immediately after. you could just do `unwrap_or_default`.

sources/ru.remanga/res/icon.png
kkantan, line 1: the icon should have an opaque background, you could just make it white.

kkantan (changes requested):

remmody:
Addressed the review notes:
- bumped to version **3** with `breakingChangeVersion: 3` (legacy source was v2)
- dropped the extra slug param / needless clones
- simplified deep-link parsing to strip the `source.json` base URL
- replaced the icon with an opaque white-background Remanga yin-yang logo

sources/ru.remanga/src/helpers.rs
```diff
@@ -0,0 +1,310 @@
...
+				"status" => {
+					for value in included {
+						params.push("status", Some(&value));
+					}
+				}
+				"age_limit" => {
+					for value in included {
+						params.push("age_limit", Some(&value));
+					}
+				}
```
kkantan: these could be generalized to the `_` branch:
```rs
_ => {
	for value in included {
		params.push(&id Some(&value));
	}
}

File: sources/ru.remanga/src/lib.rs [hunk 8]
```diff
@@ -0,0 +1,172 @@
...
+		} else {
+			(manga, None)
+		};
+
+		if needs_chapters {
+			if needs_details {
+				send_partial_result(&item);
+			}
+			let key = item.key.clone();
+			let chapters = fetch_chapters(&key, branches, |partial| {
```
kkantan:
```suggestion
			let chapters = fetch_chapters(&item.key, branches, |partial| {
```

kkantan (changes requested):

remmody:
Addressed the latest review notes:
- generalized MultiSelect catalog filters into a shared `_` branch
- refactored chapter fetching to update the manga in place (no extra key clone at the call site)

sources/ru.remanga/src/models.rs
```diff
@@ -0,0 +1,564 @@
...
+				|| id == Some(41)
+				|| lower.contains("яой")
+				|| lower.contains("yaoi")
+				|| lower.contains("юри")
+				|| lower.contains("yuri")
+				|| lower.contains("эро")
+				|| lower.contains("hentai")
+				|| lower.contains("хент")
+			{
+				nsfw = true;
```
kkantan, line 370: it's more efficient to break after this since it won't change after processing more genres. there's also no reason to iterate the genres in the first place if `nsfw` is already true.
```suggestion
				nsfw = true;
				break;
```

kkantan (approved):

remmody:
Addressed the latest note: skip genre scanning when NSFW is already known, and break after the first adult genre match.

PR #674, opened 2026-08-04
feat: add en.readcomicsonline

### LINKED ISSUE #558: readcomicsonline.ru
Readcomicsonline.ru

kkantan:
closes #558.

this source has some aggressive cloudflare requirements, so I am not sure how nice it will be to use, but people have been requesting western comic sources and this appears to be the only one suggested.


idonthaveanametoday:
> closes #558.
> 
> this source has some aggressive cloudflare requirements, so I am not sure how nice it will be to use, but people have been requesting western comic sources and this appears to be the only one suggested.

Sorry if i don't understand the technicals of github

I'm still now showing any resultd after doing the cloudfare verify and same with batcave

kkantan:
with batcave, you need to go to the settings and verify before the source will work. it may be region locked to some areas, so you may have to use a vpn.

readcomicsonline requires app version v0.8.4 for cloudflare handling to work correctly. if it fails, you can clear the network cache in advanced settings to try again.

idonthaveanametoday:
Ok thanks. I updated both sources . Batcave now works ( it's logged into my account?)

readcomics gets stuck on loading . I've cleared source cache , network cache and used vpns. The site loads fine on its own though

PR #677, opened 2026-08-05
feat(ja.mangarawjp): add sort filter for updated and ranking orderings 

kinboy56:
## Summary

Adds a sort filter to `ja.mangarawjp`. The browse tab previously had no way to reach anything by order — it was hardwired to the site's empty-search endpoint, which returns the catalogue in the site's own default order, so finding something popular or recently updated meant already knowing its name.

- New `res/filters.json` with a sort filter: 最近の更新 and ランキング, defaulting to the former
- `get_search_manga_list` maps the selected option to its own path
- `config.hidesFiltersWhileSearching` set, since the search endpoint cannot take an ordering
- Listing scraping extracted into `parse_listing_page`, shared by search and browse
- Source `version` bumped 1 → 2

## Implementation notes

The site has no sort parameter — each order is served from a separate path:

| Sort option | Endpoint |
|---|---|
| 最近の更新 (default) | `/page/N/` |
| ランキング | `/ranking/N/` |
| search | `?s=<query>&page=N` |

A few things worth a reviewer's attention:

- **A query wins over the sort value.** Searching is a separate endpoint that no ordering can be applied to, so `hidesFiltersWhileSearching` hides the filter, and a non-empty query takes precedence over whatever sort value is still stored. An empty query is not treated as a search and falls through to the ordering paths. `test_query_takes_precedence_over_sort` guards the first case and `test_empty_query_falls_through_to_the_sort` the second.
- **Page 1 of the updates ordering is the site home page**, and that page stacks four sibling `div.post-list` blocks with no distinguishing class: updates, ranking, and one per featured genre (Ecchi / オトナ). Selecting `.post-list > a` there returns 84 entries from all four blocks with 7 duplicates, instead of the 24 that actually belong to page 1. Entries are therefore scoped to the *first* `.post-list` only. Every other page — `/page/2/` onwards, `/ranking/N/`, and the search endpoint — has exactly one such block, so the narrower selector is correct everywhere, not just on the home page. `test_updated_page_1_holds_only_the_updates_block` asserts both no-duplicate-keys and a page-size bound so this can't regress silently. The home page is unavoidable here: the "もっと" link on the updates block points at `/page/2/`, so there is no dedicated URL serving page 1 of that ordering.
- **A missing `.post-list` is an error, not an empty page.** An exhausted listing still renders the block with nothing inside it (`/page/9999/` returns 200 with an empty block), so the two cases are distinguishable: no block at all means the page didn't load. Returning an empty list there would be indistinguishable from the end of pagination, and in the app it leaves the previous list on screen — which reads as the ordering having been ignored.
- **Hrefs are read as-is rather than with `abs:`.** `Manga.key` is the site-relative path (`/manga-raw/...`), which is exactly what the href already holds, so reading it directly avoids resolving to an absolute URL only to strip the base back off. Covers do need absolute URLs and are joined onto `BASE_URL`, reading `data-src` only — the plain `src` is a base64 placeholder until the page's lazy loader runs, so falling back to it would yield a data URI.
- **`/page/1/` is used rather than `/`** even though they serve identical content, to keep one URL shape per ordering instead of special-casing the first page. Same for `/ranking/1/`, which the site redirects to from `/ranking/1`.

## Test plan

- [x] `cargo test --release` — 12 tests against the live site, all passing: both orderings return entries, they return *different* leading keys (so the filter is provably not a no-op), a query beats a stored sort value while an empty query does not, updates page 2 paginates, keys stay site-relative and covers absolute, out-of-range pages end pagination cleanly, the sort index falls back to the first option when the app sends no filter value, and the home-page block-mixing regression guard
- [x] Regression guard verified by reverting the selector to `.post-list > a` and confirming `test_updated_page_1_holds_only_the_updates_block` fails (84 entries, 77 unique), then restoring
- [x] `test_query_takes_precedence_over_sort` verified by disabling the query branch and confirming it fails, then restoring — asserting on titles alone would not have caught the fallback, since the query is also the top ranking entry
- [x] `cargo clippy --release` — no warnings
- [x] `aidoku package` + `aidoku verify` — `source.json`, `filters.json` and the icon (128×128, fully opaque) pass schema validation
- [x] Verified on device (iOS): the sort filter shows both options in the filter sheet with 最近の更新 selected by default, switching to ランキング swaps the list, and the filter is hidden while a search is active

## Checklist

- [x] `Manga.key` / `Chapter.key` unchanged — no migration needed, so no `breakingChangeVersion`
- [x] Source `version` incremented
- [ ] Home components — not implemented (optional per CONTRIBUTING.md)

### DIFF
--- a/sources/ja.mangarawjp/res/filters.json
+++ b/sources/ja.mangarawjp/res/filters.json
@@ -1,0 +1,8 @@
+[
+	{
+		"type": "sort",
+		"id": "sort",
+		"canAscend": false,
+		"options": ["最近の更新", "ランキング"]
+	}
+]
--- a/sources/ja.mangarawjp/src/lib.rs
+++ b/sources/ja.mangarawjp/src/lib.rs
@@ -34,44 +34,24 @@ impl Source for MangaRawJP {
 		&self,
 		query: Option<String>,
 		page: i32,
-		_filters: Vec<FilterValue>,
+		filters: Vec<FilterValue>,
 	) -> Result<MangaPageResult> {
-		let url = format!(
-			"{BASE_URL}?s={}&page={page}",
-			encode_uri_component(query.unwrap_or_default())
-		);
+		// searching has its own endpoint that the ordering paths cannot be applied
+		// to, so a query takes precedence over the sort filter, which the app hides
+		// while searching
+		if let Some(query) = query.filter(|query| !query.is_empty()) {
+			let query = encode_uri_component(query);
+			return parse_listing_page(&format!("{BASE_URL}?s={query}&page={page}"));
+		}
 
-		let html = Request::get(&url)?.html()?;
-
-		let entries = html
-			.select(".post-list > a")
-			.map(|elements| {
-				elements
-					.filter_map(|element| {
-						let url = element.attr("abs:href")?;
-						let key = url.strip_prefix(BASE_URL).map(String::from)?;
-						let title = element.select_first("h3")?.text()?;
-						let cover = element.select_first("img").and_then(|img| {
-							img.attr("abs:data-src").or_else(|| img.attr("abs:src"))
-						});
-						Some(Manga {
-							key,
-							title: clean_title(title),
-							cover,
-							url: Some(url),
-							..Default::default()
-						})
-					})
-					.collect::<Vec<Manga>>()
-			})
-			.unwrap_or_default();
-
-		let has_next_page = !entries.is_empty();
-
-		Ok(MangaPageResult {
-			entries,
-			has_next_page,
-		})
+		// the site has no sort parameter: each ordering is served from its own path
+		let url = match sort_index(&filters) {
+			// all-time view count, descending
+			1 => format!("{BASE_URL}/ranking/{page}/"),
+			// most recently updated chapters first; page 1 is the site home page
+			_ => format!("{BASE_URL}/page/{page}/"),
+		};
+		parse_listing_page(&url)
 	}
 
 	fn get_manga_update(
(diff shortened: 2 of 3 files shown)


sources/ja.mangarawjp/src/lib.rs
```diff
@@ -223,6 +191,71 @@ impl Source for MangaRawJP {
...
+		.ok_or_else(|| error!("Manga list not found"))?;
+
+	let entries = list
+		.select("a")
+		.map(|elements| {
+			elements
+				.filter_map(|element| {
+					// hrefs are site-relative and the parser is not guaranteed to keep
+					// the base url on device, so they are joined manually rather than
+					// resolved with "abs:"
```
kkantan: > the parser is not guaranteed to keep the base url on device

is this true? in what case?

kkantan (changes requested):
I think, rather than listings, it would be better to have a sort filter that is hidden while searching. it's nicer to have a single page for the source instead of separating the search and listings, and since filters are saved locally, people can set a default sort.

sources/ja.mangarawjp/src/lib.rs
kinboy56: No — that was wrong, and I've replaced the comment.

sources/ja.mangarawjp/res/filters.json
```diff
@@ -0,0 +1,11 @@
+[
+	{
+		"type": "sort",
+		"id": "sort",
+		"canAscend": false,
+		"options": ["最近の更新", "ランキング"],
+		"default": {
+			"index": 0
+		}
```
kkantan: the default does not need to be provided.
```suggestion
		"options": ["最近の更新", "ランキング"]
```

sources/ja.mangarawjp/src/lib.rs
```diff
@@ -223,6 +203,69 @@ impl Source for MangaRawJP {
...
+					// already holds, so it is read directly rather than resolved
+					// with "abs:" and stripped back down
+					let key = element.attr("href")?;
+					let url = format!("{BASE_URL}{key}");
+					let title = element.select_first("h3")?.text()?;
+					// the plain src is a base64 placeholder until the lazy loader runs
+					let cover = element
+						.select_first("img")
+						.and_then(|img| img.attr("data-src"))
+						.map(|src| format!("{BASE_URL}{src}"));
```
kkantan: the parsing for html elements did not need to change.
```suggestion
						.and_then(|img| img.attr("abs:data-src"));
```

kkantan (changes requested):

sources/ja.mangarawjp/src/lib.rs
kinboy56: Reverted to abs:href + strip_prefix and abs:data-src, and dropped the comment that justified the change — it rested on the base-url claim I retracted above. Verified against a live listing page: the keys stay relative and the covers come out absolute, same as before. I left out the .or_else(|| img.attr("abs:src")) fallback, matching your suggestion — src is always a data:image/gif;base64,... placeholder until the lazy loader runs, so the fallback could only ever yield that placeholder as a cover.

sources/ja.mangarawjp/res/filters.json
kinboy56: Removed.

kkantan (approved):

PR #683, opened 2026-08-07
feat: add ja.senmanga

kinboy56:
Rewrites ja.senmanga for the new api. The source existed in the legacy repository, so this is version 2 rather than a new source.

The site is now a React SPA — the server only returns `<div id="root"></div>`, so there is nothing to parse out of the HTML, and none of the v1 selectors apply. The implementation targets the JSON API the site's own bundle calls, at `https://raw.senmanga.com/api`.

### Endpoints

- `/directory?page=&query=&order=&status=&type=` — backs both browsing and search
- `/manga/{slug}` — details and chapter list
- `/read/{slug}/{chapter}` — page list

### Migration from v1

The keys had to change, so `config.breakingChangeVersion` is 2 and there is a `MigrationHandler`.

v1 derived its keys from site paths (`strip_prefix(BASE_URL)`), giving `/<slug>` for a series and `/<slug>/<number>` for a chapter. Manga keys therefore only need the slashes trimmed. Chapters are less direct: the api addresses a chapter by a `<number>.<id>` key that appears nowhere but the chapter list, so the handler fetches `/api/manga/{slug}` and matches on the chapter number. A chapter that is no longer listed fails rather than resolving to something wrong. Both paths are covered by `migration_test`.

### Decisions worth explaining

- **Search goes through `/directory?query=`, not `/search`.** The site's own search page calls `/api/search?q=`, but that endpoint ignores `q` and returns the whole directory. `/directory?query=` filters correctly, and since it matches against `alt_name` as well, japanese queries reach the romanized entries (`ワンピース` finds One Piece).
- **`has_next_page` compares `currentPage` against `totalPages`.** `totalPages` is the pager's display window rather than the real total — it reports 5 on page 1 and 202 on page 200 — so it cannot be used as an absolute bound. Both fields come back null when the results fit on a single page, which is treated as having no next page.
- **The details endpoint always answers with a null `status`**, while the directory entry the manga came from carries one, so the existing value is kept rather than overwritten with `Unknown`. There is a test covering this.
- **No genre filter.** The bundle contains a genre name to id map, but sending `genre=` does not change the response and `/api/genres` is 404, so the filter is not exposed rather than shipped broken.
- **No `ImageRequestProvider`.** Pages are served from `cdn.kumacdn.club` and hotlink fine without a `Referer`.
- Chapter dates come from the api's ISO 8601 `datetime` field, parsed with `yyyy-MM-dd'T'HH:mm:ssXXX`. A literal `'Z'` parses too, but drops the offset and reads the timestamp in the device timezone.
- `Chapter.language` is left unset, since the source is japanese-only and setting it would put every chapter behind the app's language filter.
- **`minAppVersion` is `0.7.1`** — `std::parse_date` raises the api version the wasm requires, and the default of `0.7.0` fails verification.

### Verification

- `cargo fmt` and `cargo clippy --release` are clean
- `cargo test --release` — 7 tests: browse, search, details, status, page list, deep links, migration
- `aidoku package` and `aidoku verify` pass

One caveat for anyone running the tests: the api returns a sporadic 500 on otherwise valid requests, unrelated to load — parallel requests all succeed while sequential ones occasionally do not — so a run fails on a different test now and then. Re-running clears it.


sources/ja.senmanga/res/source.json
```diff
@@ -0,0 +1,11 @@
+{
+	"info": {
+		"id": "ja.senmanga",
+		"name": "Sen Manga",
+		"version": 1,
```
kkantan: this is once again a legacy source rewrite, so the version should be incremented. https://github.com/Skittyblock/aidoku-community-sources/blob/main/src/rust/ja.senmanga/res/source.json

you're using different keys, so you should optionally implement the `MigrationHandler` and set the breaking change version to 2.
```suggestion
		"version": 2,
```

sources/ja.senmanga/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+/// Format of `chapterList[].datetime`, e.g. "2026-08-06T12:04:20Z". The offset
+/// has to be read with `XXX` rather than a literal `'Z'`, or the timestamp gets
+/// interpreted in the device timezone instead of utc.
+const DATE_FORMAT: &str = "yyyy-MM-dd'T'HH:mm:ssXXX";
+
+/// Tags that mark an entry as explicit.
+const NSFW_TAGS: [&str; 6] = ["Adult", "Smut", "Lolicon", "Shotacon", "Yaoi", "Yuri"];
+
+/// Tags that mark an entry as suggestive.
+const SUGGESTIVE_TAGS: [&str; 2] = ["Ecchi", "Mature"];
```
kkantan: personally, I believe comments like these make code less readable. it is fairly obviously what all of these variables do, and good code should be self-documenting. I wouldn't mind if they were normal comments, but the unnecessary full sentence documentation comments are a hallmark of ai. that being said, it's still up to you whether you want them or not.

sources/ja.senmanga/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+		ContentRating::NSFW
+	} else if tags
+		.iter()
+		.any(|tag| SUGGESTIVE_TAGS.contains(&tag.as_str()))
+	{
+		ContentRating::Suggestive
+	} else {
+		ContentRating::Safe
+	}
+}
```
kkantan: I don't think making these separate functions provides much benefit. you can just inline it directly in the details updating.

kkantan (changes requested):

sources/ja.senmanga/res/source.json
kinboy56: Fixed

sources/ja.senmanga/src/lib.rs
kinboy56: Sorry

sources/ja.senmanga/src/lib.rs
kinboy56: Fixed

kkantan (approved):

PR #684, opened 2026-08-07
feat: add ja.spoilerplus

### LINKED ISSUE #17: [request] spoilerplus.tv
Lang: JA

https://spoilerplus.tv

kinboy56:
Adds a source for <https://spoilerplus.tv>, a Japanese raw manga site. Fix https://github.com/Aidoku-Community/sources/issues/17

The theme is NetTruyen-based, so the source is written on top of `templates/wpcomics`. Its image infrastructure is shared with `ja.mangarawjp` (same `POST /api/v1/get/c` endpoint, same `img-cdn.stackpathcdn.app` CDN, same page-descrambling scheme), which is the one part the template could not express.

## Template change

`templates/wpcomics` gains a `process_page_image` hook, mirroring `gigaviewer` and `mangareader`: a default `Impl::process_page_image` returning `AidokuError::Unimplemented`, plus a `PageImageProcessor` forward on `WpComics<T>`. Existing sources keep the default; all seven still build unchanged.

## Notes on the implementation

* **Descrambling.** Each page is shuffled into a square grid. The API returns the ordering as hex bytes XORed with the site's own domain string, so `spoilerplus.tv` is the key. The order travels to the image through `PageContext`, and `process_page_image` reassembles the grid on a canvas of the original image's dimensions.
* **`get_page_list` is the only parsing method overridden.** The reader ships empty placeholders and asks for the image list over the API, keyed by ids an inline script declares, so there is no `manga_viewer_page` markup for the template to read. Listing, details, chapters and pagination all run on template defaults with selector overrides.
* **Inline script bodies are read with `html()`.** `Element::data()` returns `None` for every `<script>` on these pages, and the reader's `window.MangaId` / `window.CNumber` values are what key the image API.
* **Keys are stored decoded.** The site percent-encodes series slugs but leaves chapter segments as raw UTF-8, and a raw UTF-8 path is answered with a 404. Since `encode_uri` also escapes `%`, passing the mixed form straight through turns `%20` into `%2520`. Keys are therefore decoded on the way in (`to_key`) and encoded on the way out (`url_for`), so both href forms collapse to one key.
* **Publication status is picked by icon.** The info list gives the alternative-titles row the same `li.row.status` classes as the publication status row, so the template's default `li.status.row > p.col-xs-8` reads the alternative title as the status. The row is selected with `:has(i.fa-rss)` instead; a regression test covers this. An unrecognised value maps to `MangaStatus::Unknown` rather than falling back to `Ongoing`.
* **Tags are anchors, not a delimited string.** Both the genre row and the tag row render as `li.kind`, each value its own `<a>`, so `manga_details_tags` points at the anchors with an empty splitter.
* **`date_uploaded` now comes from the template.** The site mixes relative (`1日前`) and absolute (`2026年07月30日`) forms; `parse_relative_date` already handles the Japanese relative units, and the absolute form is covered by `datetime_format`.
* **`contentRating` is 2 (adult).** The genre list carries permanent `Ecchi` and `エロい` categories, consistent with the sibling sites on the same CMS. `category_parser` is overridden to map the site's Japanese category names.
* **Search takes precedence over the sort filter.** Search has its own endpoint that the ordering paths cannot be applied to, so `config.hidesFiltersWhileSearching` is set and a non-empty query wins over whatever sort value is still stored.
* **`minAppVersion` is 0.7.1**, which is what the template build requires.

## Deliberately left out

* **No genre filter.** `/genre/<name>/` exists, but the list changes on the site's side, so it would need `DynamicFilters`. The template's `get_dynamic_filters` reads a Vietnamese advanced-search page this theme does not ship, so it would need its own implementation. Left for a follow-up.
* **`Home` is not registered.** The template's home layout is built around selectors this theme does not use.

## Test plan

- [x] `cargo fmt`
- [x] `cargo clippy --release` (source and template)
- [x] `cargo test --release` (17 passed)
- [x] `aidoku package && aidoku verify package.aix`
- [x] `cargo clippy --release` on all seven existing `wpcomics` sources
- [x] Verified on device: browsing, ranking, search, details, chapter list and descrambled pages


sources/ja.spoilerplus/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+	fn get_search_manga_list(
+		&self,
+		query: Option<String>,
+		page: i32,
+		filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		// searching has its own endpoint that the ordering paths cannot be applied
+		// to, so a query takes precedence over the sort filter, which the app hides
+		// while searching
+		if let Some(query) = query.filter(|query| !query.is_empty()) {
```
kkantan: I think I've said a few times that the query cannot be empty...
```suggestion
		if let Some(query) = query {
```

sources/ja.spoilerplus/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+		if needs_details {
+			manga.title = clean_title(
+				html.select_first("h1.title-detail")
+					.and_then(|e| e.text())
+					.unwrap_or(manga.title),
+			);
+			manga.cover = html
+				.select_first(".detail-info .col-image img")
+				.and_then(|el| el.attr("src"))
+				.map(|src| absolute_url(&src));
```
kkantan: why do you do this instead of `abs:src`?

sources/ja.spoilerplus/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+}
+
+register_source!(
+	SpoilerPlus,
+	PageImageProcessor,
+	ImageRequestProvider,
+	DeepLinkHandler
+);
+
+#[cfg(test)]
```
kkantan: if you could put these in `test.rs` as well that would be better.

kkantan (changes requested):
> The theme is NetTruyen-based and its
default selectors do match, but the template does not forward
PageImageProcessor, and images here are useless without descrambling.
Changing the template to support that would risk regressions in the seven
sources already using it, so the source is written standalone.

all you would need to do is add a `process_page_image` to the `Impl` and template struct that returns `Err(AidokuError::Unimplemented)`, and then you could implement it in the sources you want. it wouldn't affect any existing sources. if the parsing is the same other than the page image processing, this would be better to do.

sources/ja.spoilerplus/src/lib.rs
kinboy56: Removed

sources/ja.spoilerplus/src/lib.rs
kinboy56: No good reason. The rewrite uses `abs:src` for the details cover and `abs:data-src` for the lazy-loaded listing covers, and the `absolute_url` helper is gone.

sources/ja.spoilerplus/src/lib.rs
kinboy56: Moved

sources/ja.spoilerplus/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+
+fn read_window_number(data: &str, name: &str, fractional: bool) -> Option<String> {
+	let after = &data[data.find(name)? + name.len()..];
+	let after_eq = after[after.find('=')? + 1..].trim_start();
+	let end = after_eq
+		.find(|c: char| !c.is_ascii_digit() && !(fractional && c == '.'))
+		.unwrap_or(after_eq.len());
+	let number = after_eq[..end].trim();
+	(!number.is_empty()).then(|| number.into())
+}
```
kkantan, line 313: these should go in `helpers.rs` I think.

kkantan (changes requested):
it seems like pages fail to load with a "Manga ID not found" error. when opening the chapter url in the web view, it shows a 404 error.

sources/ja.spoilerplus/src/lib.rs
kinboy56: Moved

kkantan (approved):

PR #689, opened 2026-08-10
fix(ru.desu): update source for current API

kurisumakis3:
## Summary
- migrate catalog, search, detail, chapter, and page requests to the current desu.uno API
- update filters and source metadata for the current site
- map the API `reading_mode` field to Aidoku metadata
- refresh the source README for upstream contribution
- add deterministic source contract tests for keys, chapter mapping, and detail metadata

## Verification
- `cargo fmt` / `cargo fmt --check`
- `cargo clippy --target wasm32-unknown-unknown`
- `cargo check --target wasm32-unknown-unknown`
- `cargo test --target wasm32-unknown-unknown` — 3 tests passed
- `aidoku package`
- `aidoku verify`
- runtime smoke test on device: search, filters, title, and chapter opening

## Notes
- The README retains the legacy filter generator as a historical reference. It is explicitly marked as incompatible with the current source and should not be run unchanged.
- This update was prepared with AI assistance and validated against the live API, the repository contribution checklist, and device runtime behavior.


sources/ru.desu/src/helpers.rs
```diff
@@ -101,85 +104,193 @@ pub fn fetch_chapter_pages(manga_id: &str, chapter_id: &str) -> Result<Vec<Strin
...
-				let values: Vec<_> = included
+				let values: Vec<String> = included
 					.into_iter()
-					.chain(excluded.into_iter().map(|x| format!("!{x}")))
+					.chain(excluded.into_iter().map(|value| format!("!{value}")))
 					.collect();
-				if id.eq("genres") || id.eq("tags") {
+				if values.is_empty() {
+					continue;
+				}
```
kkantan: this shouldn't be possible, since a multi select filter will only be passed in if there is at least one value in included or excluded.

sources/ru.desu/src/helpers.rs
```diff
@@ -101,85 +104,193 @@ pub fn fetch_chapter_pages(manga_id: &str, chapter_id: &str) -> Result<Vec<Strin
...
-		))
+		format!("{}/manga/?{query}", get_base_url())
 	}
 }
+
+fn parse_response_html(response: aidoku::imports::net::Response, base: &str) -> Result<Document> {
+	if response.status_code() >= 400 {
+		bail!("HTTP {}", response.status_code());
+	}
+	Ok(Html::parse_with_url(response.get_string()?, base)?)
```
kkantan: why use the string directly?
```suggestion
	response.html()
```
I can't remember if this returns an `AidokuError` or not, so it might need to be:
```rust
	Ok(response.html()?)
```

sources/ru.desu/src/helpers.rs
```diff
@@ -101,85 +104,193 @@ pub fn fetch_chapter_pages(manga_id: &str, chapter_id: &str) -> Result<Vec<Strin
...
+}
+
+pub fn search(query: Option<String>, page: i32, filters: Vec<FilterValue>) -> Result<SearchResult> {
+	if let Some(query) = query.filter(|query| !query.trim().is_empty()) {
```
kkantan: I don't think this is necessary.
```suggestion
	if let Some(query) = query {
```

sources/ru.desu/src/helpers.rs
```diff
@@ -101,85 +104,193 @@ pub fn fetch_chapter_pages(manga_id: &str, chapter_id: &str) -> Result<Vec<Strin
...
+			.unwrap_or_default();
+		return Ok(SearchResult {
+			entries,
+			has_next_page: false,
+		});
+	}
+
+	let url = catalog_url(page, filters);
+	let html = parse_response_html(
+		apply_headers(Request::get(url.clone())?).send()?,
```
kkantan:
```suggestion
		apply_headers(Request::get(&url)?).send()?,
```

sources/ru.desu/src/lib.rs
```diff
@@ -252,6 +248,86 @@ impl DynamicSettings for Desu {
...
+		assert_eq!(ranobe_key("novel.example"), "r:novel.example");
+		assert_eq!(
+			ranobe_slug("/ranobe/novel.example/chapter-1?foo=bar").as_deref(),
+			Some("novel.example")
+		);
+
+		let (section, id) = parse_key("r:novel.example").unwrap();
+		assert!(matches!(section, Section::Ranobe));
+		assert_eq!(id, "novel.example");
+	}
```
kkantan, line 271: if this is true I don't think the `breakingChangeVersion` needs to be incremented.

kkantan (changes requested):

sources/ru.desu/src/helpers.rs
kurisumakis3: fixed, the response is now parsed directly with `response.get_html()?`. the `aidoku-rs` revision used by this source exposes `get_html()` rather than `html()`, so i used the available equivalent.

kurisumakis3:
addressed all review comments in e94332f. local fmt, clippy, check, tests, and package verification pass. `response.get_html()?` is used because the `aidoku-rs` revision in this source does not expose `response.html()`.

kkantan (approved):

remmody:
You've done a great job, of course, but why didn't you fix the light novel catalog?

kurisumakis3:
> You've done a great job, of course, but why didn't you fix the light novel catalog?

when i made this fix, i can say for sure that the light novel catalog was working as long as you were signed in to a desu account. they may have changed something again since then. unfortunately, i don’t have time to work on another fix right now

PR #690, opened 2026-08-10
feat: add ru.senkuro

### LINKED ISSUE #306: senkuro.com
Request to add a source -  https://senkuro.com
Request to add a source -  https://senkuro.me

language: ru
thank you in advance

kurisumakis3:
## Summary

- Add the Russian Senkuro source (`ru.senkuro`).
- Implement Senkuro GraphQL search, manga details, chapter pagination, and reader pages.
- Add web login handling for protected chapter reading.
- Add Home sections, listings, and manga deep links.
- Configure the official Aidoku test runner for source-level WASM tests.

Closes #306

## Verification

- `cargo fmt --check`
- `cargo clippy --target wasm32-unknown-unknown`
- `cargo check --target wasm32-unknown-unknown`
- `cargo test --target wasm32-unknown-unknown` — 1 test passed
- `aidoku package`
- `aidoku verify package.aix`
- Live API probe: search, manga details, chapter pagination, reader pages, and CDN image response.


sources/ru.senkuro/.cargo/config.toml
```diff
@@ -0,0 +1,6 @@
+[build]
+target = "wasm32-unknown-unknown"
+
+[target.wasm32-unknown-unknown]
+runner = "aidoku-test-runner"
+rustflags = ["-C", "link-arg=--export-table"]
```
kkantan: what is this?

sources/ru.senkuro/res/icon.png
kkantan, line 1: can you make the background orange as well so it's not a circle? it would look better with the app that could use any different border radius.

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,960 @@
...
+		}
+
+		let response = request.body(body).send()?;
+		if response.status_code() >= 400 {
+			return Err(error!("Senkuro API: HTTP ошибка"));
+		}
+
+		let envelope = response.get_json_owned::<GraphqlResponse<T>>()?;
+		if let Some(error) = envelope.errors.and_then(|mut errors| errors.pop()) {
+			return Err(error!("Senkuro API: {}", error.message));
```
kkantan:
```suggestion
			bail!("Senkuro API: {}", error.message);
```

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,960 @@
...
+		}
+		if !original.is_empty() {
+			String::from(original)
+		} else {
+			titles
+				.first()
+				.map(|item| item.content.clone())
+				.unwrap_or_default()
+		}
+	}
```
kkantan: I don't think this needs to be two separate functions.

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,960 @@
...
+					.iter()
+					.any(|role| roles.iter().any(|needle| role.contains(needle)))
+			})
+			.map(|member| member.person.name.clone())
+			.collect()
+	}
+
+	fn home_manga(item: HomeManga) -> Manga {
+		Manga {
+			key: item.slug.clone(),
```
kkantan: this doesn't need to be cloned if you define the url above.

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,960 @@
...
+			title: Self::title(&item.original_name, &item.titles),
+			cover: Self::cover(item.cover.as_ref()),
```
kkantan: rather than passing items as references, it's better to consume them since you're already consuming `item`. that way you can avoid cloning unnecessarily.

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,960 @@
...
+		}
+	}
+
+	fn home_manga_page(&self, query: &str, page: i32) -> Result<MangaPageResult> {
+		if page < 1 {
+			return Ok(MangaPageResult {
+				entries: Vec::new(),
+				has_next_page: false,
+			});
+		}
```
kkantan: I don't think this can ever happen.

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,960 @@
...
+		_page: i32,
+		_filters: Vec<FilterValue>,
+	) -> Result<MangaPageResult> {
+		let query = query.unwrap_or_default();
+		if query.trim().is_empty() {
+			return Ok(MangaPageResult {
+				entries: Vec::new(),
+				has_next_page: false,
+			});
+		}
```
kkantan: is it really necessary to check if it's whitespace? I assume the graphql response would be fine.
```suggestion
		let Some(query) = query else {
			return Ok(MangaPageResult::default());
		};
```

sources/ru.senkuro/src/lib.rs
kkantan, line 1: this could be refactored into at least:
- `models.rs`: serializable structs
- `graphql.rs`: graphql queries and helper function to make a graphql request
- `helpers.rs`: extra helper functions

and optionally (some other sources do this):
- `home.rs`
- `tests.rs`

though it's fine to leave the relevant code in `lib.rs`/`helpers.rs`.

kkantan (changes requested):

sources/ru.senkuro/.cargo/config.toml
kurisumakis3: removed this. --export-table was not required for the source test runner. cargo test --target wasm32-unknown-unknown passes without it.

sources/ru.senkuro/res/icon.png
kurisumakis3: fixed. the icon now uses a solid orange background instead of an orange circle on white. it was verified with aidoku verify and on device.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed in 8c7c325.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed. the two title helpers are now merged, with the different graphql field shapes handled at the call sites.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed. the url is created before moving item.slug into the key.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed. the owned cover value is now consumed directly, avoiding an unnecessary image url clone.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed. removed the unreachable page < 1 branch.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed. now uses the suggested let Some(query) = query else path.

sources/ru.senkuro/src/lib.rs
kurisumakis3: i kept the related code in lib.rs for now, since the proposed module split is optional. the existing structure remains valid and all checks pass.

kurisumakis3:
addressed the review comments in 8c7c325. removed the unnecessary export-table flag, simplified title, cover, pagination, and query handling, and updated the icon. local fmt, clippy, check, test, package, and verify gates pass. runtime smoke testing passed for home, search, manga details, chapters, and reader pages. deep links remain unverified because the altstore installation does not register aidoku in safari share sheet. the readme also notes that restarting aidoku may be required after web login and that api access may depend on the network route or vpn exit region.

sources/ru.senkuro/src/lib.rs
kkantan: the last two are optional. I think the first three should be done unless you can justify having a ~1000 line file which is not very readable.

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+			.manga_recommendations
+			.into_iter()
+			.map(Self::home_manga)
+			.collect())
+	}
+
+	fn search_manga(&self, item: SearchManga) -> Manga {
+		let title = Self::title(&item.original_name, &item.titles);
+		Manga {
+			key: item.slug.clone(),
```
kkantan, line 547: I commented on this in only one place but it's true for multiple places. manga and chapter slugs don't need to be cloned during struct initialization.

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+		match listing.id.as_str() {
+			"latest-updates" => self.home_manga_page(HOME_LATEST_QUERY, page),
+			"new-titles" => self.home_manga_page(HOME_NEW_QUERY, page),
+			"popular-day" => {
+				if page != 1 {
+					return Ok(MangaPageResult {
+						entries: Vec::new(),
+						has_next_page: false,
+					});
+				}
```
kkantan, line 761: I think I already said before that this is not possible.

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+				Ok(MangaPageResult {
+					entries: self.home_recommendations()?,
+					has_next_page: false,
+				})
+			}
+			_ => Err(error!("Senkuro: неизвестная Home-секция")),
```
kkantan, line 779:
```suggestion
			_ => bail!("Senkuro: неизвестная Home-секция"),
```

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+}
+
+impl Home for Senkuro {
+	fn get_home(&self) -> Result<HomeLayout> {
+		let latest = self.home_manga_page(HOME_LATEST_QUERY, 1)?.entries;
+		let popular = self.home_static_list(HOME_POPULAR_QUERY)?;
+		let new_titles = self.home_manga_page(HOME_NEW_QUERY, 1)?.entries;
+
+		if latest.is_empty() || popular.is_empty() || new_titles.is_empty() {
+			return Err(error!("Senkuro: Home не вернула обязательные секции"));
```
kkantan, line 791:
```suggestion
			bail!("Senkuro: Home не вернула обязательные секции");
```

sources/ru.senkuro/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+
+		Ok(HomeLayout { components })
+	}
+}
+
+impl DeepLinkHandler for Senkuro {
+	fn handle_deep_link(&self, url: String) -> Result<Option<DeepLinkResult>> {
+		let Some(path) = url
+			.strip_prefix(BASE_URL)
+			.or_else(|| url.strip_prefix("https://senkuro.com"))
```
kkantan, line 828: this will not happen because only urls that are prefixed by the urls you set in `source.json` will be passed to this function.

kkantan (changes requested):

sources/ru.senkuro/src/lib.rs
kurisumakis3: implemented in d0254af: split the source into models.rs, graphql.rs, and helpers.rs. home.rs and tests.rs were left out as optional.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed in d0254af: owned manga and chapter slugs now move directly into keys, and the pagination cursor no longer needs a clone.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed in d0254af: finite popular-day and recommendations listings now return their data directly with has_next_page set to false. latest-updates and new-titles keep cursor pagination.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed in d0254af: unknown Home sections now use the requested bail! error.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed in d0254af: missing required Home sections now use the requested bail! error.

sources/ru.senkuro/src/lib.rs
kurisumakis3: fixed in d0254af: removed the unreachable senkuro.com alias. Deep links now match the senkuro.me domain declared in source.json.

kurisumakis3:
implemented in d0254af.

- split the source into models.rs, graphql.rs, and helpers.rs.
- removed the unnecessary slug and cursor clones.
- removed the finite-listing page guards.
- switched the requested Home errors to bail!.
- removed the unreachable senkuro.com deep-link alias.

verification:
- cargo fmt --check
- cargo clippy --target wasm32-unknown-unknown
- cargo check --target wasm32-unknown-unknown
- cargo test --target wasm32-unknown-unknown (1 passed)
- git diff --check
- aidoku package and aidoku verify package.aix
- manual Aidoku smoke test: Home, search, details, chapters, reader, and images

all GraphQL query bodies and serde contracts were preserved.

sources/ru.senkuro/src/graphql.rs
```diff
@@ -0,0 +1,212 @@
...
+			.header("Origin", BASE_URL)
+			.header("Referer", BASE_URL)
+			.header("User-Agent", USER_AGENT);
+		if !cookie.is_empty() {
+			request = request.header("Cookie", &cookie);
+		}
+
+		let response = request.body(body).send()?;
+		if response.status_code() >= 400 {
+			return Err(error!("Senkuro API: HTTP ошибка"));
```
kkantan, line 200:
```suggestion
			bail!("Senkuro API: HTTP ошибка");
```

kkantan (approved):

PR #697, opened 2026-08-16
feat(ja.soraraw): split stacked chapter images into pages

kinboy56:
Some chapters here ship as one image with every page stacked on top of each other, up to 49152
pixels tall. Version 2 refused those with an error. This splits them instead.

- `get_page_list` measures the image header and expands a stacked image into one page per stacked
  page. The extension doesn't say whether the file is jpeg or webp, so the magic bytes decide.
- Only chapters of four images or fewer are measured. Longer chapters take no extra request.
- `PageImageProcessor` cuts the slice each page asks for, on whole pixel rows.
- A count past 64 pages is read as a bad header rather than an image that deep, and is left whole.
- Pages that carry no slice pass through the processor untouched.

`minAppVersion` is `0.9.0` and the source version `3`. Aidoku 0.9 carries the two fixes this
depends on:

- AidokuRunner `e44d774`: drawing into a canvas shorter than the source image.
- Aidoku `3fd2913`: pages sharing a url and differing only by their `PageContext` all rendered the
  same image.

`3fd2913` also let the `?slice=N` workaround of the earlier revision go. The reader now fetches a
stacked image once per chapter. Downloads still fetch it once per page, since `DownloadTask`
requests each page on its own.

Device: iPhone 15, iOS 26.1 and the current revision on Aidoku 0.9.


kkantan:
since this depends on a fix that is not in a public app version, we should wait to merge until the next app release. the minimum app version should be set to whatever the next version release is.

kinboy56:
@kkantan 
Hi, Aidoku version 0.9 has been released. 
Could you please take a look when you have a chance? 
Thank you.

sources/ja.soraraw/res/source.json
```diff
@@ -2,13 +2,13 @@
...
 		"name": "Soraraw",
-		"version": 2,
+		"version": 3,
 		"url": "https://soraraw.com",
 		"contentRating": 2,
 		"languages": [
 			"ja"
 		],
-		"minAppVersion": "0.7.1"
+		"minAppVersion": "0.9.0"
```
kkantan: shouldn't this be `0.9` instead of `0.9.0`? the app version itself is set as `0.9`, but I don't know how it handles semantic versions.
```suggestion
		"minAppVersion": "0.9"
```

kinboy56: Changed to `0.9`.

kkantan (approved):

PR #699, opened 2026-08-16
fix(en.asurascans): Fix app crashing due to premium chapters

### LINKED ISSUE #698: Crash after refreshing Asuna Scans title
As the title says. After the refresh, Unknown Error appears and app crashes after a few seconds. Latest Aidoku version.

jianmingyong:
Fixes #698 

Changes to premium chapters handling and also a bug with the css selector for manhwa status.

You can test it here:
[package.zip](https://github.com/user-attachments/files/31117484/package.zip)

FYI: This source is written poorly that has a lot of potential place where panics can cause the app to crash. I am not going to make that rewrite as that takes up time. For now this is a quick fix.

### DIFF
--- a/sources/en.asurascans/res/source.json
+++ b/sources/en.asurascans/res/source.json
@@ -2,7 +2,7 @@
 	"info": {
 		"id": "en.asurascans",
 		"name": "Asura Scans",
-		"version": 18,
+		"version": 19,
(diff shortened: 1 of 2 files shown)


kkantan (approved):
the only place with panics I see is the indexing for the chapter object after it's converted to a map. indexing serde values is safe.

Issue #695, opened 2026-08-15
[IOS 26.6] CloudFlare related sources will stop working


jianmingyong:
Ref: https://github.com/Aidoku/Aidoku/issues/1034

All sources that is backed by CF challenge will stop working once you update your device to ios 26.6 and above.

There is currently no easy solution for this other than the ugly webview workaround (which cannot handle captcha in any shape or form…)

I have delayed updating for quite a long while so I never hit the CF issues people are talking about. So it seems confirmed that issues started after updating the device to latest ios version causing some problems with native request not working properly.

If you can avoid updating, which I think is too late for majority of the people, please do so.

- - -

Please install v0.9 aidoku for the bug fix.

PR #706, opened 2026-08-19
fix(multi.mangadotnet): Fix search function

### LINKED ISSUE #703: MangaDotNet search function is broken
The website have updated their frontend apis and thus it isn’t working right now.

This fix is scheduled to be added with app update.

jianmingyong:
This fixes the search function. The API is a bit ugly as it is way more complicated. I am still finding ways to optimize it...

Content filters aren't working yet. The new search API doesn't tie the same function with the global 18+ switch. I need to work on new defaults.

Fixes: #703 
This update is blocked by: #695

### DIFF
--- a/sources/multi.mangadotnet/src/helpers.rs
+++ b/sources/multi.mangadotnet/src/helpers.rs
@@ -172,47 +172,44 @@ pub fn base_url_join(uri: &str) -> String {
 
 fn is_official_like(chapter: &MangaChapter) -> bool {
 	let official_groups_names = [
-		"Official",
-		"Official?",
-		"Animate International",
-		"Comikey",
-		"Dark Horse Manga",
-		"FAKKU",
-		"J-Novel Club",
-		"K-Manga",
-		"K Manga",
-		"Kodansha USA",
-		"Kodansha Comics",
-		"Manga Plus",
-		"MangaPlus",
-		"Manga UP!",
-		"One Peace Books",
-		"Seven Seas Entertainment",
-		"Square Enix Manga",
-		"Viz Manga",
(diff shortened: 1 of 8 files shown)


sources/multi.mangadotnet/src/settings.rs
kkantan, line 1: is there a particular reason why you store the cursor persistently with defaults? to me, it would make more sense to just have a property on the source struct to set, since if the app is closed it won't need to load any past cursor data.

sources/multi.mangadotnet/src/helpers.rs
```diff
@@ -171,43 +171,45 @@ pub fn base_url_join(uri: &str) -> String {
...
-	let official_scanlator_names = ["Official", "Official?", "MangaPlus", "Comikey", "K-Manga"];
-
-	let group_ids = chapter
-		.groups
-		.as_ref()
-		.is_some_and(|groups| groups.iter().any(|g| official_group_ids.contains(&g.id)));
+	let group_ids = chapter.groups.as_ref().is_some_and(|groups| {
+		groups.iter().any(|g| {
+			official_groups_names
+				.map(|name| name.to_lowercase())
```
kkantan: it seems like it would be better to make the static list already lowercase so you don't need to allocate so many extra strings, as the group names appear to only be used after making them lowercase.

jianmingyong:
Considering the current CF settings being lowered, forced app updates aren’t needed anymore.

At least people should reasonably be able to use the source. (Search doesn’t not follow the nsfw settings :()

kkantan (approved):

PR #707, opened 2026-08-19
feat(en.stonescape): add thumbnails to chapters

splat58:


kkantan (changes requested):
the source version in `source.json` should be incremented by one so that this shows as an update.

kkantan (approved):
I'm not sure why the workflow isn't completing but this should be fine.

PR #724, opened 2026-08-24
feat(en.novelbuddy): render chapter content as Markdown

BaconDroid:
### Chapter content rendered as Markdown
Chapter HTML from the API is now converted to Markdown instead of being flattened to plain text, preserving bold/italic/underline/strikethrough, headings, lists, blockquotes, inline code, links, images and horizontal rules — without ever exposing raw tags. Root-level text and inline elements are preserved as well. The approach is adapted from `en.freewebnovel`'s chapter converter and the shared libgroup template converter. Descriptions intentionally remain plain text.

Verified against live API content (e.g. Shadow Slave ch. 1): bracketed system messages such as `[Aspirant! Welcome to the Nightmare Spell…]` survive as literal text instead of turning into links, empty placement divs emit nothing, and code/pre content keeps its characters verbatim.

### Fixes
- **Chapter upload dates were silently dropped**: the previous format string never matched the API's `Z`-suffixed timestamps. Parsing now uses the quoted-literal pattern proven in asurascans/ezmanga/kagane, with a naive-UTC fallback, and a live-test assertion covers it.
- **Content rating** is now determined solely by the API's `isAdult` flag.
- **Minimum app compatibility** now declares Aidoku `0.8.3`, matching the HTML API imports required by the compiled source.

### Tests
Expanded from 9 to 24: converter coverage (inline styles, headings and breaks, lists — including ordered numbering that ignores non-`li` children — multi-line blockquotes, code/links/images/hr, literal-markdown escaping, verbatim code/pre content, empty placement divs, inline spans, root-level text and inline elements, naive timestamp fallback) plus regression coverage for each fix above. All tests run against the live API (`cargo test`: 24 passed); `aidoku package` and `aidoku verify` both pass.

### Commits
Split per change for reviewability:

1. `fix(en.novelbuddy): rate content solely from the API adult flag`
2. `fix(en.novelbuddy): parse chapter upload timestamps reliably`
3. `feat(en.novelbuddy): render chapter content as Markdown`
4. `chore(en.novelbuddy): bump version to 6 and require Aidoku 0.8.3`


sources/en.novelbuddy/src/helpers.rs
```diff
@@ -201,6 +205,192 @@ pub fn html_to_text(html: &str) -> String {
 		.unwrap_or_default()
 }
 
+fn escape_markdown(text: &str, output: &mut String) {
```
kkantan: could you use https://github.com/Aidoku/aidoku-rs/blob/a386ee40ab79896549bb9514657ea957e6625be5/crates/lib/src/helpers/string.rs#L16 instead?

sources/en.novelbuddy/src/lib.rs
```diff
@@ -279,8 +290,15 @@ mod tests {
 			key: "VYPGVZ8z".into(),
 			..Default::default()
 		};
+		// Resolve a real chapter key dynamically: hardcoded keys rot as the
+		// API re-publishes chapters.
```
kkantan: aidoku keys should be static so history is persistent. if this is true, maybe keys should be constructed differently.

kkantan (changes requested):
I would put the markdown conversion code in a new `markdown.rs` module since it's independent of the source.

sources/en.novelbuddy/src/helpers.rs
BaconDroid: Thanks for the pointer! Adopted `PlainText::escape_markdown` from aidoku-rs (via the `helpers` feature) and removed the custom escaper — its broader punctuation coverage also caught a couple of characters my list missed.

sources/en.novelbuddy/src/lib.rs
BaconDroid: Thanks for flagging this — you're right that key stability matters for reading history. I checked against the live API: chapter ids have been stable across several days for unchanged chapters (verified by comparing fetched lists), and the detail endpoint only resolves ids (a slug-based lookup returns NOT_FOUND), so keys have to remain API ids. The one id that stopped resolving corresponded to a chapter the site itself removed/re-published. I've reworded the misleading test comment accordingly.

BaconDroid:
Thanks @kkantan for the review — all three points are addressed in 53c23c9:

- The markdown conversion now lives in its own `markdown.rs` module, independent of the source.
- Text escaping uses `PlainText::escape_markdown` from aidoku-rs instead of a custom escaper.
- The chapter key resolution comment was reworded: ids verified stable across several days for unchanged chapters, the detail endpoint only resolves ids (slug lookup returns NOT_FOUND), so keys remain API ids; the test resolves from the live list since the site occasionally removes or re-publishes chapters.

All 24 tests pass against the live API, and `aidoku package` / `aidoku verify` are green.

sources/en.novelbuddy/src/lib.rs
```diff
@@ -279,8 +292,15 @@ mod tests {
 			key: "VYPGVZ8z".into(),
 			..Default::default()
 		};
+		// Resolve a chapter key from the live list rather than hardcoding
+		// one: the site occasionally removes or re-publishes chapters.
+		let chapters = fetch_chapter_list(&manga.key).expect("chapter list failed");
+		let key = chapters
+			.last()
+			.map(|c| c.key.clone())
```
kkantan, line 300: I don't think this clone is necessary since you're consuming `chapters`, but it doesn't matter for a test.

kkantan (approved):

PR #727, opened 2026-08-28
fix(mangareader): resolve lazy-loaded covers

at-wr:
## Summary

- resolve MangaReader listing and home covers through the template's existing lazy-image helper
- prefer `data-lazy-src`, `data-src`, and `data-url` before placeholder `src` values
- cover the precedence with a host unit test and a full Aidoku HTML parser fixture

MangaMura currently emits a one-pixel GIF in `src` and the actual CDN cover in `data-src`. Direct `src` reads therefore produce transparent listing thumbnails even though details and reader images work. Reusing one helper across every MangaReader cover path fixes the template without adding per-entry detail requests.

## Verification

- `cargo fmt --check` in `templates/mangareader`
- `cargo test --target aarch64-apple-darwin` (1 passed)
- `cargo test --no-run` (Wasm fixture compiles)
- `cargo clippy -- -D warnings` in `templates/mangareader`
- `cargo test --no-run`, `cargo clippy -- -D warnings`, and `cargo build --release` in `sources/ja.mangamura`
- live MangaMura homepage HTML rechecked with placeholder `src` plus real `data-src` covers

### DIFF
--- a/templates/mangareader/src/helper.rs
+++ b/templates/mangareader/src/helper.rs
@@ -4,12 +4,36 @@ pub trait ElementImageAttr {
 	fn img_attr(&self) -> Option<String>;
 }
 
+fn preferred_image_attr(mut attr: impl FnMut(&str) -> Option<String>) -> Option<String> {
+	[
+		"abs:data-lazy-src",
+		"abs:data-src",
+		"abs:data-url",
+		"abs:src",
+		"data-url",
+	]
+	.into_iter()
+	.find_map(&mut attr)
+}
+
 impl ElementImageAttr for Element {
 	fn img_attr(&self) -> Option<String> {
-		self.attr("abs:data-lazy-src")
-			.or_else(|| self.attr("abs:data-src"))
-			.or_else(|| self.attr("abs:data-url"))
-			.or_else(|| self.attr("abs:src"))
-			.or_else(|| self.attr("data-url"))
+		preferred_image_attr(|name| self.attr(name))
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn lazy_cover_precedes_placeholder_src() {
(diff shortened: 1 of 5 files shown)


templates/mangareader/src/helper.rs
```diff
@@ -4,12 +4,36 @@ pub trait ElementImageAttr {
...
+	[
+		"abs:data-lazy-src",
+		"abs:data-src",
+		"abs:data-url",
+		"abs:src",
+		"data-url",
+	]
+	.into_iter()
+	.find_map(&mut attr)
+}
```
kkantan: I don't think this new implementation is necessary, since it appears to be the same logic but somewhat more abstracted, in a way that doesn't really make sense.

templates/mangareader/src/parser.rs
```diff
@@ -185,11 +185,46 @@ pub fn parse_manga_list(html: &Document, base_url: &str) -> Vec<Manga> {
...
 }
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use aidoku::imports::html::Html;
+	use aidoku_test::aidoku_test;
+
+	#[aidoku_test]
+	fn manga_list_prefers_lazy_loaded_cover() {
```
kkantan: I think maintaining tests like these can be annoying since selectors are prone to changing, and it's not clear what input the `parse_manga_list` function requires while testing. anyways, the change is small enough that I don't think it really warrants tests.

kkantan (changes requested):
the version numbers for mangareader sources should also be incremented so that this change shows as an update for users.

at-wr:
Addressed the requested version bump in a951608: Manga Mura is the only source currently depending on the mangareader template, and its source version is now 3. I also removed the abstraction and selector tests called out in the inline review. Ready for another look.

kkantan (approved):

PR #728, opened 2026-08-30
fix(multi.kagane): rank search results by relevance

Amqx:
Searching always added `sort=total_views,desc` to a request when no sort filters are added, which replaced the api's own relevance sorting. This is bad because the api then returns every series whose title somewhat matched the search term, ordered by popularity, so an exact match can land very deep inside the results. Example series I used was "The Villainess Lives Again", which has 3 letter to letter matches on site, but the source doesn't return a single one in app. The intended behaviour should be to leave the entire sort parameter out so the source returns by Relevance (new sort filter), which causes the three matches to actually appear at the top.

While I was here I also added the other missing sort filters the api accepts:  Popular (Today), Popular (Week), Popular (Month), By Name, and Books Count.

And I defaulted search to relevance so searching is a little more reliable.

### DIFF
--- a/sources/multi.kagane/res/filters.json
+++ b/sources/multi.kagane/res/filters.json
@@ -4,7 +4,18 @@
 		"id": "sort",
 		"title": "Sort By",
 		"canAscend": false,
-		"options": ["Popular (Total Views)", "Recently Updated", "Recently Added"],
+		"options": [
+			"Relevance",
+			"Popular (Total Views)",
+			"Popular (Average Views)",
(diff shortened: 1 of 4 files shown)


kkantan (approved):
the source does not load for me, but I believe this is due to #695. if this change works for you then it should be okay to merge.

Amqx:
I have no clue what that's about, I'm on nightly (albeit personal fork, but no changes to source handling) and iOS27 and every cf source works for me

PR #735, opened 2026-09-02
feat: add multi.xcomic

PoppingMangoSources:
Add XCOMIC with multilingual GraphQL search, filters, listings, details, chapters, pages, home sections, mirror selection, and deep links.

Use Aidoku selected languages for browse and latest-upload feeds, map XCOMIC regional language codes, label chapters, and filter coverless placeholders.


sources/multi.xcomic/res/filters.json
```diff
@@ -0,0 +1,243 @@
+[
+	{
+		"type": "sort",
+		"id": "sort",
+		"canAscend": false,
+		"default": {
+			"index": 0,
+			"ascending": false
+		},
```
kkantan: this is already the default and doesn't need to be explicitly set.

sources/multi.xcomic/res/icon.png
kkantan, line 1: I think it would be nicer to shrink this down a little to have it be more like the favicon that's displayed.

sources/multi.xcomic/src/filters.rs
```diff
@@ -0,0 +1,129 @@
...
+				false,
+				owned(LANGUAGES),
+			),
+			multi_select(
+				"translated_languages",
+				"Translated Languages",
+				false,
+				false,
+				owned(LANGUAGES),
+			),
```
kkantan: I'm not convinced this filter needs to exist when the source already has the configurable languages setting.

sources/multi.xcomic/src/filters.rs
```diff
@@ -0,0 +1,129 @@
...
+	}
+	.into()
+}
+
+impl DynamicFilters for XComic {
+	fn get_dynamic_filters(&self) -> Result<Vec<Filter>> {
+		let genres = fetch_genres(&self.get_base_url()?).unwrap_or_else(|| owned(GENRES));
+		// The api only takes languages as includes.
+		Ok(vec![
+			multi_select("genres", "Genres", true, true, genres),
```
kkantan, line 88: It might be better to set this filter to use the tag style, to match with other source genre filters.

sources/multi.xcomic/res/source.json
```diff
@@ -0,0 +1,183 @@
...
+			"uk",
+			"ur",
+			"uz",
+			"vi",
+			"yo",
+			"zu"
+		],
+		"minAppVersion": "0.7.1"
+	},
+	"listings": [
```
kkantan, line 122: I think most of these don't need to exist, considering the website itself does not have any listings. these are just using the search sort filter.

PoppingMangoSources:
Site has changed a few things, i'll be fixing things up to reflect that + cleaning up the code when i'm done, hence why i've marked it as a draft for now.

sources/multi.xcomic/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+const DEFAULT_BASE_URL: &str = "https://xcomic.me";
+
+/// Manga key for a url target, looking up the series' editions when needed.
+fn comic_key(base_url: &str, target: helpers::Target) -> Result<String> {
+	match target {
+		helpers::Target::Comic(key, _) => Ok(key),
+		helpers::Target::Title(id) => helpers::resolve_title(base_url, &id)
+			.ok_or_else(|| error!("No comics are available for this series")),
+	}
+}
```
kkantan: I think this should be put in `helpers.rs` too.

sources/multi.xcomic/src/graphql.rs
```diff
@@ -0,0 +1,423 @@
...
+	pub upload_status: String,
+	pub chapter_count: String,
+	pub year_min: Option<i64>,
+	pub year_max: Option<i64>,
+}
+
+impl BrowseParams {
+	pub fn new(sortby: &str, page: i32) -> Result<Self> {
+		let excluded_genres = defaults_get::<Vec<String>>("excludedGenres").unwrap_or_default();
+		Ok(Self {
```
kkantan, line 185: instead of defining all these `String::new()` and `Vec::new()`s, you could `#[derive(Default)]` on the struct and then use `..Default::default()` to fill in default values.

sources/multi.xcomic/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+			return Ok(MangaPageResult {
+				entries: vec![manga_from_data(comic, &base_url)],
+				has_next_page: false,
+			});
+		}
+		let mut params = BrowseParams::new("field_score", page)?;
+		params.word = query.unwrap_or_default();
+		let mut types = Vec::new();
+		let mut ratings = Vec::new();
+		let mut translated = Vec::new();
```
kkantan: why are these separated from params?

sources/multi.xcomic/src/graphql.rs
```diff
@@ -0,0 +1,423 @@
...
+	pub original_status: String,
+	pub upload_status: String,
+	pub chapter_count: String,
+	pub year_min: Option<i64>,
+	pub year_max: Option<i64>,
+}
+
+impl BrowseParams {
+	pub fn new(sortby: &str, page: i32) -> Result<Self> {
+		let excluded_genres = defaults_get::<Vec<String>>("excludedGenres").unwrap_or_default();
```
kkantan: it would be nice to have a function for this in `settings.rs` since you already have a file.

kkantan (changes requested):
it looks like the search items return tags and descriptions for everything, which is usually not a good idea. you should only return the key, title, and cover, and optionally status and content rating, and then fill in the rest when fetching details.

PoppingMangoSources:
Addressed all the review points, compact search payloads, tag-style genres, translated-languages filter dropped, listings cut to four, comic_key moved to helpers, BrowseParams derives Default, settings lookups moved to settings.rs, redundant sort default removed, icon shrunk.

sources/multi.xcomic/src/home.rs
```diff
@@ -0,0 +1,163 @@
...
+						entries: top_rated,
+						auto_scroll_interval: Some(6.0),
+					},
+				},
+				HomeComponent {
+					title: Some("Most Viewed (30 Days)".into()),
+					subtitle: None,
+					value: HomeComponentValue::MangaList {
+						ranking: true,
+						page_size: Some(10),
```
kkantan: I think these page sizes are a little too big, which makes using it awkward on smaller phones. you need to scroll down to the bottom to see the 10th series, then you swipe to the next page and have to scroll up again to see the 11th. I would set to 5 max, for all of the components.

kkantan (changes requested):
when I try and open the web view urls for series and chapters, I get a 404 error. otherwise, the source looks good.

PoppingMangoSources:
Lowered all component page sizes to 5. I couldn’t reproduce the web view 404s, but the URLs relied entirely on the API’s urlPath with no fallback, both now fall back to the site’s own route shapes, so let me know if you still hit it.

kkantan (approved):

PR #739, opened 2026-09-06
feat(mangareader): split stacked chapter images into pages

kinboy56:
mangamura.me and rawotaku.com serve most chapters as one tall jpeg with every page stacked into
it. The template now reads the size off the header and hands the reader one slice at a time.

A page stands about 1.42 times its width on both sites. The ratio alone rounds 1403x38912 up to
20 pages and cuts every one of them, so the count falls back on the neighbour that divides the
image evenly: 19 leaves a page of exactly 2048. Only chapters of four images or fewer are
measured, so a chapter with a page per image pays for no extra requests.

Three more fixes came out of testing on device.

- The detail title selector is `.manga-name` on both sites, not `.manga_name`, so titles came
  back empty.
- mangamura has no `/home`. The same layout sits at the root, so the path is a param now.
- The chapter name only ever repeats the number, either bare, with a volume suffix
  (`第12.2話-v23`) or wrapped in the seo heading (`約束のネバーランド (Raw – Free) 【第181.6話】`).
  Across 118 chapters from 6 manga none carried a title of its own, so both sources turn chapter
  titles off.

`minAppVersion` is 0.9 because the canvas the slicing draws into needs the AidokuRunner fix in
e44d774.

Verified on device with a local source list.

### DIFF
--- a/sources/ja.mangamura/src/lib.rs
+++ b/sources/ja.mangamura/src/lib.rs
@@ -3,10 +3,7 @@ use aidoku::{Source, alloc::borrow::Cow, prelude::*};
 use mangareader::{Impl, MangaReader, Params};
 
 const BASE_URL: &str = "https://mangamura.me";
-// some chapters ship as one tall jpeg stacking every page into it. a page stands about 1.42 times
-// its width here: across 33 stacked images the ratio ran from 1.39 to 1.44, and the nominal b5 √2
-// rounds 1115x56000 up to 36 pages, where 56000/35 lands on a whole 1600 like every other page
-// height read off the site
+// some chapters ship as one tall jpeg; across 33 of them a page ran 1.39 to 1.44 times its width
 const PAGE_ASPECT: f32 = 1.42;
 
 struct MangaMura;
@@ -26,8 +23,7 @@ impl Impl for MangaMura {
 			page_param: "p".into(),
 			get_chapter_selector: || "#ja-chaps > li".into(),
 			get_chapter_language: |_| "ja".into(),
-			// the chapter name only ever repeats the number, sometimes with a volume suffix or the
-			// seo heading wrapped around it
+			// chapter names only ever repeat the number
 			has_chapter_titles: false,
 			get_page_url_path: |chapter_id| format!("/json/chapter?id={chapter_id}&mode=vertical"),
 			set_default_filters: |query_params| {
--- a/sources/ja.mangamura/src/lib.rs
+++ b/sources/ja.mangamura/src/lib.rs
@@ -3,6 +3,11 @@ use aidoku::{Source, alloc::borrow::Cow, prelude::*};
 use mangareader::{Impl, MangaReader, Params};
 
 const BASE_URL: &str = "https://mangamura.me";
+// some chapters ship as one tall jpeg stacking every page into it. a page stands about 1.42 times
+// its width here: across 33 stacked images the ratio ran from 1.39 to 1.44, and the nominal b5 √2
+// rounds 1115x56000 up to 36 pages, where 56000/35 lands on a whole 1600 like every other page
+// height read off the site
+const PAGE_ASPECT: f32 = 1.42;
 
 struct MangaMura;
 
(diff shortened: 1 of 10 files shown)


templates/mangareader/src/imp.rs
```diff
@@ -147,27 +155,56 @@ pub trait Impl {
...
+			let slices = match params.stacked_page_ratio {
+				Some(ratio) if measure => stacked_page_count(&url, ratio),
+				_ => 1,
+			};
+
+			if slices < 2 {
+				pages.push(Page {
+					content: if shuffled {
+						let mut context = PageContext::default();
+						context.insert("shuffled".into(), "1".into());
```
kkantan: shuffled does not need to be included in the context since it is not handled in the page image processor. this was for #11 to implement, but mangareader.io was removed so it is no longer necessary.

sources/ja.mangamura/src/lib.rs
```diff
@@ -3,6 +3,11 @@ use aidoku::{Source, alloc::borrow::Cow, prelude::*};
 use mangareader::{Impl, MangaReader, Params};
 
 const BASE_URL: &str = "https://mangamura.me";
+// some chapters ship as one tall jpeg stacking every page into it. a page stands about 1.42 times
+// its width here: across 33 stacked images the ratio ran from 1.39 to 1.44, and the nominal b5 √2
+// rounds 1115x56000 up to 36 pages, where 56000/35 lands on a whole 1600 like every other page
+// height read off the site
```
kkantan: I this this is too much explanation using unnatural language.

templates/mangareader/src/helper.rs
```diff
@@ -13,3 +23,79 @@ impl ElementImageAttr for Element {
...
+	}
+
+	// the site cuts its pages at a whole pixel, so the neighbour the ratio leans towards beats it
+	// when the image divides evenly there: 1403x38912 sits at 19.53 pages and rounds to 20, where
+	// 19 leaves a page of exactly 2048 and 20 cuts every one of them
+	let neighbour = if pages < rounded as f32 {
+		rounded - 1
+	} else {
+		rounded + 1
+	};
```
kkantan: these explanations are not necessary.

```suggestion
	let rounded = (pages + 0.5) as u32;
	if !(2..=STACKED_PAGE_LIMIT).contains(&rounded) {
		return 1;
	}
	let neighbour = if pages < rounded as f32 {
		rounded - 1
	} else {
		rounded + 1
	};
```

kkantan (changes requested):
can the splitting ever cause issues? are there pages that have incorrect ratios? I wonder if making it a setting, which is on by default, would be a good choice so that users could disable it.

this isn't strictly necessary though, just a suggestion if there is reasonable enough suspicion there could be issues.

templates/mangareader/src/imp.rs
kinboy56: Removed.

sources/ja.mangamura/src/lib.rs
kinboy56: sorry

templates/mangareader/src/helper.rs
kinboy56: sorry

kinboy56:
> can the splitting ever cause issues? are there pages that have incorrect ratios? I wonder if making it a setting, which is on by default, would be a good choice so that users could disable it.
> 
> this isn't strictly necessary though, just a suggestion if there is reasonable enough suspicion there could be issues.

I surveyed the most-viewed listings on both sites before deciding: 155 titles, 906 chapters, 20974 images. 197 of those images sat in chapters of four images or fewer, so they were measured. 193 were split and 4 were left alone. The 4 were one page-per-image chapter that slice_count correctly returned 1 for.

I also looked at every split whose page ratio fell outside 1.38-1.45, six in total. All were correct, either genuinely tall pages or a high slice count. No incorrect split turned up.

The failure mode is one-sided too: when the JPEG header cannot be read the image is passed through unsplit, which is what happens today without this PR.

So I left the setting out. I can still add one if you would rather have it.

kkantan (approved):

PR #742, opened 2026-09-06
feat(ja.spoilerplus): expose the site tabs as listings

kinboy56:
The site serves its browse tabs from separate paths, so they fit listings better than a sort filter.

- `latest`, `trending` and `ranking` replace the sort filter, which only carried two of the three.
- `trending` has no pager, so that listing stays a single page.
- The wpcomics template gains `get_listing_url`. Sources that leave it unset keep returning `Unimplemented`.

### DIFF
--- a/sources/ja.spoilerplus/res/filters.json
+++ b/sources/ja.spoilerplus/res/filters.json
@@ -1,8 +1,0 @@
-[
-	{
-		"type": "sort",
-		"id": "sort",
-		"canAscend": false,
-		"options": ["最近の更新", "ランキング"]
-	}
-]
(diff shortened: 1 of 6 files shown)


kkantan (commented):
I don't know if I agree with this change. I think listings are good when the source also implements a home page, but when it doesn't, I think the usage of the sort filter is superior. users can set a default sort when opening the source, and the search bar is immediately visible without having to press a button.

kinboy56:
You're right — I misread how the app handles this.

I checked the app source. Filter state is persisted per source (`<id>.filters`), so the sort filter effectively lets each user set their own default. Listing selection has no equivalent and always starts at the first listing. Adding listings also flips `onlySearch` to false, which replaces the search bar with the listings header.

SpoilerPlus has no home page, so the tradeoff lands the way you described. Closing this.

Issue #744, opened 2026-09-08
[zh.copymanga] Source Broken - Need URL Update


jie347:
The old site is still online but blocks IPs from mainland China.
New URL: https://www.copy4000.com
Please update the source to this address. Thanks.

PR #756, opened 2026-09-13
feat(zh.copymanga): add account favorites

rereva0611:
Summary: Adds basic CopyManga account login, token renewal, an account-favorites listing, an optional Home favorites scroller, and detail-page add/remove actions. Keeps source ID zh.copymanga and updates version 19 to 20. Data handling: Login credentials are posted only to the configured CopyManga base URL at /api/v1/login. The returned site token is kept in Aidoku defaults and sent only to CopyManga favorites endpoints as the required authorization header/cookie. Favorite writes try api.copy4000.com first, then the selected CopyManga source host as a compatibility fallback. No telemetry, analytics, third-party requests, committed credentials, or credential/token logging were added. Validation: cargo fmt --check; cargo clippy -- -D warnings; cargo test -- --test-threads=1; aidoku package; aidoku verify package.aix. Login, favorite listing, add/remove, search, and reading were manually tested on a physical device using the feature-equivalent Plus build. AI assistance: AI was used as a development aid; I reviewed and iterated on the implementation and removed unnecessary logging. Checklist: source version updated; source ID unchanged; changes limited to this source.


sources/zh.copymanga/src/auth.rs
```diff
@@ -0,0 +1,137 @@
...
+/// token 失效（401）时利用 App 保存的凭据静默重登一次。
+pub fn try_relogin() -> bool {
+	let Some(username) = defaults_get::<String>(LOGIN_USERNAME_KEY).filter(|u| !u.is_empty())
+	else {
+		return false;
+	};
+	let Some(password) = defaults_get::<String>("login.password").filter(|p| !p.is_empty()) else {
+		return false;
+	};
+	println!("copymanga: relogin after token expiry");
```
kkantan: there shouldn't be any print statements left in production builds.

sources/zh.copymanga/src/lib.rs
```diff
@@ -1,21 +1,27 @@
...
 	imports::std::send_partial_result,
-	register_source,
+	println, register_source,
 };
 use html::{ChapterPage as _, FiltersPage as _, GenresPage as _, KeyPage as _, MangaPage as _};
 use json::{chapter_list, search};
 use net::Url;
 
-struct Copymanga;
+pub(crate) struct Copymanga;
```
kkantan: what is this necessary for?

sources/zh.copymanga/src/home.rs
```diff
@@ -0,0 +1,57 @@
...
+fn favorites_component(entries: &[Manga], title: &str) -> HomeComponent {
+	let listing = Listing {
+		id: String::from("f:fav"),
+		name: String::from(title),
+		..Default::default()
+	};
+	let links: Vec<Link> = entries
+		.iter()
+		.map(|manga| Link {
+			title: manga.title.clone(),
```
kkantan: you should consume the `entries` when passed in so that the content does not need to be unnecessarily cloned.

sources/zh.copymanga/src/home.rs
kkantan, line 1: a home page should not ever be empty unless there is some error. I think favorites may be better handled in a listing, but I am not sure if you can only have a single optional dynamic listing with no home page. thus, I believe your options are to either implement a more complete home page with multiple components, or have another main listing that's the default when opening the source.

sources/zh.copymanga/src/favorites.rs
```diff
@@ -0,0 +1,492 @@
...
+		base_url()?,
+		collect_type.param(),
+	);
+
+	let mut response = Request::get(&url)?.authed()?.send()?;
+	if response.status_code() == 401 && try_relogin() {
+		response = Request::get(&url)?.authed()?.send()?;
+	}
+	if response.status_code() == 401 {
+		return Err(error!("登錄已失效，請重新在設置中登錄"));
```
kkantan: you can use `bail!`, as you do elsewhere.

kkantan (changes requested):
can you also address #744 and add the new url?

@aphronyx if you would like to review you can, otherwise I will merge.

rereva0611:
Thank you for taking the time to review this PR. This is my first contribution to the Aidoku Community, and I appreciate your guidance.

This PR has been updated based on the review feedback and subsequent validation:

- Removed all production `println!` calls.
- Limited the test module to `#[cfg(test)]`.
- Made `Copymanga` private again; `pub(crate)` was unnecessary.
- Changed the favorites-listing 401 handling to use `bail!`.
- Removed the potentially empty Home implementation. The source now uses dynamic listings: the default listing is the website’s “New Releases” discovery page, with “My Favorites” available after login. This also removes unnecessary cloning of favorite entries.

Regarding #744:

After verification, `www.copy4000.com` is still not reliably accessible from mainland China. `www.copy5000.com` is the latest confirmed working domain there, so it is now the default URL. `www.copy4000.com` remains available as a fallback.

This update also adds a detail-page comments link, independent toggles for the comments and favorite buttons, and Simplified Chinese source UI text and filter labels.

For manga already saved in the Aidoku library, Aidoku loads the cached details directly. Existing library entries need to be refreshed once for the comments and favorite actions to be saved into their local details. A note about this is included in the source settings.

The source has been rechecked with formatting, tests, Clippy, packaging, and `aidoku verify`.

sources/zh.copymanga/res/source.json
```diff
@@ -1,9 +1,11 @@
 {
 	"info": {
 		"id": "zh.copymanga",
-		"name": "拷貝漫畫",
-		"version": 19,
+		"name": "拷贝漫画",
```
kkantan, line 4: the website uses the traditional version `拷貝漫畫` instead for titles, so I believe that should be the main name. however, you could add `altNames` with `拷贝漫画` as an entry.

sources/zh.copymanga/src/html.rs
```diff
@@ -71,6 +71,51 @@ impl FiltersPage for Document {
...
+				let authors = card
+					.select("span a")?
+					.filter_map(|author| author.text())
+					.collect();
+				Some(Manga {
+					key,
+					title,
+					cover: Some(cover),
+					authors: Some(authors),
+					status: MangaStatus::Unknown,
```
kkantan, line 100: this is part of the default so it doesn't need to be set explicitly.
```suggestion
```

sources/zh.copymanga/src/html.rs
```diff
@@ -71,6 +71,51 @@ impl FiltersPage for Document {
...
+					.replace(".328x422.jpg", "");
+				let authors = card
+					.select("span a")?
+					.filter_map(|author| author.text())
+					.collect();
+				Some(Manga {
+					key,
+					title,
+					cover: Some(cover),
+					authors: Some(authors),
```
kkantan, line 99: you should avoid setting authors in listing results as long as they can be fetched in details.

sources/zh.copymanga/src/favorites.rs
```diff
@@ -0,0 +1,507 @@
...
+	auth::{AuthedRequest as _, try_relogin},
+	net::{Url, base_url, manga_url},
+};
+
+const BUTTON_HOST: &str = "https://www.copy5000.com";
+
+#[cfg(test)]
+pub(crate) const fn button_host() -> &'static str {
+	BUTTON_HOST
+}
```
kkantan, line 11: why is this necessary?

sources/zh.copymanga/src/favorites.rs
```diff
@@ -0,0 +1,507 @@
...
+use aidoku::{
```
kkantan, line 12: all `use` statements should be grouped together at the top, followed by `const` definitions.

sources/zh.copymanga/src/favorites.rs
```diff
@@ -0,0 +1,507 @@
...
+	let authors = comic
+		.author
+		.map(|authors| authors.into_iter().map(|a| a.name).collect());
+	Manga {
+		key: comic.path_word,
+		title: comic.name,
+		cover,
+		authors,
+		url,
+		status,
```
kkantan, line 103: if this is only for listing results, you should avoid setting authors and probably url. status is fine, but not necessary or really standard.

sources/zh.copymanga/src/favorites.rs
```diff
@@ -0,0 +1,507 @@
...
+	charged: Result<MangaPageResult>,
+) -> Result<MangaPageResult> {
+	if let Ok(charged) = charged {
+		free.has_next_page = free.has_next_page || charged.has_next_page;
+		free.entries.extend(charged.entries);
+	}
+	Ok(free)
+}
+
+fn error_text(err: &AidokuError) -> String {
```
kkantan, line 166: i think you could consume the error here and not clone the message.
```suggestion
fn error_text(err: AidokuError) -> String {
```

sources/zh.copymanga/src/favorites.rs
```diff
@@ -0,0 +1,507 @@
...
+		lines.push(actions);
+	}
+
+	if lines.is_empty() {
+		return None;
+	}
+	let mut out = lines.join("\n\n");
+	if !description.is_empty() {
+		out.push_str("\n\n———\n\n");
+		out.push_str(description);
```
kkantan, line 408: I think the buttons should come after the description, at the bottom, so that the preview that shows in the series view is the description text. however, if you would rather have the buttons come first, you should disable the settings enabling them by default.

sources/zh.copymanga/src/favorites.rs
```diff
@@ -0,0 +1,507 @@
...
+	let token = crate::auth::token().ok_or_else(|| error!("请先在设置中登录"))?;
+	let mut response = send(url.into(), &token)?;
+	if response.status_code() == 401
+		&& try_relogin()
+		&& let Some(fresh) = crate::auth::token()
+	{
+		response = send(url.into(), &fresh)?;
+	}
+	if response.status_code() == 401 {
+		return Err(error!(
```
kkantan, line 487: you could use `bail!` here and for the other two.

sources/zh.copymanga/src/lib.rs
```diff
@@ -1,17 +1,26 @@
...
+mod tests;
 
 use aidoku::{
-	Chapter, DeepLinkHandler, DeepLinkResult, DynamicFilters, Filter, FilterValue, Manga,
-	MangaPageResult, Page, Result, Source,
+	BasicLoginHandler, Chapter, DeepLinkHandler, DeepLinkResult, DynamicFilters, DynamicListings,
+	Filter, FilterValue, Listing, ListingKind, ListingProvider, Manga, MangaPageResult,
+	NotificationHandler, Page, Result, Source,
 	alloc::{String, Vec},
+	bail, error,
```
kkantan, line 16:
```suggestion
	prelude::*,
```

sources/zh.copymanga/src/lib.rs
```diff
@@ -112,4 +160,76 @@ impl DynamicFilters for Copymanga {
 	}
 }
 
-register_source!(Copymanga, DeepLinkHandler, DynamicFilters);
+fn listings(is_logged_in: bool) -> Vec<Listing> {
```
kkantan, line 163: is this necessary to have in a function separate from `get_dynamic_listings`?

kkantan (changes requested):

rereva0611:
Thank you for the continued review. I updated the implementation with the following changes:

- Updated the primary source name to `拷貝漫畫` and added `拷贝漫画` as a Simplified Chinese alias. The source ID remains unchanged.

- Listing entries no longer populate status, author, or other metadata that has not been verified by the detail page. Favorites entries now keep only the required key, title, and cover.

- Removed test-only visible interfaces and low-value constant tests. The related tests now live in their owning module, without expanding the production API for test access.

- Direct early-error branches use `bail!` where appropriate. `Option`-to-`Result` conversions retain `ok_or_else(...)?`, which is idiomatic for that case.

- Consolidated Aidoku imports and removed unused helper code.

- The comments and favorite buttons retain their current layout but are disabled by default and can be enabled in source settings.

- I kept `listings(is_logged_in)` because it is production logic that centralizes the dynamic-listing rule and is not a test-only helper.

Additional maintenance changes:

- Favorite state is cleared on login, logout, and account changes to avoid stale state crossing accounts.

- Favorite writes fall back to the alternate host only when the current host is clearly unavailable. Unknown outcomes are not retried automatically.

- Chapters are sorted globally by upload time, newest first.

- The favorites listing no longer uses a free/paid category filter. It is treated as one complete favorites list.

- Favorites pagination now continues when a full page is returned even if the site reports unreliable pagination metadata, preventing entries after the first 50 from being omitted.

- Added focused tests for the unfiltered favorites request, second-page offset, and continued pagination after 50 entries.

For transparency, I used AI as a review and implementation aid:

- DeepSeek V4.1 Flash was used to review each source, resource, and configuration file for redundant logic, request and data handling, test boundaries, and potential privacy concerns.

- Astra was used to consolidate those findings and compare them with existing Aidoku Community patterns and the review feedback.

- Terra was used for scoped implementation changes. AI suggestions were treated as review input rather than accepted automatically.

I reviewed the resulting changes and re-ran formatting, tests, Wasm-target static checks, packaging, and `aidoku verify`.

kkantan (approved):

MZiR429:
_I wasn't sure if this was something that needed a separate issue, so I'm leaving the feedback here instead._

_Just to be clear, I understand the intention behind adding Favorites, and I have no issue with using AI as a development tool. My concern is that this update made the overall browsing experience worse._

Previously, the home page gave direct access to **Recently Updated** and **Most Popular**. Now both are moved behind the Search button. For users who do not use login/favorites, the home page is basically left with **Recently Added** as the only content.

This is not just an extra tap. **Recently Updated** also provided a quick way to check whether any manga you follow had new chapters, without having to refresh the entire Library. That is especially useful when the Library contains a large number of titles or categories, or when a full refresh takes a long time. For example, refreshing a Library with over 1,000 manga can take 5 to 10 minutes, and the refresh time can increase quickly as the Library grows.

The new flow is also awkward because opening Search brings up the keyboard and hides the navigation bar. When browsing **Recently Updated** or **Most Popular**, users can no longer directly switch to Library, History, Settings, etc. The old flow was much simpler for normal browsing.

There is also some unnecessary mixing of Simplified and Traditional Chinese. The manga titles and category lists in the drop-down menu returned by the site are still in Traditional Chinese. Changing only some UI text to Simplified Chinese does not really improve readability and instead makes the interface feel less consistent.

Using AI to help write the code is not the issue. The issue is that the generated changes do not seem to have been properly checked by the submitter from an actual user-flow perspective. AI was then also used to review and verify those changes. That makes it easy to miss problems caused by the design decision itself.

Favorites is a useful addition, but it should have been added without making the existing browsing flow less convenient.

kkantan:
I believe the correct solution to this issue is to implement a home page for the source that matches the website, and have the listings on the website reflected inside the source. you could create an issue for this if you would like. I also agree the filter titles should remain traditional chinese. I didn't check the website when reviewing.

if you want to use the previous version of the source instead of the current one you could download it from here: https://github.com/Aidoku-Community/sources/actions/runs/28867390451/artifacts/8149826885.

rereva0611:
Thank you for explaining how you used the previous source. I understand the issue now: moving Recently Updated and Most Popular behind Search made normal browsing less convenient, and Recently Updated was also a useful way to check for new chapters without refreshing a large Library. I did not account for either use case when I changed the home page.

I added account favorites because I wanted to keep using my CopyManga collection in Aidoku, but that addition should not have made the existing browsing flow worse. I will review the v19 behavior alongside the website’s current home page and work on restoring direct access to its browsing lists. I will test the resulting flow in Aidoku, including navigation to Library and other tabs, before proposing an update.

I also agree that changing only some interface text to Simplified Chinese has made the source inconsistent. I will check the labels against the website and keep the filter titles in Traditional Chinese, as suggested. Manga titles returned by the website are a separate matter from the source’s interface labels.

I used AI to help develop and review the changes, but the missed user flows are my responsibility. The code checks I ran did not catch this design problem. Thank you for taking the time to describe the impact so clearly.

dairidong:
Why this pr was merged?

PR #758, opened 2026-09-13
feat: add en.dynasty

rereva0611:
## Summary

- Add `en.dynasty` for Dynasty Scans (Dynasty Reader).
- Includes recent updates, keyword search, site type/tag/sort filters, details, chapters, page loading, and deep links.
- The source is marked adult (`contentRating: 2`).

## Verification

- `cargo fmt --check`
- `cargo clippy -- -D warnings`
- `cargo test -- --test-threads=1` — 9 real-site tests passed
- `aidoku package` and `aidoku verify package.aix`
- Tag filtering was also verified in Aidoku on device.

## Review notes

- Source-initiated data requests are rate-limited to 1 request/second to reduce site-side 503s.
- No account login, custom cookies, telemetry, or analytics are implemented. The only source-managed persistent data is a local cache of public cover URLs.
- AI was used as a development aid; the implementation was manually reviewed and verified.


kkantan (commented):
there are a few things wrong here, but the main thing is that dynasty scans already exists as a legacy source, and a new source should handle migration properly. I was already working on a dynasty rewrite, and I finished it up in #759, so I will close this in favor of that one.

PR #762, opened 2026-09-15
fix(multi.xcomic): rebuild source against the sites new title API

### LINKED ISSUE #763: Xcomic source broken
Xcomic is not working

PoppingMangoSources:
The site moved feeds to its title API: get_comic_browse_items throws inside its resolver on every call and get_comic_latestUploads answers empty, so browse, search and Latest Update were all dead. 

1. 	Read them from get_title_browse_items, get_title_recentlyAdded, and get_title_latestUploads, expanding each title to its editions through comicNodes so a manga key stays a comic id and existing library entries keep working

2. Offer every edition on a word search and one row per work when browsing, each labelled with the team behind it

3. 	Send every selected content rating explicitly: an empty list is the site's safe default here, not "all"
4. 	Rebuild the filters against the site's own search page, genres read live, its current types and demographics, a typed chapter-count range, and no Upload Status, Most Views or duplicate content-rating options
5. Chapter lists default to the deduplicated form, nothing is lost, it still returns a full chapter list for the edition you opened. The site pages that form a thousand at a time; the full list pages a hundred at a time, and a title with hundreds of uploads got an empty 429 on the later pages.
6. Entries now carry the team behind each edition (subName), which also gives chapters the scanlator instead of the aggregator in srcName.


PoppingMangoSources:
Site has now moved its browse endpoint as well, I’ll refactor to title endpoints where necessary when I get time. Making this a draft pull request for now.

PoppingMangoSources:
Done refactoring! 

Browse, search, recently added, and the latest feed moved to the title endpoints, editions read from comicNodes so keys stay comic ids, edition teams labelled from subName, content ratings sent explicitly, genres read live from the search page, types and demographics rebuilt against the site.

Typed custom chapter count filter added, Upload Status filter and Most Views sorts dropped (legacy code that doesn’t return anything on actual site), chapter lists deduplicated by default and home sections reworked since most views is dead.

kkantan (approved):

PR #766, opened 2026-09-20
fix(multi.mangadotnet): Fix pornographic and languages filter

### LINKED ISSUE #734: MangaDotNet search for adult content is a little bugged
The current version would not display adult content properly due to a query bug. You can still search for them if you remove the content ratings in the filter.

A better bug fix will be provided once I figured out the api query behind it.

jianmingyong:
Fixes #734

Added initial language filter based on common upload seen from some official releases. For general browsing, select English would do and that should be the default option.

Logout button now actually log you out. This is for reauthentication as sometimes you can get 401 when the cookie is invalidated.

A better error message for the authentication had been added.

Forced update to 0.9.0 since I am using the new features only available from there.


sources/multi.mangadotnet/res/source.json
```diff
@@ -2,13 +2,36 @@
...
+			"ru",
+			"sv",
+			"th",
+			"tl",
+			"tr",
+			"vi",
+			"zh-hk"
 		],
-		"minAppVersion": "0.8.4"
+		"minAppVersion": "0.9.0"
```
kkantan, line 34: is the only 0.9 api you are using the `clearCookiesOnLogOut` option? I think that the source should still work with this option on older versions.

if not, shouldn't this be `0.9` instead of `0.9.0`, since the app version is `0.9`? I may be wrong though.

jianmingyong:
New site update just broke something else. Delaying the update.

sources/multi.mangadotnet/res/source.json
jianmingyong: the 0.9 and 0.9.0 seems same. I guess that it would be a matter of preference.

jianmingyong:
Fixed the bookmark page due to API changes.
Added OEL filters.
Some more better error message common for this site.
Cloudflare workaround for those who get stuck in infinite loop. Most wouldn't need to touch this, VPN users may need this button.

kkantan (approved):

